PR #20, opened 2018-08-13
Default child nodes

y-kkamil:


core/src/services/routing.js
```diff
@@ -15,6 +15,26 @@ const getViewUrl = pathData => {
   return lastElement ? lastElement.viewUrl : '';
 };
 
+const getDefaultPathSegment = function(pathData) {
```
jesusreal: I would propose to refactor the function to the following:
```
const getDefaultPathSegment = function(pathData) {
  const lastElement = [...pathData.navigationPath].pop();
  const pathExists = lastElement.children.find(
    childNode => childNode.pathSegment === lastElement.defaultPathSegment
  );
  if (lastElement.defaultPathSegment && pathExists) {
    return lastElement.defaultPathSegment;
  } else {
    return lastElement.children[0].pathSegment;
  }
};
```

It is a little less efficient because you will calculate always pathExists, but the code is imho much more readable and it pays off

core/src/services/routing.js
```diff
@@ -15,6 +15,26 @@ const getViewUrl = pathData => {
...
+  const lastElement = [...pathData.navigationPath].pop();
```
jesusreal: Even though it is longer, I would use ```const lastElement = pathData.navigationPath[pathData.navigationPath.length - 1]``` because it is more explicit.

core/src/services/routing.js
```diff
@@ -216,6 +236,12 @@ export const handleRouteChange = async (path, component, node, config) => {
     const params = parseParams(pathUrl.split('?')[1]);
     const nodeParams = getNodeParams(params);
 
+    if (path !== '' && !viewUrl) {
```
jesusreal: Should we also write unit tests for this new piece of code?

core/test/routing.spec.js
```diff
@@ -546,5 +547,92 @@ describe('Routing', () => {
       component.set({ viewUrl: 'http://otherurl.de/app.html!#/someUrl' });
       assert.isTrue(routing.isNotSameDomain(config, component));
     });
+
```
jesusreal: We could refactor this tests to remove duplication of the mock data. The following approach should be fine, since each test case is receiving a fresh mock data object, so mutation in one test case does not affect other test cases:
```
    describe('defaultPathSegments', () => {
      const getDefaultPathSegment = routing.__get__('getDefaultPathSegment');
      const getBasePathData = function() {
        return {
          navigationPath: [
            {
              // DOESN'T MATTER
            },
            {
              pathSegment: 'groups',
              children: [
                {
                  pathSegment: 'stakeholders',
                  viewUrl:
                    '/sampleapp.html#/projects/1/users/groups/stakeholders'
                },
                {
                  pathSegment: 'customers',
                  viewUrl: '/sampleapp.html#/projects/1/users/groups/customers'
                }
              ]
            }
          ],
          context: {}
        }
      }

      it('should return first child if no dps set', () => {
        const pathData = getBasePathData();
        assert.equal(getDefaultPathSegment(pathData), 'stakeholders');
      });

      it('should child with pathSegment equal to defaultPathSegment', () => {
        const pathData = getBasePathData();
        pathData.navigationPath[1].defaultPathSegment = 'customers';
        assert.equal(getDefaultPathSegment(pathData), 'customers');
      });

      it('should return first child if given defaultPathSegment does not exist', () => {
        const pathData = getBasePathData();
        pathData.navigationPath[1].defaultPathSegment = 'NOSUCHPATH';
        assert.equal(getDefaultPathSegment(pathData), 'stakeholders');
      });
    });
```

Additionally, please also remove the describe block in line 447 if you agree it is unnecessary.

core/test/routing.spec.js
```diff
@@ -546,5 +547,92 @@ describe('Routing', () => {
...
+    describe('defaultPathSegments', () => {
+      const getDefaultPathSegment = routing.__get__('getDefaultPathSegment');
+
+      it('should return first child if no dps set', () => {
```
jesusreal: I would write always default path segment instead of dps, since this is not straightforward for everybody

PR #28, opened 2018-08-31
Bugfix/client hash navi

maxmarkus:
**Description**

When luigi.navigate event gets dispatched, luigi changes the whole location, which triggers a hard-reload when using hash based routing. This small patch fixes it, if there is a hash in the url, we exchange only the hash.


jesusreal:
Well done!

PR #45, opened 2018-09-07
Luigi UMD

aartek:
**Description**
Here is my example of Luigi as UMD. I've also changed directory structure a bit - 'core/public' dir is no longer needed, because it's created by webpack task. `core/package.json` indicates `public/luigi.js` as main library file, so it can be easy imported and 2 package.jsons are no longer needed. I've also added 2 examples of how luigi could be used with webpack or with classic global import. Please treat this PR as an example, not as a bulletproof and well tested solution.


jesusreal:
Hi @aartek, thanks a lot for the feedback you have provided through your pull request. I took a deep look at it and I summed up in [this issue](https://github.com/kyma-project/luigi/issues/68) the enhancements we will perform, which are still subject to changes after internal grooming. 

I am closing this pull request, since it was meant for review but not for merging, and next steps will be taken on scope of the aforementioned issue.

Best regards,
Jesús Real

PR #67, opened 2018-09-10
Navigation Node visibility with custom nodeAccessibilityResolver

### LINKED ISSUE #44: Visibility restriction base implementation
**Description**

As a Luigi core developer i want to define a function easily where i can set the visibility of a node/microfrontend based on my constraints.

While/after the Luigi nav tree is constructed, it should be processed through a visibility checking mechansism which removes non-visible nodes.
Add a demo to the sample angular application and a documentation

**Solution Proposal** : 
Introduce a configurable `nodeAccessibilityResolver` function

maxmarkus:
**Description**

Implemented `Luigi.config.navigation.nodeAccessibilityResolver` for checking Node access restrictions

**Related issue(s)**
Fixes #44


core/examples/luigi-sample-angular/src/assets/sampleconfig.js
```diff
@@ -1,5 +1,21 @@
...
-  return new Promise(function(resolve) {
+var navigationPermissionChecker = function (nodeToCheckPermissionFor, parentNode, currentContext) {
+  // depending on the current path and context returns true or false
+  // true means the current node is accessible, false the opposite
+  const mockCurrentUserGroups = ['admins'];
+  if (nodeToCheckPermissionFor.constraints) {
+    // check if user has required groups
+    return nodeToCheckPermissionFor.constraints.filter(
+      function (c) {
+        return mockCurrentUserGroups.indexOf(c) !== -1;
```
dariadomagala: you could use
```
return mockCurrentUserGroups.includes(c)
```

core/examples/luigi-sample-angular/src/assets/sampleconfig.js
```diff
@@ -1,5 +1,21 @@
-var getAllProjects = function() {
-  return new Promise(function(resolve) {
+var navigationPermissionChecker = function (nodeToCheckPermissionFor, parentNode, currentContext) {
```
dariadomagala: In navigation.js you use this function with variables _currentPath, currentNavigationStructure, node_ - could you use similar naming in both places? Because at the beginning I wasn't sure what this function uses 🙂

maxmarkus: done

dariadomagala: Great :) But I saw you changed arrow functions to use ES5 - could you also remove this one 'const'  used here since it's not ES5 compatible as well?

docs/navigation-configuration.md
```diff
@@ -58,6 +59,10 @@ window.Luigi.setConfig({
 
 ### Navigation parameters
 
+- **nodeAccessibilityResolver** allows you to define a permission checker function that gets executed on every Node. If it returns false, the Node and it's children are removed from the navigation structure. You may want to add custom values to Nodes to receive it in the **nodeAccessibilityResolver** function. See [angular sampleconfig.js](../core/examples/luigi-sample-angular/src/assets/sampleconfig.js) for our constraints example.
```
kazydek: * it's children -> `its children`
* it returns false -> ``it returns `false` ``
* preferably it would be great if you could replace the passive voice here with such a structure:``If it returns `false`, {subject} removes the Node and its children from the navigation structure.``
* You may want to add custom values to a Node to receive it in the **nodeAccessibilityResolver** function. ->  ` To receive a Node in the **nodeAccessibilityResolver** function, add custom values to it. `(I hope the meaning is kept?)

docs/navigation-configuration.md
```diff
@@ -58,6 +59,10 @@ window.Luigi.setConfig({
...
+
+## Nodes
```
kazydek: Is `##` instead of`###`intentional here?

maxmarkus: Yes, since this is a standalone feature while Nodes section is a collection of settings for a Node.

    {
      navigation: {
        nodeAccessibilityResolver: someFunction,
        nodes: [node1, node2] // every nodeN consists of parts described in ## Nodes
      }
    }

PR #89, opened 2018-09-17
Fix authSuccessfull execution on each page refresh

dariadomagala:
**Description**

- Fix authSuccessfull execution on each page refresh


kwiatekus:
We should rename the  event handlers, as they might be confusing for the user.
Because actually the intention was to call it every time the auth data was read and validated ("the user is authenticated"). Include @hardl & beta users

kwiatekus:
We agreed that this event should be triggered only after successful authentication (not after page reload)

PR #104, opened 2018-09-25
Dynamic pathSegments in navigation Nodes

### LINKED ISSUE #94: "Dynamic nodes" feature
The dynamic node functionality enables the user to specify the path segment not to be a static string but a dynamic value that is then passed to the context of the view.

Two implementation options:
1) as in yaas builder using a dedicated property "dynamic". Its value is used as the key in the context object.
2) convention in existing pathSegment property, e.g. ":"-prefix, so ":context_key"  

See https://devportal.yaas.io/tools/buildersdk/index.html#IntegrateaBuildermodule, section "Dynamic nodes in navigation tree" for more info.

maxmarkus:
**Description**
Added possibility to add dynamic nodes. pathSegments prefixed with colon (:) will be substituted in pathSegment, viewUrl and context and are available as LuigiClient.getPathParams().

**Related issue(s)**
Resolves #94


docs/lifecycle.md
```diff
@@ -9,9 +9,11 @@ LuigiClient.addInitListener(initFn)
...
+LuigiClient.getPathParams()
 ````
 
 - **addInitListener(initFn)** registers a listener that is called with a context object as soon as Luigi is instantiated. Defer your application bootstrap if you are dependent on authentication data from Luigi.
 - **addContextUpdateListener(contextUpdatedFn)** registers a listener that is called upon any navigation change, and calls the **contextUpdatedFn** with a new context object.
 - **getNodeParams()** returns the configuration object of the active navigation Node.
-- **getEventData()** returns the context object. Usually it is not required as the **addContextUpdateListener** receives the same values.
\ No newline at end of file
+- **getEventData()** returns the context object. Usually it is not required as the **addContextUpdateListener** receives the same values.
+- **getPathParams()** returns the dynamic path parameters of the active url.
```
kazydek: `URL`

docs/navigation-configuration.md
```diff
@@ -64,7 +65,7 @@ window.Luigi.setConfig({
 
 ## Nodes
 
-- **pathSegment** specifies the partial URL of the current segment. A static settings example reflects `luigidomain.test/settings`, while a dynamic one loads on any other value.
+- **pathSegment** specifies the partial URL of the current segment. A static settings example reflects `luigidomain.test/settings`, while a dynamic, prefixed with colon, loads on any other value. **pathSegments** must not contain slashes.
```
kazydek: `a dynamic one, prefixed with a colon, `

Issue #108, opened 2018-09-26
Backdrop layer reveal some gaps - CSS bug 


kwiatekus:
**Description:**

Backdrop layer does not cover the left navigation area properly. There is narrow gap on the right side and a gap at the bottom

![Screen Shot 2018-09-26 at 15.21.56.png](https://images.zenhubusercontent.com/5b8689081dd32d2cc77238ca/cd2d7347-3375-45c6-9fca-16d17c4d18cd)

PR #116, opened 2018-09-27
Change per-component backdrop to global one

parostatkiem:
**Description**
Instead of having separate backdrop for `LeftNav` and `TopNav` components we now cover the whole `App` component with backdrop which is easier to understand and more _natural_ according to what the backdrop actually does. The `<iframe>` with page content is displayed above the backdrop layer so it has to have its own backdrop (no changes here).

- Change per-component backdrop to global one

**Related issue(s)**
#108


core/src/App.html
```diff
@@ -1,14 +1,17 @@
-<div id="app">
-    <div class="fd-page {hideNav? 'iframeContainerNoNav' : 'iframeContainer'}" use:init="context"></div>
-    <TopNav pathData={navigationPath} />
-    <LeftNav pathData={navigationPath} />
-</div>
+<Backdrop>
```
jesusreal: I personally wouldn't add the Backdrop element outside the application container. And I believe we don't need to wrap any element inside the Backdrop. I would write this markup like that:
```
<div id="app">
    <Backdrop></Backdrop>
    <div class="fd-page {hideNav? 'iframeContainerNoNav' : 'iframeContainer'}" use:init="context"></div>
    <TopNav pathData={navigationPath} />
    <LeftNav pathData={navigationPath} />
</div>
```

core/src/Backdrop.html
```diff
@@ -1,10 +1,9 @@
 {#if backdropActive}
-    <div
-     class="fd-ui__overlay fd-overlay fd-overlay--modal {heightCssClass}"
-     aria-hidden="false"
-    >
-    </div>
+<div class="fd-ui__overlay fd-overlay fd-overlay--modal {heightCssClass}" aria-hidden="false">
```
jesusreal: Why do we need these slot tags here?

core/src/App.html
```diff
@@ -220,6 +218,10 @@
       cursor: pointer;
     }
 
+    .fd-page{
```
jesusreal: I agree this is the right way to do it, otherwise it will be inconsistent, and we will continue having displaying issues. But I am not convinced with the fact that we are providing a feature but it does not work out of the box, since every app using luigi needs to wrap the iframe in a container styled with this z-index value.

parostatkiem: This z-index value is assigned at Lugii Core's level so it works out of the box. The app developer doesn't have to implement any z-index value to the container.

core/src/App.html
parostatkiem: I think my approach is more natural and makes the code more referring to what we can actually see and do in the app. The backdrop wraps whole content and makes it unusable/unclickable, that's why I wrapped `content` in `backdrop`. I can change it if you really want, it's basically just code-understanding thing :)

core/src/Backdrop.html
parostatkiem: Because we wrap App's content in `<Backdrop>` tags. `<slot>` is for displaying what's inside component tags. It's the same as `<ng-content>` in Angular. Am I wrong?

core/src/App.html
parostatkiem: How about modifying your suggestion to
```
<div id="app">
    <Backdrop>
    <div class="fd-page {hideNav? 'iframeContainerNoNav' : 'iframeContainer'}" use:init="context"></div>
    <TopNav pathData={navigationPath} />
    <LeftNav pathData={navigationPath} />
</Backdrop>
</div>
```

?

Issue #114, opened 2018-09-27
Remove legacy luigi client compatibility properties


hardl:
Luigi.token = currentContext.context.idToken;
    // Luigi.linkManagerData = currentContext.context.linkManagerData;
    // Luigi.currentEnvironmentId = currentContext.context.currentEnvironmentId;
    // Luigi.sessionId = currentContext.context.sessionId;

PR #129, opened 2018-10-05
Add automatic loading indicator to show processing

### LINKED ISSUE #100: Processing overlay feature
**Description**

Hi, 
It would be great if Luigi will provide possibility to display processing overlay during component initialisation.

**Reasons**

As a developer I would like to have opportunity to inform user that component is initialising and then that initialisation is finished by using processing overlay to provide better user experience.

maxmarkus:
* Implemented loading indicator to automatically show on new iframe creation and hide triggered by Luigi init callback
* added config params for completely disabling the indicator and for disabling auto-hide feature
* added LuigiClient methods for show and hide: `LuigiClient.uxManager().showLoadingIndicator()` and `LuigiClient.uxManager().hideLoadingIndicator()` 
* updated externalExample.html to use Fundamental-UI and LuigiClient
* added a demo indicator for externalExample.html
* added unit and e2e tests
* added documentation

Resolves #100 and #114

### DIFF
--- a/docs/ux-manager.md
+++ b/docs/ux-manager.md
@@ -11,7 +11,8 @@ LuigiClient.uxManager().showLoadingIndicator()
 LuigiClient.uxManager().hideLoadingIndicator()
 ````
 
+
 - **addBackdrop()** adds a backdrop to block the top and side navigation. It is based on Fundamental UI Modal, which you can use in your micro front-end to achieve the same behaviour.
 - **removeBackdrop()** removes the backdrop.
-- **showLoadingIndicator()** adds a backdrop with a loading indicator for the micro front-end frame. This overrides the `loadingIndicator.enabled` setting.
+- **showLoadingIndicator()** adds a backdrop with a loading indicator for the micro front-end frame. This overrides the **loadingIndicator.enabled** setting.
 - **hideLoadingIndicator()** removes the loading indicator. Use it after calling **showLoadingIndicator()** or to hide the indicator when you use the `loadingIndicator.hideAutomatically: false` Node configuration.


docs/navigation-configuration.md
```diff
@@ -73,6 +73,8 @@ window.Luigi.setConfig({
 - **context** sends the specified object as context to the view. Use this parameter in combination with the dynamic **pathSegment** to receive the context through the context listeners of **Luigi client**. This is an alternative to using the dynamic value in the **viewUrl**.
 - **defaultChildNode** sets the child node that Luigi activates automatically if the current node has no **viewUrl** defined. Provide **pathSegment** of the child node you want to activate as a string.
 - **isolateView** renders the view in a new frame when you enter and leave the Node. This setting overrides the same-domain frame re-usage. The **isolateView** is disabled by default.
+- **loadingIndicator.enabled** shows a loading indicator when switching between micro front-ends. Super fast micro front-ends might want to disable this feature. The **loadingIndicator.loadingIndicator** is enabled by default.
```
klaudiagrz: 1. ...`when you switch between micro front-ends.`
2. "Super fast micro front-ends might want to disable this feature." -> Super fast micro front-ends are not humans so they cannot literally "want" anything. Please paraphrase.

docs/navigation-configuration.md
```diff
@@ -73,6 +73,8 @@ window.Luigi.setConfig({
...
+- **loadingIndicator.hideAutomatically** disables automatic hiding of the loading indicator. Use `LuigiClient.uxManager().hideLoadingIndicator()` to hide it manually in your micro front-end during startup. The **loadingIndicator.hideAutomatically** is enabled by default.
```
klaudiagrz: 1. `the automatic hiding`
2. `during the startup.`

docs/ux-manager.md
```diff
@@ -7,7 +7,11 @@ Here is a code sample and a definition of the different parameters:
...
 LuigiClient.uxManager().removeBackdrop()
+LuigiClient.uxManager().showLoadingIndicator()
+LuigiClient.uxManager().hideLoadingIndicator()
 ````
 
 - **addBackdrop()** adds a backdrop to block the top and side navigation. It is based on Fundamental UI Modal, which you can use in your micro front-end to achieve the same behaviour.
-- **removeBackdrop()** removes the backdrop.
\ No newline at end of file
+- **removeBackdrop()** removes the backdrop.
+- **showLoadingIndicator()** adds a backdrop with a loading indicator for the micro front-end frame
```
klaudiagrz: Add a dot at the end.

docs/navigation-configuration.md
maxmarkus: Something like "If you have a super fast micro frontend, you can disable this feature to prevent  loading indicator flickering."

klaudiagrz: Just two corrections:
1. `front-end` as in the rest of the document
2.  `prevent loading indicator from flickering.` ?

core/test/routing.spec.js
```diff
@@ -131,6 +135,43 @@ describe('Routing', () => {
         component.get().hideNav,
         window.Luigi.config.settings.hideNavigation
       );
+      assert.equal(component.get().showLoadingIndicator, true);
```
jesusreal: If possible, please try not to add this additional assert to this test, and for testing the loading indicator logic create two tests as small as possible

client/luigi-client.js
```diff
@@ -79,6 +83,10 @@
       if (window._init) {
         window._init(currentContext.context);
       }
+
```
jesusreal: Maybe we could save us the roundtrip and hide the loading indicator automatically from core.

core/examples/luigi-sample-angular/src/assets/sampleconfig.js
```diff
@@ -341,16 +341,28 @@ Luigi.setConfig({
         {
           pathSegment: 'ext',
           label: 'External Page',
+          loadingIndicator: {
+            enabled: true,
```
jesusreal: I would remove `enabled: true` here and two times more below, since this is what we consider already as default value.

core/test/routing.spec.js
```diff
@@ -10,14 +10,17 @@ import { afterEach } from 'mocha';
 
 describe('Routing', () => {
   let component;
+  let lastObj;
```
jesusreal: I would rewrite this part as follows:
```
beforeEach(() => {
  let lastObj = {};
  component = {
    set: obj => {
      Object.assign(lastObj, obj);
    },
    get: () => lastObj
  };
});
```
Like that:
* Defining the `lastObj` variable makes clear you are not using it anywhere outside the `beforeEach` block in the spec file.
* You can define the get function once instead of redefining it every time the set function is called 
* No need to reassign `lastObj`, since Object.assign sets the merged object to its first parameter.

core/test/routing.spec.js
maxmarkus: I'll leave it, since loading indicator is standard now and enabled by default

maxmarkus:
Docu for navigation-configuration.md and ux-manager.md got updated after approval, please check again @klaudiagrz

docs/ux-manager.md
```diff
@@ -7,7 +7,11 @@ Here is a code sample and a definition of the different parameters:
...
 LuigiClient.uxManager().removeBackdrop()
+LuigiClient.uxManager().showLoadingIndicator()
+LuigiClient.uxManager().hideLoadingIndicator()
 ````
 
 - **addBackdrop()** adds a backdrop to block the top and side navigation. It is based on Fundamental UI Modal, which you can use in your micro front-end to achieve the same behaviour.
-- **removeBackdrop()** removes the backdrop.
\ No newline at end of file
+- **removeBackdrop()** removes the backdrop.
+- **showLoadingIndicator()** adds a backdrop with a loading indicator for the micro front-end frame. This overrides the `loadingIndicator.enabled` setting.
```
klaudiagrz: **loadingIndicator.enabled** -> please make it bold as with the rest of the parameters

Issue #134, opened 2018-10-10
mockAuth not working and also not found enough documentation on it.


webdesignermumbai:
I like this project but found more scope for documentation.

I tried by adding mockAuth setting in basicConfiguration.js 
But its loading overview page directly without login page.

    Luigi.setConfig({
      auth: {
        use: 'mockAuth',
        mockAuth: {
          authorizeUrl: `${window.location.origin}/assets/auth-mock/login-mock.html`,
          logoutUrl: `${window.location.origin}/assets/auth-mock/logout-mock.html`,
          post_logout_redirect_uri: '/logout.html',
          authorizeMethod: 'GET',
          oAuthData: {
            client_id: 'egDuozijY5SVr0NSIowUP1dT6RVqHnlp'
          }
        }

please let me know if something more to configure to open authorizeUrl (login-mock.html) in application before overview page.

Please help me or guide me to use above mockAuth features.

**Expected result**

**Actual result**

**Steps to reproduce**

**Troubleshooting**

maxmarkus:
Your settings are fine.

I think it is ignored because `sampleConfiguration.js` is referenced in `index.html` instead of `basicConfiguration.js`. 

I've identified some more issues with the basic configuration and the dev server for path routing, so we need to rely on hash routing for now. Theses issues will be fixed when closing this issue.

After changing index.html to reference `basicConfiguration.js` and adding the configuration below, start dev server with `npm run start`. 

    Luigi.setConfig({
      auth: {
        use: 'mockAuth',
        mockAuth: {
          authorizeUrl: `${window.location.origin}/assets/auth-mock/login-mock.html`,
          logoutUrl: `${window.location.origin}/assets/auth-mock/logout-mock.html`,
          post_logout_redirect_uri: '/logout.html',
          authorizeMethod: 'GET',
          oAuthData: {
            client_id: 'egDuozijY5SVr0NSIowUP1dT6RVqHnlp'
          }
        }
      },
      navigation: {
        nodes: () => [
          {
            pathSegment: 'home',
            label: 'Home',
            children: [
              {
                pathSegment: 'hw',
                label: 'Hello World!',
                viewUrl: '/assets/sampleexternal.html'
              },
              {
                pathSegment: 'one',
                label: 'Section one',
                viewUrl: '/assets/sampleexternal.html#one'
              },
              {
                pathSegment: 'two',
                label: 'Section two',
                viewUrl: '/assets/sampleexternal.html#two'
              }
            ]
          }
[...]

PR #135, opened 2018-10-11
Improve basic configuration and documentation

maxmarkus:
- Improved basicConfiguration.js
- Extended example documentation
- Adapted webpack configuration

Resolves `#134`


core/examples/luigi-sample-angular/src/assets/basicexternal.html
```diff
@@ -0,0 +1,56 @@
...
+</head>
+<body>
+<a name="ext"></a>
+<header class="fd-page__header">
+  <div class="fd-action-bar">
+    <div class="fd-action-bar__header">
+      <h1 class="fd-action-bar__title">
+          Basic Example page
+      </h1>
+      <p class="fd-action-bar__description">This page behaves equally to an external hosted micro front-end.</p>
```
pekura: I do not get what is the intent of saying this. Maybe it would be interesting to say that this example behaves like a single page app.

maxmarkus: How about: `This micro front-end behaves like a single page application.`

pekura: That is better I think.

core/examples/luigi-sample-angular/README.md
```diff
@@ -30,13 +30,22 @@ To have this application running, follow these steps:
     ```
 
 4. Start the example application from the `luigi/core/examples/luigi-sample-angular` folder.
+
+`/assets/sampleConfiguration.js` is the default configuration with a showcase of all available features. If you want to try out the much simpler example, change the configuration reference in `index.html` to `basicConfiguration.js`.
```
bszwarc: try out the much simpler > try out a much simpler

core/examples/luigi-sample-angular/README.md
```diff
@@ -30,13 +30,22 @@ To have this application running, follow these steps:
...
+
+To run Luigi with hash based routing, change the **routing.useHashrouting** configuration to `true` and run the following command:
+    ```bash
+    npm run start
+    ```
+
+To run Luigi with path based routing, change the **routing.useHashrouting** configuration to `false` and run the following command:
```
bszwarc: Maybe we can have it as an unordered list:
- To run Luigi with hash based routing ...
- To run Luigi with path based routing ...

It would make the instructions clearer.

PR #140, opened 2018-10-16
Fix css issues in core and angular example

jesusreal:
**Description**

- Make logout button visible over iframe container (core)
- Fix logout button styling and functionality
- Fix styling on external page (angular example)
- Fix styling on logout page (angular example)

**Related issue(s)**


core/src/App.html
```diff
@@ -1,16 +1,16 @@
...
-    {#if showLoadingIndicator}
-    <div in:fade='{delay: 250, duration: 250}' out:fade='{duration: 250}' class="fd-page spinnerContainer" aria-hidden="false" aria-label="Loading">
-        <div class="fd-spinner"><div></div></div>
-    </div>
-    {/if}
-    
-    <TopNav pathData={navigationPath} />
-    <LeftNav pathData={navigationPath} />
-    </Backdrop>
+  </Backdrop>
```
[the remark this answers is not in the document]
jesusreal: I tested all those options. The old implementation was not working as expected, none of the elements was inside the backdrop. Apart from that, you need to have the backdrop at the same DOM level as the top and left nav, so that it covers the elements properly. If you had those navs inside the backdrop container, you would have displaying issues due to how we display the different containers. Additionally, having the iframe container inside the backdrop is IMHO a good approach, because it allows you to apply the iframe on top of the backdrop without using any z-index.

core/src/Backdrop.html
```diff
@@ -1,9 +1,6 @@
-{#if backdropActive}
-<div class="fd-ui__overlay fd-overlay fd-overlay--modal {heightCssClass}" aria-hidden="false">
+<div class="{backdropActive ? 'fd-ui__overlay fd-overlay fd-overlay--modal ' + heightCssClass : ''}" aria-hidden="false">
```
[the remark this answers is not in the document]
jesusreal: slot was not working as expected, the containers from Authorization.html were always rendered outside of the backdrop.

core/src/App.html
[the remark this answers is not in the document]
parostatkiem: I didn't see any unexpected behaviour but if you did, let's leave it like this.

core/examples/luigi-sample-angular/src/logout.html
```diff
@@ -9,6 +9,7 @@
   <meta name="description" content="">
   <meta name="keywords" content="logout">
 
+  <link rel='stylesheet' href='../node_modules/fundamental-ui/dist/fundamental-ui.min.css'>
```
maxmarkus: I fear this approach will not work once the user has don a `npm run build` and deploys the dist folder, as it will miss the fundamental-ui then?

core/examples/luigi-sample-angular/webpack.config.js
```diff
@@ -397,6 +397,27 @@ module.exports = {
           from: {
             glob: 'luigi-client.js'
           }
+        },
```
maxmarkus: The css file is copied, but not injected into index.html
A quick tryout of `webpack.config.js`  entry.styles: ['./src/styles.css', './node_modules/fundamental-ui/dist/fundamental-ui.min.css']   did not work in my case, needs investigation how to reference the css properly.

maxmarkus: Alternatively manually put it into sampleapp.html and remove the injection reference in angular.json

jesusreal: `fundamental-ui.min.css` injected via `webpack.config.js`. Please check again.

maxmarkus: Good job!

PR #142, opened 2018-10-17
Allow cross origin preserve view

maxmarkus:
Disabled cross origin check when using go back functionality. 

Simple test case: 

- Open `sampleexample.html` and add go back link `<a href="javascript:LuigiClient.linkManager().goBack('some back value')">go back</a>`
- Open `project.component.html` and add preserve view link for external view: 
`<li class="fd-list-group__item"><a href="javascript:void(0)" (click)="luigiClient.linkManager().navigate('/ext', null, true)">with preserved view: ext cross origin</a></li>`


core/src/services/routing.js
```diff
@@ -87,6 +87,9 @@ const removeElementChildren = node => {
 export const isNotSameDomain = (config, component) => {
   if (config.iframe) {
     const componentData = component.get();
+    if (componentData.isNavigateBack) {
```
kwiatekus: I wouldnt change the logic here. The fact that we are in the middle of back navigation doesnt make suddenly that source and target url belong  to the same origin.

IMHO this logic should be unchanged, but we should rather mainpulate the if statement in the navigateIframe function

sth like this
```
    if(!componentData.isNavigateBack &&
    (isNotSameDomain(config, component) ||
    hasIframeIsolation(component) ||
    Boolean(config.builderCompatibilityMode))
{
...
}
```

Issue #127, opened 2018-10-04
Possibility to check whether other microfrontend exists 


golojg:
**Description**

It would be great if Luigi will provide possibility to check whether other microfrontend exists. Someting like:
`
LuigiClient.linkManager()
            .path("/project/other-module")
            .exists()
`

**Reasons**

We want show option to navigate to other microfrontend only if it is available.

kwiatekus:
A way to go:
make an async call to the main application to check if the target node exists (is configured, not if its target view is up and running)

Do we want to include dynamic nodes in the check? From one point of view it might be a performance issue to resolve all (also dynamic) nodes of the examined path. On the other hand it's not really nice to only check statically defined nodes.

kwiatekus:
It should work also for the  paths relative to the current location

PR #154, opened 2018-10-23
Small refactorings

jesusreal:
**Description**

- As part of #127, I identified some parts of the codebase that could be refactored and decided to create a separate pull request for those changes.

**Related issue(s)**


client/luigi-client.js
```diff
@@ -196,10 +198,10 @@
           var navigationOpenMsg = {
             msg: 'luigi.navigation.open',
             sessionId: sessionId,
-            params: Object.assign(
-              { link: path, relative: relativePath },
-              options
-            )
+            params: Object.assign(options, {
```
maxmarkus: You know, that this updates the options object, is this intentionally? Previously it was a new object that receives additional options.

jesusreal: We are mutating this object in other places, so this way we are making sure the content of this object is always up to date. Additionally, as discussed offline the object is generated with default values every time we use the link manager

Issue #58, opened 2018-09-10
404 support for non-existing paths


kwiatekus:
**Description**

If user provides a non-existing path (like http://localhost:4200/#/projects/pr5/users in luigi-sample-angular) he is not given any error message.

Instead, luigi traverses nodes structure and displays the deepmost node that is reachable (in this case, Project overview).

It's not the worst solution, but not perfect either. Implement some kind of 404s handling:
 - keep the mechanism for navigating to the deep-most node possible, and additionally
 - show a error notification (https://sap.github.io/fundamental/demo-pages/alert-overlay-demo-page.html) saying sth like "Could not map the exact target node for the requested route"

Remove TODO from routing.js::handleRouteChange function.

**Reasons**

 User is not given a clear message that the exact path was not reachable.

kwiatekus:
When the exact match was not found we route to the parent, but the wrong address stays in the url

![Screen Shot 2018-11-09 at 11.49.24.png](https://images.zenhubusercontent.com/5b8689081dd32d2cc77238ca/d8bcedb5-0b56-4e6a-be5e-7f334807a2a3)

kwiatekus:
As agreed, the message should contain the desired route

PR #162, opened 2018-10-25
404 support for non existing paths

parostatkiem:
**Description**

- create a helper function to detect unexpected segments in the URL
- display an alert (taken from FundamentalUI) when when needed
- test the helper function
- move existing test files to separate folders (according the code hierarchy)

**Related issue(s)**

Refeers #58


core/src/App.html
```diff
@@ -226,6 +240,20 @@
               console.error('goBack() not possible, no preserved views found.');
             }
           }
+
+          if ('luigi.displayAlert' === e.data.msg) {
+            if (e.data.detail) {
```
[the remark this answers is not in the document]
parostatkiem: `errorMessage` looks good to me. `message` would be even better but it'd collide with the postMessage name which is the same.

core/src/utilities/helpers.js
```diff
@@ -79,3 +79,25 @@ export const prependOrigin = path => {
...
+
+export const containsAllSegments = (sourceUrl, targetPathSegments) => {
+  const mandatorySegments = sourceUrl.split('/');
+  let result = true;
+
+  if (targetPathSegments.length !== mandatorySegments.length + 1) {
+    return false; //we can already tell segments will not match so no loop is needed
+  }
+
+  mandatorySegments.forEach((segment, index) => {
```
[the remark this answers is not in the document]
parostatkiem: Good point, I'll improve it

core/src/App.html
```diff
@@ -226,6 +240,20 @@
...
+              this.set({ alertMessage: e.data.detail });
+            } else {
+              console.error(
+                'Error alert has been triggered but no message was provided in "detail" property'
```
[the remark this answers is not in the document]
dariadomagala: I wouldn't show to the user an error saying that developers misconfigured something. That wouldn't tell the user anything, but could cause a confusion. In my opinion it would be sufficient that this information is visible in the console and in the UI there is a more generic message. 
We can discuss it with the team.

dariadomagala:
Code looks better, but please also add an example and UI test as I requested previously.

core/examples/luigi-sample-angular/src/app/project/project.component.html
```diff
@@ -1,76 +1,97 @@
...
+				<a href="javascript:void(0)" (click)="luigiClient.linkManager().navigate('/overview')">absolute: to overview</a>
+				<app-code-snippet data="luigiClient.linkManager().navigate('/overview')"></app-code-snippet>
+			</li>
+			<li class="fd-list-group__item">
+				<a href="javascript:void(0)" (click)="luigiClient.linkManager().navigate('users/groups/stakeholders')">relative: to stakeholders</a>
+				<app-code-snippet data="luigiClient.linkManager().navigate('users/groups/stakeholders')"></app-code-snippet>
+			</li>
+			<li class="fd-list-group__item">
+				<a href="javascript:void(0)" (click)="luigiClient.linkManager().fromClosestContext().navigate('/users/groups/stakeholders')">closest
+					parent: to stakeholders</a>
```
dariadomagala: why did you split it in the middle of the sentence? :)

core/examples/luigi-sample-angular/src/app/project/project.component.html
```diff
@@ -1,76 +1,97 @@
...
+<section class="fd-section" *ngIf="projectId && projectId !== 'overview'">
+	<div class="fd-panel">
+		<p><strong>LuigiClient - wrong paths in linkManager.navigate():</strong></p>
+		<ul class="fd-list-group">
+			<li class="fd-list-group__item">
+				<a href="javascript:void(0)" (click)="luigiClient.linkManager().navigate('/overview/maskopatol')">Partitialy wrong link</a>
+				<app-code-snippet data="luigiClient.linkManager().navigate('/overview/maskopatol')"></app-code-snippet>
+			</li>
+			<li class="fd-list-group__item">
+				<a href="javascript:void(0)" (click)="luigiClient.linkManager().navigate('/maskopatol/has/a/child')">Totally wrong link</a>
```
dariadomagala: When I click on it the first time, it shows the error message, which is good. Then I close this message, and click on it the second time, the error message is not shown again. 
If we would have more 'broken' links, this message would be shown only for the first one, not all of them, and it's not how it should work.

core/src/services/routing.js
```diff
@@ -313,6 +313,8 @@ export const handleRouteChange = async (path, component, node, config) => {
     }
 
     if (!containsAllSegments(pathUrl, pathData.navigationPath)) {
+      const matchedPath = await matchPath(pathUrl);
+      window.history.pushState('object or string', 'Title', '#' + matchedPath);
```
kwiatekus: Is it supported by all blowsers?

jesusreal:
The implementation is not working for a url with query parameters like http://localhost:4200/#/projects/pr2/settings?~foo=bar&

Additionally, there are some refactorings that I would apply. I generated the following [Uploading patch-404-support-for-non-existing-paths.zip…](patch file) you can apply it with this command: `git apply patch-404-support-for-non-existing-paths`.

core/src/utilities/helpers.js
```diff
@@ -86,6 +86,36 @@ export const prependOrigin = path => {
   return window.location.origin;
 };
 
+export const containsAllSegments = (sourceUrl, targetPathSegments) => {
```
jesusreal: There is still a bug when the user introduces a url with a trailing slash like '/projects/pr2/' (`sourceUrl` param in function)
Furthermore the implementation could be simpler. My proposal is [the following](https://github.com/kyma-project/luigi/files/2594680/404-support-for-non-existing-paths-2.zip) (unit tests are still green after refactoring)

PR #169, opened 2018-10-25
Enable Luigi Client (microfrontend) to check whether a path exists in main app

jesusreal:
**Description**

- Implement feature: client can ask core if a path exists 
- Update documentation
- Add an example in angular app
- Add UI tests

**Related issue(s)**


core/examples/luigi-sample-angular/src/app/project/project.component.html
```diff
@@ -53,6 +53,23 @@ <h1 class="fd-section__title">Project {{ projectId }}</h1>
         <a href="javascript:void(0)" (click)="luigiClient.linkManager().navigate('/settings', null, true)">with preserved view: project to global settings and back</a>
         <app-code-snippet data="luigiClient.linkManager().navigate('/settings', null, true)"></app-code-snippet>
       </li>
+      <li class="fd-list-group__item check-path">
+        <a href="javascript:void(0)" (click)="luigiClient.linkManager().navigate('/settings', null, true)">Check if path
```
maxmarkus: Please remove the click handler since it navigates to settings.

client/luigi-client.js
```diff
@@ -267,6 +278,42 @@
...
+         *    () => { pathExists = true; },
+         *    () => { pathExists = false; },
+         *  );
+         */
+        pathExists: function pathExists(path) {
+          var currentId = Date.now();
+          pathExistsPromises[currentId] = {
+            success: function() {},
+            error: function() {},
+            then: function(success, error) {
```
maxmarkus: Is there a reason that the user can provide success and error functions instead of just having one simple `pathExists().then((result: Boolean) => {})`. I know it represents a Promise interface with "resolve, reject", but it feels a bit complicated.

core/examples/luigi-sample-angular/e2e/tests/luigi-client-features.spec.js
```diff
@@ -107,6 +107,29 @@ describe('Luigi client features', () => {
             expect(loc.hash).to.eq('#/projects/pr2');
           });
         });
+
+      // check if path exists
+      cy.go(-2);
```
maxmarkus: I tend to not like relying on browser history, but since the flows in here rely on eachother it's ok.

core/examples/luigi-sample-angular/src/app/project/project.component.html
jesusreal: Good catch, I forgot to do that before creating the pr.

client/luigi-client.js
maxmarkus: I initially thought of using it directly inside a template, but since it is out of Zone.js it would not work anyway. So it is fine for me.

PR #187, opened 2018-11-02
Luigi core config refactorings

jesusreal:
**Description**

- Create class for Luigi Config
- Avoid using interval to check for Luigi config existency -> Luigi initializes up to 50ms faster
- Update unit tests
- Use correct key for screenshots folder in Cypress config

**Related issue(s)**


core/src/services/config.js
```diff
@@ -1,85 +1,96 @@
...
-  const propertyPath = property.split('.');
-  while (nextValue && propIndex < propertyPath.length) {
-    nextValue = nextValue[propertyPath[propIndex++]];
+class LuigiConfigManager {
+  constructor() {
+    this.configReadyTimeout = {
+      valueMs: 2000,
+      id: undefined
+    };
+    this.config;
```
parostatkiem: What doest this line do? `this.config` stays undefined after `constructor()` finished so there's no need to "declare" this variable... Correct me if I'm wrong.

core/src/services/config.js
```diff
@@ -1,85 +1,96 @@
...
+
+    this.configReadyCallback = function() {};
```
parostatkiem: I think here's the same story as with `this.config;` above. It doesn't change anything if you remove this line.

core/src/services/config.js
```diff
@@ -1,85 +1,96 @@
...
-  object,
-  property,
-  ...parameters
-) => {
-  let value = getConfigValueFromObject(object, property);
-  if (isFunction(value)) {
-    value = value.apply(this, parameters);
-    if (isPromise(value)) {
-      return value;
+  getConfig() {
```
parostatkiem: You can be more _modern_ here and use a getter.
```
 get Config() {
    return this.config;
  }
```
and then
`this.getConfig()` -> `this.Config`.

I'm not requesting this change but I'd like you to know it :)

core/test/config.spec.js
```diff
@@ -1,49 +1,41 @@
 const chai = require('chai');
 const assert = chai.assert;
-const config = require('../src/services/config');
+const LuigiConfig = require('../src/services/config').LuigiConfig;
```
parostatkiem: How about `import { LuigiConfig } from '../src/services/config';`?

parostatkiem: _+_  respectively in other tests

core/src/services/config.js
```diff
@@ -1,85 +1,96 @@
...
- */
-export const getConfigBooleanValue = property => {
-  const configuredValue = getConfigValueFromObject(
-    window.Luigi.config,
-    property
-  );
-  if (configuredValue === true || configuredValue === 'true') {
-    return true;
+  setConfig(configInput) {
+    // console.log("Set config");
```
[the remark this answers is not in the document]
jesusreal: removed

core/src/services/config.js
jesusreal: This initial value is meant to be able to call this.configReadyCallback without the need of an `if` block to prove if it is defined or not.

core/src/services/config.js
jesusreal: I consider that option, but I see unnecessary to create a breaking change for this tiny improvement.

core/src/services/config.js
jesusreal: removed

core/src/services/config.js
```diff
@@ -1,85 +1,94 @@
...
- */
-export const getConfigValueFromObject = (object, property) => {
-  let propIndex = 0;
-  let nextValue = object;
-  const propertyPath = property.split('.');
-  while (nextValue && propIndex < propertyPath.length) {
-    nextValue = nextValue[propertyPath[propIndex++]];
+class LuigiConfigManager {
+  constructor() {
+    this.configReadyTimeout = {
```
dariadomagala: Do we need this object?
The `valueMs` is not used anywhere. My suggestion would be to change this `configReadyTimeout` from an object to a string with just the `id` value.
The other option would be to use this `valueMs` in the timeout below, but then I don't like the name of the object, because what is an id of a timeout then?

core/src/services/config.js
```diff
@@ -1,85 +1,94 @@
...
- */
-export const getConfigValueAsync = (property, ...parameters) => {
-  return getConfigValueFromObjectAsync(
-    window.Luigi.config,
-    property,
-    parameters
-  );
-};
+// Expose it window for user app to call Luigi.setConfig()
+window.Luigi = LuigiInstance;
```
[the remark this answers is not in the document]
jesusreal: For consistency I decided to expose the whole Luigi config instance, as this is what we are exporting through es module. But on the other side, I am in favour of exposing as less as possible to the window object, so I will apply the change you proposed. Thanks!

core/src/utilities/async-helpers.js
```diff
@@ -21,3 +28,23 @@ export const waitForKeyExistency = (
...
+  ...parameters
+) => {
+  let value = getConfigValueFromObject(object, property);
+  if (isFunction(value)) {
+    value = value.apply(this, parameters);
+    if (isPromise(value)) {
+      return value;
+    }
+  }
+  return wrapAsPromise(value);
```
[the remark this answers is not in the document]
jesusreal: Good catch, removed

core/src/services/config.js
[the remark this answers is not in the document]
jesusreal: Now I know why I did not do that. Luigi config is a class, so setConfig is interacting with other members of the class. Therefore IMHO the best solution is to expose the whole instance

core/src/services/config.js
jesusreal: Id of a timeout makes sense to me, it is the way you identify a timeout you set in case you want to do something with it, like removing it. I will use valueMs instead the magic number '2000' when setting the timeout and keep the object.

docs/navigation-configuration.md
```diff
@@ -286,7 +286,7 @@ When loading, the **viewUrl** uses the following dynamic URL parameters:
 - `sort = asc`
 
 ```
-Luigi.setConfig({
+window.Luigi.setConfig({
```
dariadomagala: It still should work with just `Luigi.setConfig` right?

jesusreal: yes, everywhere is used with Luigi.setConfig. I just added it here to the documentation for consistency

dariadomagala: So it's ok with this task, but I guess for the future we should set it to just `Luigi.setConfig`, it just seems easier ;)

PR #192, opened 2018-11-05
Split main js files

### LINKED ISSUE #164: Split routing.js and navigation.js to smaller files
Those two files are growing and we should think about splitting them into smaller chunks. This mostly applies to `routing.js`
Move private helper functions to `helpers` file.
Pay attention which parts are dealing with navigation and which with routing (we are having a mix right now)
Split also the tests.

parostatkiem:
**Description**

- `routing.js`, `navigation.js` and `helpers.js` are split into files with _100 to 200_ lines of code
- tests are split into smaller files respectively
- *Mocha* scans `/test` directory recursively
- some functions are moved from `navigation.js` to `routing.js` and vice versa

**Related issue(s)**

Resolves #164


kwiatekus:
We should merge this task AFTER all other luigi tasks are merged.
At the current state it is hard to follow and reproduce all the changes that are pushed to the master as part of other tasks.

We should start a new PR from scratch and start splitting from scratch. we should do it at the time when there is no other luigi tasks being worked on (i.e.  use the time of console luigification).

PR #196, opened 2018-11-05
Fix for goBack when not using micro-frontend without routing

maxmarkus:
Skip url change in LuigiClient when going back.


client/luigi-client.js
```diff
@@ -109,10 +109,13 @@
         var hashRoutingModeActive =
```
jesusreal: Please move this variable initialisation to the inner scope of the new `if` block you created.

maxmarkus: done

Issue #208, opened 2018-11-08
Two dynamic parameters don't seem to load the correct url


corey-aloia-sap:
**Description**
Two dynamic parameters don't seem to load the correct url
**Expected result**

When a go to a url:
`http://tenant.luigi/sites/123456/2`
I would expect Luigi to navigate to
`http://microfrontend/sites/123456/2`
**Actual result**
It navigates to
`http://microfrontend/sites/:siteId/2`

My config.json looks like the following:
```
{
        "pathSegment": "sites",
        "label": "My UI",
        "viewUrl": "http://localhost:4200/sites",
        "category": "Development",
        "keepSelectedForChildren": true,
        "navigationContext": "my-root",
        "children": [
          {
            "pathSegment": "import",
            "label": "Import",
            "viewUrl": "http://localhost:4200/sites/import"
          },
          {
            "pathSegment": ":siteId",
            "label": "Sites Details",
            "viewUrl": "http://localhost:4200/sites/:siteId",
            "children": [
              {
                "pathSegment": ":pageId",
                "label": "Page Details",
                "viewUrl": "http://localhost:4200/sites/:siteId/:pageId"
              }
            ]
          }
        ]
}
```

If I open the url
`http://tenant.luigi/sites/123456/2`
it contacts the url
`http://localhost:4200/sites/:siteId/2`

**Troubleshooting**
I work at SAP, so please feel free to contact me on slack and I can do a screen share.

Corey Aloia

corey-aloia-sap:
Thanks for fixing this! When can we expect the 0.3.8 release?

PR #211, opened 2018-11-12
Multiple path parameters do not get replaced in view url

pekura:
Fix for: multiple path parameters do not get replaced in view url #208.


core/src/services/routing.js
```diff
@@ -123,20 +123,30 @@ const escapeRegExp = string => {
   return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
 };
 
-const replaceVars = (viewUrl, params, prefix) => {
+const replaceVars = (viewUrl, params, prefix, noParenthesis) => {
```
jesusreal: Shall we use a variable called 'parenthesis' so that below we use `parenthesis` instead of `!noParenthesis`? The only changes needed (apart from the usage of the variable) are:
`const replaceVars = (viewUrl, params, prefix, parenthesis = true) => { // line 126`
`viewUrl = replaceVars(viewUrl, componentData.pathParams, ':', false // line 158);`

core/test/routing.spec.js
```diff
@@ -380,6 +395,78 @@ describe('Routing', () => {
       );
     });
 
+    it('should set component data with path param', async () => {
```
jesusreal: Unit tests need refactoring but this will be addressed in scope of other tasks.

core/src/services/routing.js
jesusreal: If my tests were correct here we are replacing just the second dynamic parameter. If this is true, where are we replacing the first dynamic parameter and why aren´t we replacing all parameters at once in the same place?

jesusreal: Could you please explain this issue in detail so that we make sure this is fixed in that ticket?

pekura: It is explained in the related issue #213.

PR #215, opened 2018-11-13
mock auth logout page fix

### LINKED ISSUE #214: mockAuth error when auto-logout is enabled
When mockAuth is in use, after auto logout there is an error appearing in the console - logout page tries to fetch post_logout_redirect_uri from the window.location, and it's not there - this prevent Luigi from displaying proper logout page.

y-kkamil:
**Description**

Fixes the problem when using auto-logout 

**Related issue(s)**
Resolves #214


y-kkamil:
it's not needed, great catch. i have removed it.

PR #222, opened 2018-11-19
Login fix

hardl:

- Fixed issue with no nodes being visible if auth not being used

### DIFF
--- a/core/src/Authorization.html
+++ b/core/src/Authorization.html
@@ -52,10 +52,7 @@
 
   export default {
     oncreate() {
-      const idpProviderName = LuigiConfig.getConfigValue('auth.use');
-      const idpProviderSettings = LuigiConfig.getConfigValue(`auth.${idpProviderName}`);
-      const authorizationEnabled = !!idpProviderSettings;
-      if (!authorizationEnabled) {
+      if (!LuigiConfig.isAuthorizationEnabled()) {
         return;
       }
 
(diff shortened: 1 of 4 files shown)


core/src/navigation/services/navigation.js
```diff
@@ -1,12 +1,20 @@
...
+import {
+  getConfigValueFromObjectAsync
+} from '../../utilities/async-helpers';
 
 const isNodeAccessPermitted = (
   nodeToCheckPermissionFor,
   parentNode,
   currentContext
 ) => {
-  if (!isLoggedIn()) return false;
```
jesusreal: I see this already 3 times in our codebase, we could create a helper in LuigiConfig class and reuse it instead of duplicating code. [Here my proposal](https://github.com/kyma-project/luigi/files/2599011/luigi-222-login-fix.zip).

jesusreal: Pushed by myself as requested from pr creator

core/src/Authorization.html
```diff
@@ -52,10 +52,7 @@
 
   export default {
     oncreate() {
-      const idpProviderName = LuigiConfig.getConfigValue('auth.use');
```
kwiatekus: Looks like this is needed in the code executed below:
```
L82: idpProviderInstance = getIdpProviderInstance(idpProviderName, idpProviderSettings);
```

PR #226, opened 2018-11-20
Add check if modules are same domain

dariadomagala:
**Description**

- Add check if modules are same domain

**Related issue(s)**

### DIFF
--- a/core/src/services/routing.js
+++ b/core/src/services/routing.js
@@ -96,9 +96,27 @@ export const isNotSameDomain = (config, component) => {
       componentData.previousNodeValues.viewUrl
     );
     const nextUrl = getUrlWithoutHash(componentData.viewUrl);
-    return previousUrl != nextUrl;
+    const previousUrlOrigin = getLocation(previousUrl);
+    const nextUrlOrigin = getLocation(nextUrl);
+    if (previousUrl === nextUrl) {
+      return false;
+    } else if (previousUrlOrigin === nextUrlOrigin) {
+      const previousViewGroup = componentData.previousNodeValues.viewGroup;
+      const nextViewGroup = componentData.viewGroup;
+      if (!(previousViewGroup && nextViewGroup)) {
+        return true;
+      }
+      return previousViewGroup !== nextViewGroup;
+    }
+    return true;
   }
   return true;
+};
+
+const getLocation = url => {
+  var element = document.createElement('a');
+  element.href = url;
+  return element.origin;
 };
 
 export const hasIframeIsolation = component => {
@@ -288,6 +306,14 @@ const buildRoute = (node, path, params) =>
     ? path + (params ? '?' + params : '')
     : buildRoute(node.parent, `/${node.parent.pathSegment}${path}`, params);
 
+const findViewGroup = node => {
+  if (node.viewGroup) {
+    return node.viewGroup;
+  } else if (node.parent) {
+    return findViewGroup(node.parent);
+  } else return;
+};
+
 export const handleRouteChange = async (path, component, node, config) => {
   const defaultPattern = [/access_token=/, /id_token=/];
   const patterns =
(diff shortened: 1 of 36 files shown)


core/src/services/routing.js
```diff
@@ -93,11 +93,27 @@ export const isNotSameDomain = (config, component) => {
...
+      const nextUrlDomain = getDomainFromUrl(nextUrl);
+      return iFrameDomain != nextUrlDomain;
+    } else {
+      return true;
+    }
   }
   return true;
 };
 
+const getDomainFromUrl = url => {
```
kwiatekus: We should be able to detect multiple SPAs under one domain, i.e:
`https://example.com/foo/` <-- root of one SPA
`https://example.com/bar/` <-- root of another SPA

maxmarkus: I guess we do not have a concept / configuration option for that, since you would be currently only able to detect hash based SPAs, path based are undetectable by only the url. In my opinion this is a followup feature, which requires grooming.

core/src/services/routing.js
```diff
@@ -98,11 +98,29 @@ export const isNotSameDomain = (config, component) => {
       componentData.previousNodeValues.viewUrl
     );
     const nextUrl = getUrlWithoutHash(componentData.viewUrl);
-    return previousUrl != nextUrl;
```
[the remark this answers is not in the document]
dariadomagala: I will refactor the function so it doesn't have negation, but I don't like the `isSameViewGroup` nam, because in hash routing we don't have viewGroups. Maybe something more like `isSameApplication`, `isSameIFrame`?

docs/navigation-configuration.md
```diff
@@ -244,7 +246,7 @@ The node parameters are as follows:
 - **keepSelectedForChildren** focuses the navigation on its current hierarchy, omitting the display of children.
 - **loadingIndicator.enabled** shows a loading indicator when switching between micro front-ends. If you have a fast micro front-end, you can disable this feature to prevent flickering of the loading indicator. This parameter is enabled by default.
 - **loadingIndicator.hideAutomatically** disables the automatic hiding of the loading indicator once the micro front-end is loaded. It is only considered if the loading indicator is enabled. It does not apply if the loading indicator is activated manually with the `LuigiClient.uxManager().showLoadingIndicator()` function. If the loading indicator is enabled and automatic hiding is disabled, use `LuigiClient.uxManager().hideLoadingIndicator()` to hide it manually in your micro front-end during the startup. This parameter is enabled by default.
-
+- **viewGroup** defines a group of views in the same domain sharing a common security context, which allows for example frame re-using for better performance. This property is used only for the views that use path routing internally.
```
```suggestion
- **viewGroup** defines a group of views in the same domain sharing a common security context. Use this property to manage settings such as frame reuse for better performance. **viewGroup** is used only for the views that use path routing internally.
```

docs/navigation-configuration.md
```diff
@@ -244,7 +246,7 @@ The node parameters are as follows:
 - **keepSelectedForChildren** focuses the navigation on its current hierarchy, omitting the display of children.
 - **loadingIndicator.enabled** shows a loading indicator when switching between micro front-ends. If you have a fast micro front-end, you can disable this feature to prevent flickering of the loading indicator. This parameter is enabled by default.
 - **loadingIndicator.hideAutomatically** disables the automatic hiding of the loading indicator once the micro front-end is loaded. It is only considered if the loading indicator is enabled. It does not apply if the loading indicator is activated manually with the `LuigiClient.uxManager().showLoadingIndicator()` function. If the loading indicator is enabled and automatic hiding is disabled, use `LuigiClient.uxManager().hideLoadingIndicator()` to hide it manually in your micro front-end during the startup. This parameter is enabled by default.
-
+- **viewGroup** defines a group of views in the same domain sharing a common security context. Use this property to manage settings such as frame reuse for better performance. **viewGroup** is used only for the views that use path routing internally.
```
```suggestion
- **viewGroup** defines a group of views in the same domain sharing a common security context. This improves performance through reusing the frame. Use **viewGroup**  only for the views that use path routing internally.
```

bszwarc: As agreed with Daria :)

PR #238, opened 2018-11-22
viewUrl should not be mandatory in dynamic node

maxmarkus:
Added bugfix for error occured when not having defined a viewUrl within a dynamic node.


maxmarkus:
@jesusreal other keys are not an issue, as they are not mandatory by default. this was just an implementation issue, as `viewUrl` should have never been made mandatory.

PR #240, opened 2018-11-22
Improve luigi client docu and generate new

jesusreal:
**Description**

- Update documentation for context object methods in Luigi Client

**Related issue(s)**

### DIFF
--- a/client/luigi-client.js
+++ b/client/luigi-client.js
@@ -188,7 +188,7 @@
       return false;
     },
     /**
-     * Registers a listener that is called with a context object upon any navigation change.
+     * Registers a listener that is called with the context object upon any navigation change.
      * @param {function} contextUpdatedFn the listener function that is called every time Luigi context was changed
      * @memberof lifecycle
      */
--- a/client/luigi-client.js
+++ b/client/luigi-client.js
@@ -188,7 +188,7 @@
       return false;
     },
     /**
-     * Registers a listener that is called upon any navigation change.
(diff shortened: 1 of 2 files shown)


client/luigi-client.js
```diff
@@ -219,7 +219,7 @@
     },
     /**
      * Returns the context object. Typically it is not required as the {@link #addContextUpdateListener addContextUpdateListener()} receives the same values.
-     * @returns {{idToken: string, sessionId: string, currentEnvironmentId: string} } node parameters.
+     * @returns {Object} current context data
```
bszwarc:
```suggestion
     * @returns {Object} current context data.
```

client/luigi-client.js
```diff
@@ -188,7 +188,7 @@
       return false;
     },
     /**
-     * Registers a listener that is called upon any navigation change.
+     * Registers a listener that is called with a context object upon any navigation change.
```
bszwarc:
```suggestion
     * Registers a listener that is called with the context object upon any navigation change.
```
I have noticed we are  linking to a general {Object} definition throughout the document, but not in the case of context object - is it because it's a particular instance of the {Object}? I'm absolutely not saying it's wrong - just want to know :)

Below we have also mentioned the context object (Returns the context object) and a line below we have {Object} current context data.

jesusreal: The context object could contain any parameter, it reflects the state of the microfrontend

PR #251, opened 2018-11-27
Improvements for dex auth

maxmarkus:
Improvements which are required for console. 

- allow window.location.origin to be the redirect_uri (prevent routing if access_token is in the url)
- resilience for invalid iframe (on logout, re-login and)
- context switcher is now hidden if autologin is disabled and user logged out
- refactored backdrop due to some oncreate side effects, seems like there is a bug in svelte when using string concatenation inside template

### DIFF
--- a/core/src/Backdrop.html
+++ b/core/src/Backdrop.html
@@ -1,4 +1,4 @@
-<div class="{backdropActive ? 'fd-ui__overlay fd-overlay fd-overlay--modal ' + heightCssClass : ''}" aria-hidden="false">
+<div class="{backdropClass}" aria-hidden="false">
   <slot></slot>
 </div>
 
@@ -7,18 +7,14 @@
   export default {
     data() {
       return {
-        backdropActive: false
+        backdropActive: false,
+        backdropClass: ''
       }
     },
     oncreate() {
       const backdropDisabled = LuigiConfig.getConfigValue('settings.backdropDisabled');
       if (!backdropDisabled) {
-        if (this.options.data && this.options.data.heightCssClass) {
-          this.set({ heightCssClass: this.options.data.heightCssClass })
-        } else {
-          this.set({ heightCssClass: 'height-auto' })
-        }
-
+        this.setBackdropClass();
         window.addEventListener("message", (e) => {
           if ('luigi.add-backdrop' === e.data.msg) {
             this.set({ backdropActive: true });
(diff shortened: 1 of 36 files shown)


core/src/services/routing.js
```diff
@@ -288,6 +288,12 @@ const buildRoute = (node, path, params) =>
     : buildRoute(node.parent, `/${node.parent.pathSegment}${path}`, params);
 
 export const handleRouteChange = async (path, component, node, config) => {
+  if (
+    window.location.href.indexOf('access_token') !== -1 ||
```
pekura: 1) Can we do this check a bit more specific? At least indexOf('access_token=').
2) Can we make it configurable? Like list of regex expressions for paths to be ignored by Luigi routing? That way the Luigi app owner would have full control over it.

maxmarkus: Yes, thats a good idea :-)

docs/navigation-configuration.md
```diff
@@ -212,6 +212,7 @@ You can configure the way Luigi tackles routing in your application in the **Rou
 
 - **useHashRouting** defines either hash-based (`url.com/#/yourpath`) or path-based (`url.com/yourpath`) routing.
 - **nodeParamPrefix** sets the prefix character when using the `LuigiClient.linkManager().withParam()` function, which provides a way to simply attach query parameters to the view URL for activities such as sorting and filtering. The URL contains the parameters to allow deep linking. If you want to use a different character prefix, define yours here. The default character is `~`.
+- **skipRoutingForUrlPatterns** defines regex patterns that should be skipped by the router when listening for path changes. Used for excluding redirect_uri parameters. Default patterns are `[/access_token=/, '/id_token=/]`
```
```suggestion
- **skipRoutingForUrlPatterns** defines regex patterns to be skipped by the router when listening for path changes. This parameter is used for excluding **redirect_uri** parameters. Default patterns are `[/access_token=/, '/id_token=/]`.
```

core/src/navigation/TopNav.html
```diff
@@ -51,6 +53,7 @@
   import LogoTitle from './LogoTitle.html';
   import Authorization from '../Authorization.html';
   import { handleRouteClick } from '../services/routing.js';
+  import * as Navigation from './services/navigation.js';
```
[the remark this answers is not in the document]
kwiatekus: done

core/src/Backdrop.html
```diff
@@ -29,6 +25,24 @@
...
     },
+    onstate({ changed, current, previous }) {
+      if (changed.backdropActive) {
+        this.setBackdropClass();
+      }
+    },
+    methods: {
+      setBackdropClass() {
+        const baseClasses = 'fd-ui__overlay fd-overlay fd-overlay--modal ';
+        if (!this.get().backdropActive) {
```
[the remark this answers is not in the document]
maxmarkus: Less code, but more dependencies. Not a fan of this. I guess if we want to have it, we can write it by ourself with a small helper function. similar to `['class1', 'class2'].join(' ')`.

PR #252, opened 2018-11-27
Unsaved changes modal

### LINKED ISSUE #210: "Unsaved Changes" notifications
A microfrontend with form elements and "save/store" functionality should be able to mark its state "dirty" in order to trigger a confirmation modal when navigating to a different microfrontend to prevent losing unsaved changes.

Needed for preserving functionality introduced in https://github.com/kyma-project/kyma/issues/143

parostatkiem:
This change definitely needs a *backdrop* support which should be provided with #352. 
I left some comments with my suggestions about how Core's backdrop can be handled.

- Create new component called `ConfirmationModal`
- Handle the modal showing and hiding in `App.js` and return a promise with the result
- Trigger the modal in `routing.js::handleRouteChange()` if current page is dirty, then wait for result

**Related issue(s)**

Resolves #210 

## ‼️ IMPORTANT ‼️
Don't forget to swap your Console repo to the branch from [this PR:](https://github.com/jesusreal/console/pull/11)


core/examples/luigi-sample-angular/src/app/overview/overview.component.ts
```diff
@@ -57,4 +57,11 @@ export class OverviewComponent {
       description: 'navigation node configuration to redirect to another path'
     }
   ];
+
+  private sendDirtyEvent = () => {
```
pekura: It would be nice to have a dedicated method for setting dirty/not dirty state in luigi client - unless we decided to skip it for this issue.

core/src/services/routing.js
```diff
@@ -299,6 +299,44 @@ export const handleRouteChange = async (path, component, node, config) => {
...
   try {
+    if (
+      component &&
+      typeof component.get === 'function' &&
+      component.get().isDirty
+    ) {
+      const newUrl = window.location.href;
+      const oldUrl = component.get().persistUrl;
+
+      //pretend the url hasn't been changed
```
pekura: There is one scenario where it does not work well. When you keep pressing the browser back button and "No" in the dialog you land on the login page after a while.

core/src/services/routing.js
```diff
@@ -299,6 +299,47 @@ export const handleRouteChange = async (path, component, node, config) => {
...
+      component.get().unsavedChanges &&
+      component.get().unsavedChanges.isDirty
+    ) {
+      const newUrl = window.location.href;
+      const oldUrl = component.get().unsavedChanges.persistUrl;
+
+      //pretend the url hasn't been changed
+      oldUrl && history.replaceState(window.state, '', oldUrl);
+
+      if (
```
jesusreal: We should make sure the component is ready to deal with unsavedChanges feature before applying any logic like `history.replaceSate`. Therefore, I would propose to have this checks also above. My proposal would be sth like this:
```
const componentImplementsUnsavedChangesFeature = component &&
  typeof component.get === 'function' &&
  component.get().unsavedChanges &&
  component.get().unsavedChanges.isDirty &&
  typeof component.showModal === 'function' &&
  typeof component.hideModal === 'function';

if (componentImplementsUnsavedChangesFeature) {
  const newUrl = window.location.href;
  const oldUrl = component.get().unsavedChanges.persistUrl;

  //pretend the url hasn't been changed
  oldUrl && history.replaceState(window.state, '', oldUrl);
```

jesusreal: As stated in another comment, this might not be necessary: `typeof component.get === 'function' &&`

core/examples/luigi-sample-angular/src/app/overview/overview.component.ts
parostatkiem: Good idea, I'll create it

core/src/services/routing.js
parostatkiem: I'm working on it...

core/src/services/routing.js
```diff
@@ -299,6 +299,47 @@ export const handleRouteChange = async (path, component, node, config) => {
...
+        typeof component.showModal !== 'function' ||
+        typeof component.hideModal !== 'function'
+      ) {
+        return; //in case provided component doesn't have this functionality
+      }
+      component
```
jesusreal: Here `component` is always 'AppComponent'. Therefore the inputs to showModal method could be removed and rewrite the showModal function as it follows:
```
      showModal() {
        const title = 'Unsaved changes detected';
        const text = 'It looks like you might lose some data if you leave this page. Are you sure you want to do this?'
...
```
This does not affect the reusability of the Confirmation modal component

core/src/services/routing.js
jesusreal: Actually, just realised we don't need so many checks, since as stated in other of my comments, here `component` is always 'AppComponent'. This means *most of the checks are always true*. So the conditions can be reduced to:
```
if (component.get().unsavedChanges.isDirty) {
  const newUrl = window.location.href;
  const oldUrl = component.get().unsavedChanges.persistUrl;

  //pretend the url hasn't been changed
  oldUrl && history.replaceState(window.state, '', oldUrl);
```

core/examples/luigi-sample-angular/src/app/overview/overview.component.html
```diff
@@ -1,36 +1,40 @@
...
+	<div class="fd-section__header">
+		<h1 class="fd-section__title">Luigi Core Features</h1>
+	</div>
+	<div class="fd-panel">
+		<ul class="fd-list-group">
+			<li class="fd-list-group__item clickable" *ngFor="let item of luigiCoreLinks" (click)="luigiClient.linkManager().navigate(item.link)">
+				<strong>{{item.text}}</strong> &nbsp;&ndash; {{item.description}}
+			</li>
+			<li class="fd-list-group__item clickable" (click)="sendDirtyEvent()">
+				<strong>Pretend there are unsaved changes</strong> &nbsp;&ndash; shows a confirmation modal when user wants to leave
```
jesusreal: Please provide better guiding text that makes more straightforward to understand how to test the feature and actually test it.

core/src/services/routing.js
```diff
@@ -298,6 +297,44 @@ export const handleRouteChange = async (path, component, node, config) => {
...
+          'It looks like you might loose some data if you leave this page. Are you sure you want to do this?'
+        )
+        .then(() => {
+          // YES pressed
+          component.set({ isDirty: false });
+          path &&
+            handleRouteChange(path, component, node, config) &&
+            history.replaceState(window.state, '', newUrl);
+        })
+        .catch(() => {
```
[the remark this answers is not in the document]
jesusreal: Was mistaken, we can remove `catch + then` sequence and use the rejection handler in the first `then`.

core/src/ConfirmationModal.html
```diff
@@ -0,0 +1,44 @@
+<div class="fd-modal">
```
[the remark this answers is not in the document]
parostatkiem: 🤦‍♂️ I forgot about it...

core/src/services/routing.js
parostatkiem: They are not in the unit tets, that's why I wrote so many checks.

core/src/services/routing.js
parostatkiem: It does, you are unable to use the confirmation modal for any other purpose than this one.

jesusreal: I see your point, it does not affect reusing the ConfirmationModal component in other component, but it does for reusing it on App component

core/test/services/routing.spec.js
```diff
@@ -493,58 +493,6 @@ describe('Routing', () => {
         expectedPath
       );
     });
-
```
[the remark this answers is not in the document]
parostatkiem: It became way too hard to mock a component that could handle the "modal" (a promise) properly. We have two e2e tests for this feature, I think it is enough ;)

core/src/services/routing.js
```diff
@@ -322,7 +322,7 @@ export const handleRouteChange = async (path, component, node, config) => {
             });
             path &&
               handleRouteChange(path, component, node, config) &&
-              history.replaceState(window.state, '', newUrl);
+              history.pushState(window.state, '', newUrl);
```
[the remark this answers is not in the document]
parostatkiem: As far as I tested it, it works well, at least in scope of the problem Pete's discovered. If you can still find any issues with this, a separate ticket would be a good idea.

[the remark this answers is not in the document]
parostatkiem: * updated solution, it works better now (I'd say _perfectly_)

pekura:
Also the back button behaviour is weird. If I am on a dirty page and press the browser back button, I get the unsaved changes popup multiple times, I click no couple of times, then it is closed for good. Then I press the back button again and land on the login page.

jesusreal:
Changes:
- Support for external links
- Support for logout view
- Centralize as much as possible logic for unsaved changes modal:
  - Used once in authorization component for logout
  - Used once for Internal app navigation (client navigation +  core navbars navigation) through _handleInsideAppNavigation_ method (new name for _handleRouteClick_ method).
 - Used once for browser actions in _handleRouteChange_, namely clicks on history back and forward buttons and changing the url manually in the address bar. *Only here we are manipulating browser history.*
- Fix double click needed when navigating back through browser back button and clicking _No_ button in modal. For this user scenario we are adding some duplicated entries to the browser, but it could not be perfect and we considered this as the best approach

pekura:
It does not work properly in preserveView/goBack scenario: when you are on a page where there is a go back possibility with preserved view and you try to go back using linkManager().goBack() and the current page is dirty the content goes back to the preserved view, the URL stays the same, and the unsaved changes modal pops up. Independent of what you choose (Yes or No) the situation won't get better.

pekura:
Don't we want to have a backdrop for the unsaved changes modal?

jesusreal:
> Don't we want to have a backdrop for the unsaved changes modal?

Good point. This was not requested as part of this ticket, if needed I would do it in a separate task.

docs/luigi-client-api.md
```diff
@@ -47,35 +47,35 @@ Returns **[Object](https://developer.mozilla.org/docs/Web/JavaScript/Reference/G
...
--   reflect the route
--   keep the navigation state in Luigi
+-   Route inside micro front-ends.
+-   Reflect the route.
+-   Keep the navigation state in Luigi.
 
 ### navigate
 
-Navigates to the given path in the hosting Luigi application. Contains either a full absolute path or a relative path without a leading slash that uses the active route as a base. This is a standard navigation.
+Navigates to the given path in application hosted by Luigi. It contains either a full absolute path or a relative path without a leading slash that uses the active route as a base. This is the standard navigation.
```
kazydek:
```suggestion
Navigates to the given path in the application hosted by Luigi. It contains either a full absolute path or a relative path without a leading slash that uses the active route as a base. This is the standard navigation.
```

docs/luigi-client-api.md
```diff
@@ -126,15 +126,15 @@ Sends node parameters to the route, which are then used by the `navigate` functi
...
+// Can be chained with context setting functions such as:
 LuigiClient.linkManager.fromContext("currentTeam").withParams({foo: "bar"}).navigate("path")
 ```
 
 Returns **[linkManager](#linkmanager)** link manager instance.
 
 ### pathExists
 
-Checks if a path you can navigate to exists in the main application. You can use this helper method to perform actions such as conditional display of a DOM element like a button.
+Checks if the path you can navigate to exists in the main application. For example, you can use this helper method conditionally display a DOM element like a button.
```
kazydek:
```suggestion
Checks if the path you can navigate to exists in the main application. For example, you can use this helper method conditionally to display a DOM element like a button.
```

docs/luigi-client-api.md
```diff
@@ -189,8 +189,16 @@ Removes the loading indicator. Use it after calling [showLoadingIndicator()](#sh
...
-Adds a backdrop to block the top and side navigation. It is based on Fundamental UI Modal, which you can use in your micro front-end to achieve the same behavior.
+Adds a backdrop to block the top and side navigation. It is based on the Fundamental UI Modal, which you can use in your micro front-end to achieve the same behavior.
 
 ### removeBackdrop
 
 Removes the backdrop.
+
+### setDirtyStatus
+
+Makes current page dirty or not
```
kazydek:
```suggestion
Makes current page dirty or not.
```

kazydek: Could you explain in more details what is meant in this sentence? As for the periods at the end of the sentence and starting it with the verb without a subject I am against but added it here to keep the consistency in the whole text as I can see the rest starts and end similarly.

docs/luigi-client-api.md
```diff
@@ -189,8 +189,16 @@ Removes the loading indicator. Use it after calling [showLoadingIndicator()](#sh
...
+
+#### Parameters
+
+-   `isDirty` **[boolean](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Boolean)** tells if current page/component has any unsaved changes at the moment
```
kazydek:
```suggestion
-   `isDirty` **[boolean](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Boolean)** tells if the current page or component has any unsaved changes at the moment
```

docs/luigi-client-api.md
jesusreal: This method informs the main application that the current page in the iframe is dirty, like for example a page with form fields which has been modified from their original values.

docs/luigi-client-api.md
jesusreal: done

docs/luigi-client-api.md
jesusreal: done

docs/luigi-client-api.md
jesusreal: done

docs/luigi-client-api.md
jesusreal: done

docs/luigi-client-api.md
```diff
@@ -189,8 +189,16 @@ Removes the loading indicator. Use it after calling [showLoadingIndicator()](#sh
...
-Adds a backdrop to block the top and side navigation. It is based on Fundamental UI Modal, which you can use in your micro front-end to achieve the same behavior.
+Adds a backdrop to block the top and side navigation. It is based on the Fundamental UI Modal, which you can use in your micro front-end to achieve the same behavior.
 
 ### removeBackdrop
 
 Removes the backdrop.
+
+### setDirtyStatus
+
+This method informs the main application that the current view in the iframe has unsaved changes, like for example a view with form fields which have been edited but not yet submitted.
```
kazydek:
```suggestion
This method informs the main application that there are unsaved changes in the current view in the iframe. For example, that can be a view with form fields which were edited but not submitted.
```

jesusreal: done

PR #261, opened 2018-11-29
Fix the relative path bug

dariadomagala-sap:
**Description**

- Fix the relative path bug

**Related issue(s)**

### DIFF
--- a/core/src/services/routing.js
+++ b/core/src/services/routing.js
@@ -7,8 +7,7 @@ import {
   containsAllSegments,
   isIE,
   getConfigValueFromObject,
-  addLeadingSlash,
-  normalizePath
+  addLeadingSlash
 } from '../utilities/helpers';
 
 const iframeNavFallbackTimeout = 2000;
@@ -464,10 +463,14 @@ export const buildFromRelativePath = node => {
   if (node.parent && node.parent.pathSegment) {
     // use only this part of the current path that refers to the parent of the node (remove additional parts refering to the sibiling)
     // remove everything that is after the parents pathSegment 'parent/keepSelectedForChildren/something' -> 'parent'
-    const nodePathSegments = normalizePath(getNodePath(node.parent)).split('/');
-    const windowPathSegments = normalizePath(windowPath).split('/');
+    const nodePathSegments = trimLeadingSlash(getNodePath(node.parent)).split(
+      '/'
+    );
+    const windowPathSegments = trimLeadingSlash(windowPath).split('/');
     if (windowPathSegments.length > nodePathSegments.length) {
-      windowPath = nodePathSegments.join('/');
+      windowPath = windowPathSegments
+        .slice(0, nodePathSegments.length)
+        .join('/');
     }
   }
   return addLeadingSlash(concatenatePath(windowPath, node.link));


core/src/services/routing.js
```diff
@@ -455,12 +455,18 @@ export const navigateTo = async route => {
...
-  if (LuigiConfig.getConfigValue('routing.useHashRouting')) {
-    return window.location.hash + '/' + path;
-  } else {
-    return window.location.pathname + '/' + path;
+export const buildFromRelativePath = node => {
+  let windowPath = LuigiConfig.getConfigValue('routing.useHashRouting')
+    ? window.location.hash
+    : window.location.pathname;
+  if (node.parent && node.parent.pathSegment) {
+    windowPath = windowPath.substr(
```
maxmarkus: Even tough I am a bit confused what this code is doing, it works. Maybe you want to add a comment what you are achieving here, since it is quite hard to read.

core/src/services/routing.js
```diff
@@ -455,12 +455,20 @@ export const navigateTo = async route => {
   window.dispatchEvent(event);
 };
 
-export const buildFromRelativePath = path => {
-  if (LuigiConfig.getConfigValue('routing.useHashRouting')) {
-    return window.location.hash + '/' + path;
-  } else {
-    return window.location.pathname + '/' + path;
+export const buildFromRelativePath = node => {
```
pekura: I think it is safer to align the path by number of path segments. My suggestion would be:

```
export const buildFromRelativePath = node => {
  let windowPath = LuigiConfig.getConfigValue('routing.useHashRouting')
    ? getPathWithoutHash(window.location.hash)
    : window.location.pathname;
  if (node.parent && node.parent.pathSegment) {
    debugger;
    const nodePathSegments = getNodePath(node.parent).split('/');
    const windowPathSegments = windowPath.split('/');
    if (windowPathSegments.length > nodePathSegments.length - 1) {
      windowPath = windowPathSegments.slice(0, nodePathSegments.length - 1).join('/');
    }
  }
  return addLeadingSlash(concatenatePath(windowPath, node.link));
};
```

jesusreal: A small suggestion here, for consistency and avoiding issues with leading / trailing slashes in the future or between routing strategies, I would use helper methods to normalize `nodePathSegments` and `windowPathSegments` before applying split, something like:
```
const nodePathSegments = Helpers.normalizePath(getNodePath(node.parent)).split('/');
const windowPathSegments = Helpers.normalizePath(windowPath).split('/');
```
Also what is being checked in `if (windowPathSegments.length > nodePathSegments.length - 1)` is at least for me not straightforward to understand, maybe a variable with a meaningful name would be helpful.

pekura: Yeah, I removed the '-1', it was not necessary with properly prepared paths.

dariadomagala: I applied suggested changes :) 
The only difference is that instead of 
```windowPath = windowPathSegments.slice(0, nodePathSegments.length).join('/');```
I used 
```windowPath = nodePathSegments.join('/');```

pekura: I would rather stick to `windowPath = windowPathSegments.slice(0, nodePathSegments.length).join('/');` because of dynamic path segments. It might work now but in the future after a planned refactoring (because of mutating the navigation data structure that we do not want to happen) it will get broken for sure.

dariadomagala: You are right, I reverted my latest commit.

Issue #170, opened 2018-10-26
Improve Luigi Client API (-related) documentation 


bszwarc:
A task for improving Luigi Client (-related) documentation.
Documents to improve:
- https://github.com/kyma-project/luigi/tree/master/client should refer to the autogenerated luigi client api documentation
- https://github.com/kyma-project/luigi/blob/master/docs/luigi-client-api.md . Here we need some more general improvements (but let's wait for this one: https://app.zenhub.com/workspace/o/kyma-project/luigi/issues/155 to be closed to avoid merging conflicts).

kwiatekus:
Few improvements introduced within https://github.com/kyma-project/luigi/pull/197

PR #262, opened 2018-11-30
Add improvements to Luigi Client API comments and annotations

bszwarc:
**Description**

- Add improvements to Luigi API docs 

**Related issues**

See #170

### DIFF
--- a/client/luigi-client.js
+++ b/client/luigi-client.js
@@ -197,7 +197,7 @@
     },
     /**
      * Removes a context update listener.
-     * @param {string} id the ID that was returned by the `addContextUpdateListener` function
+     * @param {string} id the id that was returned by the `addContextUpdateListener` function
      * @memberof lifecycle
      */
     removeContextUpdateListener: function removeContextUpdateListener(id) {
--- a/client/luigi-client.js
+++ b/client/luigi-client.js
@@ -169,7 +169,7 @@
       return id;
     },
     /**
-     * Removes the init listener.
+     * Removes an init listener.
      * @param {string} id the id that was returned by the `addInitListener` function
      * @memberof lifecycle
      */
@@ -196,7 +196,7 @@
       return id;
     },
     /**
-     * Removes the context update listener.
+     * Removes a context update listener.
      * @param {string} id the ID that was returned by the `addContextUpdateListener` function
      * @memberof lifecycle
      */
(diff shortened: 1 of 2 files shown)


client/luigi-client.js
```diff
@@ -169,8 +169,8 @@
       return id;
     },
     /**
-     * Removes a init listener
-     * @param {string} id the id that was returned by `addInitListener`
+     * Removes the init listener.
```
maxmarkus: There can be multiple init listeners, the function removes the listener based on the id. 
The @param change is fine, `the` is ok here.

client/luigi-client.js
```diff
@@ -196,8 +196,8 @@
       return id;
     },
     /**
-     * Removes a context update listener
-     * @param {string} id the id that was returned by `addContextUpdateListener`
+     * Removes the context update listener.
```
maxmarkus: Same as above, it removes a specific listener, based on its id.

client/luigi-client.js
```diff
@@ -310,7 +310,7 @@
         },
 
         /**
-         * Sets the current navigation context, which is then used by the `navigate` function. This has to be a parent navigation context, it is not possible to go to child navigation contexts.
+         * Sets the current navigation context which is then used by the `navigate` function. This has to be a parent navigation context, it is not possible use the child navigation contexts.
```
dariadomagala: Shouldn't that be
```suggestion
         * Sets the current navigation context which is then used by the `navigate` function. This has to be a parent navigation context, it is not possible to use the child navigation contexts.
```

client/luigi-client.js
bszwarc: Thank you, updated.

client/luigi-client.js
bszwarc: Yes, you are right - updated this one.

client/luigi-client.js
bszwarc: Done.

client/luigi-client.js
```diff
@@ -169,8 +169,8 @@
       return id;
     },
     /**
-     * Removes a init listener
-     * @param {string} id the id that was returned by `addInitListener`
+     * Removes an init listener.
+     * @param {string} id the id that was returned by the `addInitListener` function
```
dariadomagala: `id` or `ID`? There is an inconsistency compared to the line 200

bszwarc: let's stay with `id`.

bszwarc:
Fixes #170

PR #266, opened 2018-12-04
Luigi client creates extra entry in browser navigation history

pekura:
Fixes following issues:

1) Hash-based navigation not working in the Firefox browser.
2) Context switcher dose not get closed on blur or selection in the Firefox browser.
3) Three dot top navigation menu on narrow screens is displayed even if there are no visible (non-hidden from nav) top navigation entries.
4) Top nav links from the three dots menu (visible on narrow screens) do not work properly - content gets changed but the browser URL stays.
5) In the Edge browser the content does not get rendered at all.
6) Main app title vertical alignment is wrong.

Issues 2 to 6 are independent of the routing mode (hash/path).


maxmarkus:
On screens below 360px width, I have seen horizontal scrollbar. We could set a different width for the context switcher on very narrow screens.

maxmarkus:
LGTM

PR #271, opened 2018-12-05
POC : Add extendedConfiguration transpilation to use ES6

maxmarkus:
Implemented webpack config that runs prior to build tasks and in parallel of start tasks in order to allow quick development. `extendedConfiguration.js` file position has been changed to `./config/`.


kwiatekus:
Propabably as afollow up : (consult @hardl )
We want to check if angular CLI is able to bundle the config as well.
So that with npm start (ng serve) we could transpile both the app and config and watch also for changes in the config.

maxmarkus:
During investigation I found a way to extend angular-cli's with webpack config, I used it to remove our old webpack config. 
After ticket https://github.com/kyma-project/luigi/issues/287 is merged, we can also try to switch the separate config from this ticket to multi-entrypoint and include it in the webpack-extra.

maxmarkus:
Unfortunately it is not possible to provide another _*.js_ target in angular-cli's webpack config and therefore, the compilation spits out a webpack module instead of a babelified config. We should stick to the solution for now unless we want to write a custom parser.

jesusreal:
Fixed on https://github.com/kyma-project/luigi/pull/288

docs/application-setup.md
```diff
@@ -47,8 +47,9 @@ $ npm init -y
 $ npm i -S @kyma-project/luigi-core@latest @kyma-project/luigi-client@latest fiori-fundamentals@latest
```
jesusreal: The angular example is not working anymore, I get _Ups.. Looks like Luigi was not configured. Please use Luigi.setConfig(config) function to configure Luigi._. Please check this and all the examples.

maxmarkus:
Solved in #313

PR #285, opened 2018-12-10
Split main js files

### LINKED ISSUE #164: Split routing.js and navigation.js to smaller files

parostatkiem:
**Description**

- Split `routing.js` and `navigation.js` into smaller files
- Mirror test hierarchy to actual code files hierarchy
- Rename all tests to match one convention 

**Related issue(s)**

Resolves #164


core/src/Authorization.html
```diff
@@ -10,10 +10,15 @@
 
 <script type="text/javascript">
   import { LuigiConfig } from './services/config.js';
-  import { isFunction, isPromise, deepMerge, getConfigValueFromObject } from './utilities/helpers.js';
+  import {
+    isFunction,
+    isPromise,
+    deepMerge,
+    getConfigValueFromObject
+  } from './utilities/helpers/helpers.js';
```
maxmarkus: I would love to see `helpers.js` being renamed to something more meaningful. `basic-helpers` or maybe something better?

core/src/Authorization.html
```diff
@@ -104,26 +125,28 @@
...
           logoutCallback();
           window.location.href = idpProviderInstance.settings.logoutUrl;
-        };
-      },
+        }
+      }
     },
-    components: {},
+    components: {}
     // store: () => store
```
maxmarkus: Feel free to remove obsolete things when cleaning up. Guess this `store` is there since ages and never used.

core/src/Authorization.html
parostatkiem: `generic-helpers`?

maxmarkus: ok

maxmarkus:
Since we talked already, I am not sure if it makes sense to refactor everything again now to regain module based imports similar to `import * as GenericHelpers from helpers/generic-helpers.js`. We will do it later when we refactor to class based helpers. Please just consider it in future refactorings.

core/src/utilities/helpers/navigation-helpers.js
```diff
@@ -0,0 +1,61 @@
...
+      context[p] = addition[p];
+    }
+  }
+  if (navigationContext) {
+    context.parentNavigationContexts.unshift(navigationContext);
+  }
+  return context;
+};
+
+export const groupBy = (nodes, property) => {
```
maxmarkus: We could rename this fn to groupNodesByProperty  or groupNodesBy, since the function looks like you could just group anything.

core/src/utilities/helpers/routing-helpers.js
```diff
@@ -0,0 +1,196 @@
...
+
+export const findViewGroup = node => {
+  if (node.viewGroup) {
+    return node.viewGroup;
+  } else if (node.parent) {
+    return findViewGroup(node.parent);
+  }
+};
+const defaultContentViewParamPrefix = '~';
+export const getContentViewParamPrefix = () => {
```
maxmarkus: defaultContentViewParamPrefix could be placed inside getContentViewParamsPrefix function, since it is only used there.

core/src/utilities/helpers/routing-helpers.js
```diff
@@ -0,0 +1,196 @@
...
+  if (lastElement.defaultChildNode && pathExists) {
+    return lastElement.defaultChildNode;
+  } else if (children && children.length > 0) {
+    return children[0].pathSegment;
+  } else {
+    return '';
+  }
+};
+
+export const isExistingRoute = function(path, pathData) {
```
maxmarkus: Transform to arrow function, please.

core/src/utilities/helpers/routing-helpers.js
```diff
@@ -0,0 +1,196 @@
...
+import { escapeRegExp, getUrlWithoutHash } from './generic-helpers';
+import { LuigiConfig } from '../../services/config';
+import { getModifiedPathname } from '../../services/routing';
+
+export const getLastNodeObject = pathData => {
+  const lastElement = [...pathData.navigationPath].pop();
+  return lastElement ? lastElement : {};
+};
+
+export const getDefaultChildNode = async function(pathData) {
```
maxmarkus: Transform to arrow function, please.

core/src/utilities/helpers/routing-helpers.js
parostatkiem: I vote against. Now it's easier to find the constant. It would be even better to move all constants to the up. I can negotiate 😄

jesusreal:
As discussed in a call, please apply the following changes:
- [ ] imports should import the whole object, not individual functions. With this we achieve shorter imports, is easier to identify where each function belongs to when reading the code, and we come closer to the approach of using es6 classes.
- [ ] Add a description at the beginning of each service and helper, explaining which code should come on each file. This will help to understand why we are split the logic this way and will make easier to decide where to put new code.
- [ ] Please move iframe-related logic from routing.js to a iframe.js (new file), and remove routing-handlers.js, placing its logic in routing.js. With that we don't introduce the concept of 'handlers' files additionally to already existing 'service' and 'helpers'. We should only introduce new filetypes if needed, otherwise it can become confusing which logic should be placed in each file.

PR #286, opened 2018-12-10
Add alert about resource quotas limits exceeded

dariadomagala:
**Description**

- Create reusable alert component
- Add alert about resource quotas limits exceeded

**Related issue(s)**

### DIFF
--- a/core/src/Alert.html
+++ b/core/src/Alert.html
@@ -1,0 +1,63 @@
+{#if alert}
+  {#if alert.msg === "luigi.route-not-found"}
+    <div class="fd-alert fd-alert--error fd-alert--dismissible" role="alert" id="j2ALl423">
+      <button on:click="set({alert:false})" class="fd-alert__close" aria-controls="j2ALl423" aria-label="Close"></button>
+      {alert.message} {alert.link}
+    </div>
+  {/if}
+  {#if alert.msg === "luigi.limit-exceeded"}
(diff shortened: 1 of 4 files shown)


core/src/Alert.html
```diff
@@ -0,0 +1,63 @@
...
+    <div class="fd-alert fd-alert--error fd-alert--dismissible" role="alert" id="j2ALl423">
+      <button on:click="set({alert:false})" class="fd-alert__close" aria-controls="j2ALl423" aria-label="Close"></button>
+        <div>Error! The following resource quota limit has been exceeded by the given resource:</div>
+        {#each alert.exceededLimits as limit}
+          {#each limit.affectedResources as affectedResource}
+            <div>- '{limit.resourceName}' by '{affectedResource}' ({limit.quotaName})</div>
+          {/each}
+        {/each}
+        <div>
+          See <a on:click="goTo('/environments/'+alert.environment+'/resources')">Resources Configuration</a> for details.
```
kwiatekus: We cant have console specific code here.
The problem with this case if the link *Resources Configuration* link that should be part of the alert message. 
Challage : How to use generic alert component in luigi and display an alert from console that includes a link?

kwiatekus: screen shot 2018-12-07 at 12 29 31

kwiatekus:
We agreed during sprint review, that for now we will display the alert in the views themselves (as in old console)

Please tweak the positioning of the alert (left)

PR #291, opened 2018-12-12
Add `refreshContextSwitcher` to LuigiClient

### LINKED ISSUE #290: Implement ability to refresh env list in ContextSwitcher
**Description**
~Luigi Client should provide such a possibility.~

Luigi core api needs a function to refresh options of context switcher.

The microfrontend should then use a custom event defined by the luigi core app (console in this example) to fire a semantic change event, e.g. envChanged or so.

**Reasons**

**Attachments**

parostatkiem:
**Description**
#### To see how it works, switch your Console repo to the branch from [THIS PR](https://github.com/jesusreal/console/pull/26)

- Add `refreshContextSwitcher` method to LuigiClient
- Catch certain _postMessage_ in ContextSwitcher component and refresh environment list

**Related issue(s)**

Resolves #290

### DIFF
--- a/client/luigi-client.js
+++ b/client/luigi-client.js
@@ -431,6 +431,15 @@
          */
         removeBackdrop: function removeBackdrop() {
           window.parent.postMessage({ msg: 'luigi.remove-backdrop' }, '*');
+        },
+        /**
+         * Removes the backdrop.
+         */
+        refreshContextSwitcher: function refreshContextSwitcher() {
+          window.parent.postMessage(
+            { msg: 'luigi.ux.refreshContextSwitcher' },
+            '*'
+          );
         }
       };
     }
--- a/client/luigi-client.js
+++ b/client/luigi-client.js
@@ -433,7 +433,7 @@
           window.parent.postMessage({ msg: 'luigi.remove-backdrop' }, '*');
         },
         /**
-         * Removes the backdrop.
+         * Reloads the environment list in the Context Switcher
          */
         refreshContextSwitcher: function refreshContextSwitcher() {
           window.parent.postMessage(
--- a/core/src/navigation/ContextSwitcher.html
+++ b/core/src/navigation/ContextSwitcher.html
@@ -53,20 +53,24 @@
 
 <script type="text/javascript">
   import { ContextSwitcherHelpers } from './services/context-switcher.js';
-  import { LuigiConfig } from '../services/config.js'
+  import { LuigiConfig } from '../services/config.js';
   import * as Helpers from '../utilities/helpers.js';
   import * as Routing from '../services/routing.js';
 
   export default {
     async oncreate() {
-      const contextSwitcherConfig = LuigiConfig.getConfigValue('navigation.contextSwitcher');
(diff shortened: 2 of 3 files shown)


client/luigi-client.js
```diff
@@ -431,6 +431,15 @@
          */
         removeBackdrop: function removeBackdrop() {
           window.parent.postMessage({ msg: 'luigi.remove-backdrop' }, '*');
+        },
+        /**
+         * Reloads the environment list in the Context Switcher
```
bszwarc:
```suggestion
         * Reloads the environment list in the Context Switcher.
```

and I would leave the environment for now :D if there is a need for renaming, we will rename it later.

maxmarkus:
I know it is not super easy to achieve, but would be great to see an e2e test for that feature.

Implementation suggestion: Enhancing the example to also read from session storage and if there is something, add it to the list. With cypress you store some value that the refresh will pick up.

pekura:
After you add a new environment in the sample app and switch to it it looks like this: 
bildschirmfoto 2018-12-18 um 14 39 40 - not nice.

core/examples/luigi-sample-angular/src/app/project/miscellaneous2/miscellaneous2.component.html
```diff
@@ -1,8 +1,8 @@
 <section class="fd-section">
   <div class="fd-section__header">
-    <h1 class="fd-section__title">Miscellaneous 2</h1>
+    <h1 class="fd-section__title">A normal Node</h1>
```
pekura: Normal node does not sound right, maybe Sample node? Or, do we need it at all?

core/examples/luigi-sample-angular/src/app/services/helpers.ts
```diff
@@ -3,3 +3,7 @@ export const toTitleCase = (str: string = ''): string => {
     return txt.charAt(0).toUpperCase() + txt.substr(1).toLowerCase();
   });
 };
+
+export const slugify = (str: string = ''): string => {
+  return str.toLowerCase().replace(/[^A-Z0-9]+/gi, '-');
```
pekura: This will filter out all Umlaut-like characters which is not intended I guess.

core/examples/luigi-sample-angular/src/app/project/miscellaneous2/miscellaneous2.component.html
maxmarkus: Removed completely. Misc2 is gone.

core/examples/luigi-sample-angular/src/app/services/helpers.ts
maxmarkus: Intended, as it is used for pathSegment only in the demo app.

pekura: Yes, sorry I misinterpreted the usage of this function.

pekura:
Let's keep this PR opened until we have a proper discussion about the context chooser. Let's talk about it after the holiday season.

PR #299, opened 2018-12-18
It is not possible to have a root node with empty path segment and a view

pekura:
It's not possible to set the root node's path segment as an empty string or a slash - fix.


pekura:
OK, maybe the navigation structure you provide is not as expected. This only works if you provide the nodes not as an array but as a single root node object (with children) where pathSegment is set to empty string. Maybe you provide an array (as we always had it like that in our examples, but now you can actually specify the tree from the root node (which is only one and that is why it may not be an array)).

dariadomagala:
@pekura You are right. Unfortunately I need an additional node in the top-navigation, so I need to have two main nodes, and because of that it won't work for me.

core/src/navigation/services/navigation.js
```diff
@@ -2,18 +2,29 @@
...
+    return [{}];
   }
   try {
+    let rootNode;
     const topNavNodes = await rootNavProviderPromise;
-    rootNode.children = topNavNodes;
+    if (GenericHelpers.isObject(topNavNodes)) {
+      rootNode = topNavNodes;
+    } else {
+      rootNode = {};
```
jesusreal: this could be one line: `rootNode = { children : topNavNodes }`

core/src/utilities/helpers/generic-helpers.js
```diff
@@ -97,18 +97,22 @@ export const getNegatedBoolString = str => {
...
+    !targetPathSegments ||
+    !targetPathSegments.length
+  ) {
     console.error(
       'Ooops, seems like the developers have misconfigured something'
     );
     return false;
   }
-
+  const mandatorySegmentsUrl = trimTrailingSlash(sourceUrl.split('?')[0]);
```
jesusreal: It would be great if you could explain this part to me, maybe with a short call, because I don't understand for which case we needed to make the slicing conditional

jesusreal: As discussed, it would be good to express the condition differently if possible or to leave a comment in the code to make clear why we need that

PR #301, opened 2018-12-18
Fix dropdowns behavior on click events

dariadomagala:
**Description**

- Fix dropdowns behavior on click events

**Related issue(s)**

### DIFF
--- a/core/src/Authorization.html
+++ b/core/src/Authorization.html
@@ -1,11 +1,13 @@
 <ul class="fd-menu__list">
-  <li>
     {#if isLoggedIn}
-      <a aria-label="Logout" on:click=onLogoutClick() class="fd-menu__item">Logout</a>
+      <li on:click=onLogoutClick()>
+        <a aria-label="Logout" class="fd-menu__item">Logout</a>
+      </li>
     {:else}
-      <a aria-label="Login" on:click=startAuthorization() class="fd-menu__item">Login</a>
+      <li on:click=startAuthorization()>
+        <a aria-label="Login" class="fd-menu__item">Login</a>
+      </li>
     {/if}
-  </li>
 </ul>
 
 <script type="text/javascript">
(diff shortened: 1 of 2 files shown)


pekura:
It still does not work perfectly - in chrome when you click on a bottom part of the entry (somewhere in the last quarter of the highlighted area, below the text) it does not get activated - in the context chooser as well as the Logout button. In firefox the context chooser works fine but the Logout button has the same problem as in chrome.

core/src/navigation/TopNav.html
```diff
@@ -163,19 +168,24 @@
...
-        );
+        const dropDownState = GenericHelpers.getNegatedBoolString(dropDownStates[name]);
+        this.closeAllDropdowns();
+        dropDownStates[name] = dropDownState;
         this.set({
           dropDownStates
         });
       },
-      closeDropdown(name) {
+      closeAllDropdowns() {
```
jesusreal: I would refactor this method, also in `ContextSwitcher` component, to the following:
```
closeAllDropdowns() {
  const dropDownStates = this.get().dropDownStates || {};
  Object.keys(dropDownStates).forEach(k => { dropDownStates[k] = 'false' });
  this.set({dropDownStates});
}
```

core/src/navigation/TopNav.html
```diff
@@ -1,86 +1,91 @@
...
-    </div>
-    <div class="fd-shellbar__group fd-shellbar__group--end">
+  <div class="fd-shellbar__group fd-shellbar__group--start">
+    <LogoTitle />
+  </div>
+  <div class="fd-shellbar__group fd-shellbar__group--end">
+    <div class="fd-shellbar__actions">
 
+      {#if !authorizationEnabled || isLoggedIn}
+        <ContextSwitcher {dropDownStates} bind:dropDownStates />
```
jesusreal: Not something you introduced, but `{dropDownStates}` is not needed here.

core/src/navigation/TopNav.html
dariadomagala: `dropDownStates` is an empty object at the beginning. Adding the _foreach_ loop without checking if the array of keys is not empty, would cause errors.

PR #305, opened 2018-12-21
When adding a listener via Luigi client API, call only the listener being added

jesusreal:
**Description**

- When adding an init or context update listener via Luigi Client API, call only the listener being added.

**Related issue(s)**

Fixes https://github.com/kyma-project/console/issues/429

### DIFF
--- a/client/luigi-client.js
+++ b/client/luigi-client.js
@@ -59,16 +59,23 @@
   }
 
   /**
+   * Simple function check.
+   * @private
+   * @param item mixed
+   * @returns {boolean}
+   */
+  function isFunction(item) {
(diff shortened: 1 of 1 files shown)


client/luigi-client.js
```diff
@@ -58,17 +58,24 @@
...
   }
 
+  /**
+   * Simple function check.
+   * @private
+   * @param item mixed
+   * @returns {boolean}
+   */
+  function isFunction(item) {
+    return item === 'function';
```
kwiatekus: shouldn't it be
```
return typeof item == 'function'
```

Issue #281, opened 2018-12-07
Going to “absolutely” wrong path in path routing doesn’t redirect to the root


dariadomagala:
**Expected result**

After going to an "absolutely" wrong path when using path routing (e.g. myapp.com/non-existing-path), Luigi should show an error message, and redirect user to the root.

**Actual result**

Going to a wrong path shows an error message, but doesn't redirect to the root.

**Steps to reproduce**

* open a Luigi app with a path routing
* go to a non-existing path
* error is shown but no redirection

parostatkiem:
PR: https://github.com/kyma-project/luigi/pull/310

PR #310, opened 2018-12-31
Improve 404 handling

parostatkiem:
**Description**
## `lerna bootstrap` required!

- Alert has a `ttl` property which lets it _survive_ **n** redirects (set to 1)
- Microfrontend developer can define a custom behaviour (function) when page is not found (**404**)
- `Could not map the exact target...` error appears when user tries to enter a route with a defaultChildNode at the end **and** he's redirected to that node.
- Migrate to **Babel 7.0** and **@babel/preset-env** to be able to use spread operator
- Design a fancy, isolated 404 page (uncomment `extendedConfiguration.js::583` to be able to see it)

**Related issue(s)**

Resolves #310, #281


core/src/services/routing.js
```diff
@@ -187,35 +187,41 @@ export const handleRouteChange = async (path, component, node, config) => {
...
         navigateTo(`${pathUrl ? `/${pathUrl}` : ''}/${defaultChildNode}`);
       } else {
-        const alert = {
-          message: 'Could not find the requested route',
-          link: pathUrlRaw
-        };
-
-        component.set({ alert });
-        navigateTo('/');
+        if (defaultChildNode && pathData.navigationPath.length > 2) {
```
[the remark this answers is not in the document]
parostatkiem: 1. In case (probably impossible) `pathData.navitationPath` was empty, there will be no errors.
2. I've just changed it to `1` because `overview` is a defaultChild for `/`. If it was set to _>0_ or not present at all, "Could not find exact route" error would appear instead of "Could not find requested route" due to automatic redirection to `/`. 

Thanks for noticing `2` was the wrong number 👍

maxmarkus:
> Design a fancy, isolated 404 page (uncomment extendedConfiguration.js::583 to be able to see it)
Currently the fancy isolated 404 page is enabled, do we want to enable it by default?

core/examples/luigi-sample-angular/src/assets/404.html
```diff
@@ -0,0 +1,87 @@
...
+</style>
+</head>
+<body>
+<header class="fd-page__header">
+      <h1 class="fd-action-bar__title">
+          Ooooooooops :(
+      </h1>
+      <h2 class="fd-action-bar__description">Yes, you guessed right, it's a 
+        <strong class="error-number--real">404</strong>
+        <strong class='error-number--fake'>404</strong>
```
maxmarkus: --spacer instead of fake?

core/src/App.html
```diff
@@ -206,7 +206,18 @@
           navigateOk: null,
           builderCompatibilityMode: Boolean(window.builderCompatibilityMode)
         };
-
+        window.addEventListener('popstate', async e => {
+          const alert = this.get().alert;
+          if (alert && alert.ttl !== undefined) {
```
maxmarkus: I would use `typeof number` here, just to be save.

parostatkiem: I'd love to use `isInteger()` but Internet Explorer's compatibility blocks me from doing this :(

core/examples/luigi-sample-angular/src/assets/extendedConfiguration.js
```diff
@@ -320,6 +320,11 @@ var projectsNavProviderFn = function(context) {
   });
 };
 
+var customPageNotFoundHandler = function(wrongPath, wasAnyPathFitted) {
```
dariadomagala: Maybe it would be good to add to documentation that this function might receive `wrongPath` and `wasAnyPathFitted`?

core/src/services/routing.js
```diff
@@ -264,3 +265,32 @@ export const handleRouteClick = node => {
...
+    return;
+  }
+
+  const alert = {
+    message: isAnyPathMatched
+      ? 'Could not map the exact target node for the requested route'
+      : 'Could not find the requested route',
+    link: notFoundPath,
+    ttl: pathToRedirect === '/' ? 2 : 1 //how many redirections the alert will 'survive'.
+    //Redirecting to '/' causes another redirection to '/overview', that's why TTL needs to be bigger
```
dariadomagala: In our case it's ok, because we don't have a view for '/', so it redirects to '/overview', but what if someone has a view on '/' and doesn't have this additional redirection?

parostatkiem: 🤔

core/src/services/routing.js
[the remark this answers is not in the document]
jesusreal: I would rather create a variable out of `pathData.navigationPath.length > 1` to make clear what are we checking here through the variable name.

docs/navigation-configuration.md
```diff
@@ -217,6 +217,9 @@ You can configure the way Luigi tackles routing in your application in the **Rou
 - **useHashRouting** defines either hash-based (`url.com/#/yourpath`) or path-based (`url.com/yourpath`) routing.
 - **nodeParamPrefix** sets the prefix character when using the `LuigiClient.linkManager().withParam()` function, which provides a way to simply attach query parameters to the view URL for activities such as sorting and filtering. The URL contains the parameters to allow deep linking. If you want to use a different character prefix, define yours here. The default character is `~`.
 - **skipRoutingForUrlPatterns** defines regex patterns to be skipped by the router when listening for path changes. This parameter is used for excluding **redirect_uri** parameters. Default patterns are `[/access_token=/, '/id_token=/]`.
+- **pageNotFoundHandler** is a function defining custom behaviour when the 404 (page not found) error occured. By default it's handled by Luigi. Leave its body empty if you have an external 404 handling. It takes two parameters:
```
```suggestion
- **pageNotFoundHandler** is a function defining custom behavior when the 404 (page not found) error occurs.  Luigi handles it by default. Leave its body empty if you have an external 404 handling. This function takes the following parameters: 
```

docs/navigation-configuration.md
```diff
@@ -217,6 +217,9 @@ You can configure the way Luigi tackles routing in your application in the **Rou
 - **useHashRouting** defines either hash-based (`url.com/#/yourpath`) or path-based (`url.com/yourpath`) routing.
 - **nodeParamPrefix** sets the prefix character when using the `LuigiClient.linkManager().withParam()` function, which provides a way to simply attach query parameters to the view URL for activities such as sorting and filtering. The URL contains the parameters to allow deep linking. If you want to use a different character prefix, define yours here. The default character is `~`.
 - **skipRoutingForUrlPatterns** defines regex patterns to be skipped by the router when listening for path changes. This parameter is used for excluding **redirect_uri** parameters. Default patterns are `[/access_token=/, '/id_token=/]`.
+- **pageNotFoundHandler** is a function defining custom behaviour when the 404 (page not found) error occured. By default it's handled by Luigi. Leave its body empty if you have an external 404 handling. It takes two parameters: 
+  - **wrongPath**(string): the path that user tried to navigate to
+  - **wasAnyPathFitted**(bool): is true if Luigi managed to fit a valid path which means _wrongPath_ was only "partly wrong". Otherwise it is false.
```
```suggestion
  - **wasAnyPathFitted**(bool): it is true if Luigi managed to fit a valid path which means **wrongPath** was only partially wrong. Otherwise it is false.
```

PR #317, opened 2019-01-04
Improve defaultChildNode

parostatkiem:
**Description**

- Improve defaultChildNode function
- add unit test

**Related issue(s)**

### DIFF
--- a/core/src/utilities/helpers/routing-helpers.js
+++ b/core/src/utilities/helpers/routing-helpers.js
@@ -23,8 +23,9 @@ export const getDefaultChildNode = async pathData => {
 
   if (lastElement.defaultChildNode && pathExists) {
     return lastElement.defaultChildNode;
-  } else if (children && children.length > 0) {
-    return children[0].pathSegment;
+  } else if (children && children.length) {
+    const validChild = children.find(child => child.pathSegment);
+    return validChild ? validChild.pathSegment : '';
   } else {
     return '';
   }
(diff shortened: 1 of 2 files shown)


core/src/utilities/helpers/routing-helpers.js
```diff
@@ -23,8 +23,9 @@ export const getDefaultChildNode = async pathData => {
 
   if (lastElement.defaultChildNode && pathExists) {
     return lastElement.defaultChildNode;
-  } else if (children && children.length > 0) {
-    return children[0].pathSegment;
+  } else if (children && children.length) {
+    const validChild = children.find(child => child.pathSegment);
```
pekura: If a node (except for the root one) has no pathSegment it is a clear misconfiguration. We should rather check this when loading the navigation configuration and either fail or remove this node from the tree. 
For the default child behaviour we should rather maybe check if the view Url is set and choose first node with viewUrl set?

parostatkiem: Ok, I'll do the part with checking `viewUrl` and `externalLink`. Let's leave the other part for a separate task.

core/src/utilities/helpers/routing-helpers.js
```diff
@@ -23,8 +23,11 @@ export const getDefaultChildNode = async pathData => {
 
   if (lastElement.defaultChildNode && pathExists) {
     return lastElement.defaultChildNode;
-  } else if (children && children.length > 0) {
-    return children[0].pathSegment;
+  } else if (children && children.length) {
+    const validChild = children.find(
```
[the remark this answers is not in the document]
parostatkiem: I like it and I'll apply it 👍

PR #319, opened 2019-01-07
Implement anonymous content feature

hardl:


core/src/utilities/helpers/navigation-helpers.js
```diff
@@ -1,13 +1,22 @@
...
   parentNode,
   currentContext
 ) => {
-  if (LuigiConfig.isAuthorizationEnabled() && !AuthHelpers.isLoggedIn())
-    return false;
+  if (LuigiConfig.isAuthorizationEnabled()) {
+    const loggedIn = AuthHelpers.isLoggedIn();
+    const anon = nodeToCheckPermissionFor.anonymousAccess;
+
+    if (loggedIn && anon === "exclusive" ||
```
parostatkiem: What is going on with `true` and `"exclusive"` values for _anon_? Please provide some documentation to this setting.

hardl: documentation is covered in a separate ticket https://github.com/kyma-project/luigi/issues/296

PR #325, opened 2019-01-08
Webpack for LuigiClient

parostatkiem:
**Description**

### `lerna bootstrap` required in `/client`

- bundle LuigiClient with Babel using `preset-env`
- add 2 new commands for _/client_: `npm run bundle` and `npm run bundle-develop`
- remove manual UMD wrapper from the source file (it's now added by Webpack in slightly changed form)
- source file code *is not changed* (except removing UMD part). We should introduce class-approach and other things in a separate task. For now, an object is exported. It's up to us how that object is created.

**Related issue(s)**


client/package.json
```diff
@@ -2,8 +2,10 @@
   "name": "@kyma-project/luigi-client",
```
dariadomagala: Please change the name of this package. It won't be the `@kyma-project/luigi-client` package anymore, the transpiled  version will be.

client/public/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+Luigi is a micro front-end orchestration framework written in Svelte. Use Luigi Core to implement the core application, and Luigi Client to ensure communication between the application and the micro front-end it hosts. 
+
+For details on Luigi Client, see [this](https://github.com/kyma-project/luigi/tree/master/client) document.
+
+For details on Luigi Core, see [this](https://github.com/kyma-project/luigi/tree/master/core) document.
+
+If you want to try Luigi out, see the [examples](https://github.com/kyma-project/luigi/tree/master/core/examples).
+
+For documentation on Luigi Core, see [Luigi documentation](https://github.com/kyma-project/luigi/tree/master/docs).
```
bszwarc: Maybe we can change it to:

`For more documentation on Luigi framework, see [Luigi documentation](https://github.com/kyma-project/luigi/tree/master/docs). ` 

because it is not only about Luigi Core. 🤔

client/package.json
```diff
@@ -1,9 +1,12 @@
...
-  "main": "luigi-client.js",
+  "private": true,
+  "description": "luigi-client development version, only the public folder is being published to npm as a package",
+  "main": "public/luigi-client.bundle.js",
   "scripts": {
+    "bundle": "webpack --mode production",
+    "bundle-develop": "webpack --mode development --watch",
     "test": "echo \"Error: no test specified\" && exit 1",
     "docu": "npm run docu:validate && npm run docu:generate:section",
     "docu:generate:new-file": "documentation build luigi-client.js -f md --markdown-toc=false -o ../docs/luigi-client-api.md",
```
maxmarkus: Documentation is not working (path references and some babel errors)

parostatkiem:
The output file is now `client/luigi-client.js` to keep backwards compatibility. The source file is `client/src/luigi-client.js` (of course it can be extended to many files via `import).

client/.npmignore
```diff
@@ -0,0 +1 @@
+/luigi-client.js
```
dariadomagala: But.. we want to publish `luigi-client.js`. We don't want to publish `/src`.

client/package.json
```diff
@@ -16,6 +16,7 @@
   ],
   "main": "luigi-client.js",
   "scripts": {
+    "install": "npm run bundle",
```
dariadomagala: What is this script for?

docs/luigi-client-api.md
```diff
@@ -205,4 +211,4 @@ This method informs the main application that there are unsaved changes in the c
 
 #### Parameters
 
--   `isDirty` **[boolean](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Boolean)** indicates if there are any unsaved changes on the current page or in the component
+-   `isDirty` **[boolean](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Boolean)** tells if there are any unsaved changes on the current page or component
```
bszwarc: Revert to the original version in the js file:

```
`isDirty` **[boolean](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Boolean)** indicates if there are any unsaved changes on the current page or in the component

```

PR #327, opened 2019-01-09
Auto formatting for svelte

parostatkiem:
**Description**
Works well with HTML, JS and CSS parts of Svelte files.

- update Prettier module
- include `*.html` in pre-commit Prettier hook

**Related issue(s)**


package.json
```diff
@@ -23,5 +23,8 @@
       "pre-commit": "./scripts/apply-prettier.sh && ./scripts/remove-test-prefixes.sh",
       "pre-push": "lerna run prepush"
     }
+  },
+  "scripts": {
+    "prepush": " ./scripts/apply-prettier.sh && ./scripts/remove-test-prefixes.sh"
```
y-kkamil: Why duplicate scripts that are part of pre-commit?

y-kkamil:
@dariadomagala true. i think applying vue parser to all .html files resolves the issue as it should format <script> tag contents and leave svelte-specific syntax intact. i will add it

PR #332, opened 2019-01-10
Update luigi version in core, client, and angular example app

jesusreal:
**Description**

- Update luigi version in core, client, and angular example app

**Related issue(s)**


kwiatekus:
should'nt core/examples/luigi-sample-angular/package-lock.json also be modified?

PR #342, opened 2019-01-11
Ignore path params in context switcher label

parostatkiem:
**Description**

- Ignore path params in context switcher label
- Unit tests of the new feature

**Related issue(s)**
Fixes https://github.com/kyma-project/luigi/issues/335

### DIFF
--- a/core/src/navigation/services/context-switcher.js
+++ b/core/src/navigation/services/context-switcher.js
@@ -66,8 +66,8 @@ export const ContextSwitcherHelpers = {
     const selectedId = currentPath
       .replace(parentNodePath, '')
       .split('/')[0]
-      .split(nodeParamPrefix)[0] //ignore everythin after nodeParamPrefix
-      .split('?')[0]; //ignore everythin after '?' in case nodeParamPrefix was not used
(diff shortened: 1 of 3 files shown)


core/src/navigation/services/context-switcher.js
```diff
@@ -62,7 +63,11 @@ export const ContextSwitcherHelpers = {
     }
 
     // we are inside the context switcher base path
-    const selectedId = currentPath.replace(parentNodePath, '').split('/')[0];
+    const selectedId = currentPath
+      .replace(parentNodePath, '')
+      .split('/')[0]
+      .split(nodeParamPrefix)[0] //ignore everything after nodeParamPrefix
```
pekura: I would remove .split(nodeParamPrefix)[0] and just let the .split('?')[0] there. There is no rule that forbids nodeParamPrefix to be a character not allowed in a path segment.

parostatkiem: Would you like to see this?
image

pekura: yes!

parostatkiem: Allright then...

PR #343, opened 2019-01-11
Add docs and examples for anonymous content

dariadomagala:
**Description**

- Add docs and examples for anonymous content

**Related issue(s)**


docs/navigation-configuration.md
```diff
@@ -248,6 +248,7 @@ The node parameters are as follows:
...
 - **viewGroup** defines a group of views in the same domain sharing a common security context. This improves performance through reusing the frame. Use **viewGroup** only for the views that use path routing internally.
 - **icon** is the name of an icon from the [OpenUI](https://openui5.hana.ondemand.com/1.40.10/iconExplorer.html) or a custom link (relative or absolute) to an image displayed next to the Node label in the side navigation or instead of the label in the top navigation.
+- **anonymousAccess** defines if the node should be visible for the anonymous user. If **anonymousAccess** is set to _true_, all users will see the node, if it's set to _exclusive_, the node will be visible only for users that are not logged in. It is not propagated to the node's children and is disabled by default.
```
```suggestion
- **anonymousAccess** defines if the node should be visible for the anonymous user. If set to `true`, all users see the node, while if set to `exclusive`, only the users that are not logged in can see it. The property is not propagated to the node's children and is disabled by default.
```

PR #349, opened 2019-01-15
Fix preserve views node params

dariadomagala:
**Description**

- Fix preserve views node params

**Related issue(s)**
Fixes : https://github.com/kyma-project/luigi/issues/337


core/src/App.html
```diff
@@ -115,8 +115,10 @@
     if (data.params.preserveView) {
       const pv = component.get().preservedViews;
       const nextPath = buildPath(component, data.params);
+      const pathUrlRaw = component.get().pathUrlRaw;
```
parostatkiem: The solution works well 👍 
I can see you're calling `component.get()` multiple times. Wouldn't it be better to use 
```javascript
const {pathUrlRaw, preservedViews, currentNode} = component.get();
```
I'm not sure but there might be also a performance advantage for this approach.

jesusreal: I like this suggestion!

core/src/services/routing.js
```diff
@@ -236,6 +237,7 @@ export const handleRouteChange = async (path, component, node, config) => {
       pathParams,
       isolateView,
       viewGroup,
+      pathUrlRaw,
```
jesusreal: As the only thing we are doing with `patUrlRaw` is extract the url params in _App.html_, why don't we just save the url params instead?

core/src/services/routing.js
```diff
@@ -148,6 +149,9 @@ export const handleRouteChange = async (
 
     const pathUrlRaw =
       path && path.length ? GenericHelpers.getPathWithoutHash(path) : '';
+    const nodeParamsRaw = pathUrlRaw.split('?')[1]
```
parostatkiem: I think ` const nodeParamsRaw = pathUrlRaw.split('?')[1];` would do exactly the same.

PR #352, opened 2019-01-15
Fix babel polyfill error

parostatkiem:
**Description**

- [x] LuigiCore unit tests
- [x] Console e2e tests

There is a bug (I'd say) in OpenIDclient library that throws an error if `global._babelPolyfill` is already defined which means `require('babel-polyfill')` has been already called somewhere. The normal behavior in this case would be just not requiring it and this is what all other libraries do 😡

- Remove `babel-polyfill` completely from Core in favor of `useBuiltIns` option for `@babel/preset-env`

**Related issue(s)**

### DIFF
--- a/core/.babelrc
+++ b/core/.babelrc
@@ -1,3 +1,10 @@
 {
-  "presets": ["@babel/preset-env"]
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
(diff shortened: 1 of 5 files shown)


parostatkiem:
Got any other idea? `useBuiltIns: 'entry'` doesn't work :(

PR #373, opened 2019-01-22
Publish 0.4.5-rc.3

kwiatekus:
**Description**

- prepare for publishing 0.4.5-rc.3


kwiatekus:
> Would it make sense to upgrade luigi dependencies in sample apps as well?

I usually do it as a first thing after release. So.. once the requested luigi version is actually there on npm js

pekura:
Ok!

PR #385, opened 2019-01-28
Reload microfrontend when clicking on selected node

maxmarkus:
Implemented reload of iframe without triggering spinner.


maxmarkus:
I think it is not possible to trigger a content refresh only, so to keep the asses loaded. This would require eg. a Luigi Client listener, which needs a `state update/route refresh` mechanism implemented in every view for on microfrontend side. Guess this is over-complication. a plain `location.refresh()` does the same, and sources are cached locally anyway

PR #387, opened 2019-01-30
Html svelte formatter

### LINKED ISSUE #383: prettier does not work
Steps to reproduce:
-> newest master
-> npm install
-> change something in TopNav (any if in html code) + save
-> git add .
-> ./scripts/apply-prettier.sh
-> refresh your ide (if necessary)

y-kkamil:
**Description**

- Adds prettyhtml that takes care of formatting svelte components

**Related issue(s)**
Resolves #383


y-kkamil:
Rember to `npm install` in the *root* folder to update hooks.

parostatkiem:
As we discussed, the new formatter formats every .html file in Luigi, including Angular and Vue ones in Examples. Although this is not bad cause it does this quite well, I'd love to see Prettier's `Handlebars` parser on these. It's just better IMO. Please consider it and report your thoughts/results.

y-kkamil:
to keep things consistent, I have added a rule that formats Angular html template files with Prettier alone, there was a recent addition to Prettier (see https://prettier.io/blog/2018/11/07/1.15.0.html#support-html-vue-and-angular-5259-by-ikatyang-4753-by-evilebottnawi-2083-by-azz). For svelte components residing in luigi/core/src/ it still uses prettyhtml. We can discuss it further if we're happy with the solution.

PR #389, opened 2019-01-30
Add generic confirmation modal

jesusreal:
**Description**

- Implement generic confirmation modal, based on promises and reusing existing confirmation modal for unsaved changes
- In angular example app, use 'projects/pr1' for Luigi client ux manager methods and 'projects/pr2' for Luigi client link manager methods
- In angular sample app, add example and e2e test for luigi confirmation modal 
- Add documentation for luigi client
- No need for html sanitation, Svelte is doing it by itself

**Related issue(s)**

### DIFF
--- a/client/src/luigi-client.js
+++ b/client/src/luigi-client.js
@@ -424,7 +424,7 @@ const LuigiClient = {
        * Shows a confirmation modal.
        * @param {Object} content the content of the confirmation modal
        * @param {string} content.header the content of the modal header
-       * @param {string} content.body text for modal body
+       * @param {string} content.body the content of the modal body
        * @param {string} content.buttonConfirm text for modal confirm button
        * @param {string} content.buttonDismiss text for modal dismiss button
        * @returns {promise} A promise which is resolved when accepting confirmation modal and rejected when dismissing it.
--- a/client/src/luigi-client.js
+++ b/client/src/luigi-client.js
@@ -10,9 +10,8 @@ var _onContextUpdatedFns = {};
 var _onInitFns = {};
 var authData = {};
 var pathExistsPromises = {};
-var pathExistsPromises = {};
 let promises = {
-  ux: {}
+  confirmationModal: {}
 };
 /**
  * Creates a random Id
@@ -107,9 +106,9 @@ function luigiClientInit() {
 
     if ('luigi.ux.confirmationModal.hide' === e.data.msg) {
       const data = e.data.data;
-      const promise = promises.ux[data.correlationId];
+      const promise = promises.confirmationModal;
       data.confirmed ? promise.resolveFn() : promise.rejectFn();
-      delete promises.ux[data.correlationId];
+      delete promises.confirmationModal;
     }
   });
 
@@ -427,23 +426,19 @@ const LuigiClient = {
        * @returns {promise} A promise which is resolved when accepting confirmation modal and rejected when dismissing it.
        */
       showConfirmationModal: function showConfirmationModal(content) {
-        const id = Date.now();
         window.parent.postMessage(
           {
             msg: 'luigi.ux.confirmation-modal-show',
-            data: {
-              content,
-              correlationId: id
-            }
+            data: { content }
           },
           '*'
         );
-        promises.ux[id] = {};
-        promises.ux[id].promise = new Promise((resolve, reject) => {
-          promises.ux[id].resolveFn = resolve;
-          promises.ux[id].rejectFn = reject;
+        promises.confirmationModal = {};
+        promises.confirmationModal.promise = new Promise((resolve, reject) => {
+          promises.confirmationModal.resolveFn = resolve;
+          promises.confirmationModal.rejectFn = reject;
         });
-        return promises.ux[id].promise;
+        return promises.confirmationModal.promise;
       }
     };
   }
--- a/client/src/luigi-client.js
+++ b/client/src/luigi-client.js
@@ -427,7 +427,7 @@ const LuigiClient = {
        * @param {string} content.body the content of the modal body
        * @param {string} content.buttonConfirm the label for the modal confirm button
        * @param {string} content.buttonDismiss the label for the modal dismiss button
-       * @returns {promise} A promise which is resolved when accepting confirmation modal and rejected when dismissing it.
+       * @returns {promise} A promise which is resolved when accepting the confirmation modal and rejected when dismissing it.
        */
       showConfirmationModal: function showConfirmationModal(content) {
         window.parent.postMessage(
--- a/client/src/luigi-client.js
+++ b/client/src/luigi-client.js
@@ -423,7 +423,7 @@ const LuigiClient = {
       /**
        * Shows a confirmation modal.
        * @param {Object} content the content of the confirmation modal
-       * @param {string} content.header text for modal header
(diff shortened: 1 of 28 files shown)


client/src/luigi-client.js
```diff
@@ -10,7 +10,9 @@ var _onContextUpdatedFns = {};
 var _onInitFns = {};
 var authData = {};
 var pathExistsPromises = {};
-
+let promises = {
```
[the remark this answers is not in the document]
jesusreal: I created it having in mind a future refactoring to include all promises in same object. Apart from this one, we have pathExists promise, and we will soon have generic alert promise. This solution scales better in my opinion

core/src/ConfirmationModal.html
```diff
@@ -21,15 +21,19 @@ <h1 class="fd-modal__title">{title}</h1>
 <script type="text/javascript">
   export default {
     data() {
-      //fallback data in case title or text props weren't provided
+      // fallback data in case not provided
```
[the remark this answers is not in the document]
jesusreal: So we have a confirmation modal component. We are using it currently once in App.html, in this case, we are alway providing the data to the component. Whenever you provide the data, it will never use the fallback data, even if the data you provide is undefined. I also tried with single variables instead of an object. If you give variables as input when using the component, it will never take the default values. 
Default values would be taken if we use somewhere else the component and we do not give the 'content' as input. We could make it more complicated to match what you (and mostly anyone) would expect. Let´s discuss the best solution offline.

core/examples/luigi-sample-angular/src/app/project/project.component.html
```diff
@@ -1,104 +1,128 @@
...
+          </p>
+        </li>
+        <li class="fd-list-group__item">
+          <button class="fd-button" data-cy="show-luigi-confirmation-modal" (click)="showConfirmationModal()">Show Luigi confirmation modal</button>
+          <p>
+            <app-code-snippet data="luigiClient.uxManager().showConfirmationModal(content)"></app-code-snippet>
+          </p>
+          <p class="fd-has-font-style-italic" data-cy="luigi-confirmation-modal-result"
+            ngClass="{{(confirmationModalResult.currentKey === 'confirmed') ? 'fd-has-color-status-1' : 'fd-has-color-status-2'}}">
+            {{confirmationModalResult.info[confirmationModalResult.currentKey]}}
```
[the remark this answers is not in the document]
jesusreal: I totally agree with you, this is overcomplicated. Thanks for pointing that out. Refactored

docs/luigi-client-api.md
```diff
@@ -212,3 +212,13 @@ This method informs the main application that there are unsaved changes in the c
...
 
 -   `isDirty` **[boolean](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Boolean)** indicates if there are any unsaved changes on the current page or in the component
+
+### showConfirmationModal
+
+Shows a confirmation modal.
+
+#### Parameters
+
+-   `content` **[Object](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object)** Text for different elements in confirmation modal
```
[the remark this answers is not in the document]
jesusreal: Absolutely, I have added them

client/src/luigi-client.js
```diff
@@ -410,6 +419,30 @@ const LuigiClient = {
           { msg: 'luigi.set-page-dirty', dirty: isDirty },
           '*'
         );
+      },
+      /**
+       * Shows a confirmation modal.
+       * @param {Object} content the content of the confirmation modal
+       * @param {string} content.header text for modal header
```
bszwarc:
```suggestion
       * @param {string} content.header the content of the modal header
```

client/src/luigi-client.js
```diff
@@ -410,6 +419,30 @@ const LuigiClient = {
...
+       * @param {string} content.body text for modal body
```
bszwarc:
```suggestion
       * @param {string} content.body the content of the modal body
```

client/src/luigi-client.js
```diff
@@ -410,6 +419,30 @@ const LuigiClient = {
...
+       * @param {string} content.buttonConfirm text for modal confirm button
```
bszwarc:
```suggestion
       * @param {string} content.buttonConfirm the label for the modal confirm button
```

client/src/luigi-client.js
```diff
@@ -410,6 +419,30 @@ const LuigiClient = {
...
+       * @param {string} content.buttonDismiss text for modal dismiss button
```
bszwarc:
```suggestion
       * @param {string} content.buttonDismiss the label for the modal dismiss button
```

bszwarc: I thought that maybe in the case of a button `label` will be better? But it's not a must.

client/src/luigi-client.js
```diff
@@ -410,6 +419,30 @@ const LuigiClient = {
...
         );
+      },
+      /**
+       * Shows a confirmation modal.
+       * @param {Object} content the content of the confirmation modal
+       * @param {string} content.header the content of the modal header
+       * @param {string} content.body the content of the modal body
+       * @param {string} content.buttonConfirm the label for the modal confirm button
+       * @param {string} content.buttonDismiss the label for the modal dismiss button
+       * @returns {promise} A promise which is resolved when accepting confirmation modal and rejected when dismissing it.
```
bszwarc:
```suggestion
       * @returns  a {promise} which is resolved when accepting the confirmation modal and rejected when dismissing it.
```

jesusreal: This is not correct regarding documentation. First comes the type of the returned value and the the additional info. I will just add the "the"

bszwarc: Ok :) I was just looking at the entire document, and under `has Back` I've found:
`Returns boolean indicating if there is a preserved view you can return to.` that's why I wanted to adjust the other part of the doc. Should we change this one as well, or is it a different case?

Also let's then write `a promise`  - lower case, to make it consistent :)

jesusreal: The piece of documentation you mentioned is:
```
       * @returns {boolean} indicating if there is a preserved view you can return to.
```
I will replace the line we are talking about with:
```
       * @returns {promise} which is resolved when accepting the confirmation modal and rejected when dismissing it.
```
This makes it consistent imho. Is that ok for you?

core/src/ConfirmationModal.html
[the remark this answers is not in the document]
jesusreal: Added mechanism to set default values if not provided. If values are provided, even with a falsy value like empty string or undefined, they will be taken

core/src/ConfirmationModal.html
```diff
@@ -45,12 +45,7 @@ <h1 class="fd-modal__title">{content.header}</h1>
   $width: 32rem;
   .fd-modal {
     position: absolute;
-    box-sizing: border-box;
-    width: $width;
-    max-width: 90vw;
-    top: calc(#{$topNavHeight} + 1rem);
     left: calc(50% + #{$leftNavWidth}/ 2 - #{$width}/ 2);
```
dariadomagala: This should be removed as well if we want to have it in the middle of the screen.

PR #390, opened 2019-01-30
Implement test coverage feature

parostatkiem:
**Description**
### ⚠️"Why do I see an _alpha_ package?!"
`Istanbul` hasn't been officially updated for 2 years since _0.4.5_ version came out. That version (the newest stable one) uses `babel-core 6.x.x` and LuigiCore uses `@babel/core ^7.0.0`. 
Somehow it just can't run together in harmony, furthermore having 2 Babels would be confusing (at least). I've searched the whole Internet and everyone says the only solution is including an _alpha_ version of `Istanbul` which works really well (they say) and uses `@babel/core`.

#### "What about `test:watch`?"
It works as it did before. The report is generated after you end the watch process (_ctrl+c_).
No, there's no other solution. Istanbul collects data from the output of Mocha's test process. An output is generated at process' exit.

Although final changes are really small, this wasn't an easy task. I've tried everything I could to avoid _alpha_ version of `Istanbul` + there were many more complications.

- Use `Istanbul` lib to generate test coverage report
- ...
- ...

**Related issue(s)**

### DIFF
--- a/core/package-lock.json
+++ b/core/package-lock.json
@@ -33,44 +33,6 @@
         "resolve": "^1.3.2",
         "semver": "^5.4.1",
         "source-map": "^0.5.0"
-      },
-      "dependencies": {
-        "debug": {
-          "version": "4.1.1",
-          "resolved": "https://registry.npmjs.org/debug/-/debug-4.1.1.tgz",
-          "integrity": "sha512-pYAIzeRo8J6KPEaJ0VWOh5Pzkbw/RetuzehGM7QRRX5he4fPHx2rdKMB256ehJCkX+XRQm16eZLqLNS8RSZXZw==",
-          "dev": true,
-          "requires": {
-            "ms": "^2.1.1"
-          }
-        },
-        "json5": {
-          "version": "2.1.0",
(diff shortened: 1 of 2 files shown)


dariadomagala:
Is there anything else other than `Istanbul`? Maybe there is a different package that is stable and properly tested?

parostatkiem:
Didn't work, the other `Istanbul` extension (forgot the name) also didn't work.

y-kkamil:
I have managed to run [nyc](https://github.com/istanbuljs/nyc) as a coverage reporter. It gives different output than your solution, though. I think we should investigate this further - where this difference comes from, which one is correct, and is there any added value for using nyc over old istanbul.

parostatkiem:
@y-kkamil Looks like NYC is just a command line overlay for Istanbul so they seem to have common roots. I've got no idea what causes the coverage differences then 😆 

IMO we can stick to my solution. As I said on daily, in the worst case that the _alpha_ package fails, no customer/Luigi user will be harmed. We just won't be able to see test coverage, not a big deal...
Furthermore many people reported this package to be stable.

PR #394, opened 2019-02-01
fix application setup issues

JohannesDoberer:
fix application setup issues


parostatkiem:
... aaaand this is the result of Angular setup script:
image
Again, works well when done with separate commands.

docs/application-setup.md
```diff
@@ -66,7 +66,7 @@ curl https://raw.githubusercontent.com/kyma-project/Luigi/master/core/examples/l
```
jesusreal: Not on this line, but in line 12 'Intitial' should be replaced with 'Initial'

JohannesDoberer: I think this commands are important, because a customer who wants to play with Luigi and do a hello world example should see what's going on. If you just execute a script you will not see what happen and I doubt that developers look in the script before or afterwards (at least the most of them). If you look at hello world examples like angular, webpack..etc you always have to copy and paste lines of code (unless you have the possibility to use a cli tool). Actually the scripts are just for us (or for some guys who have to do this hello world example more than one time) to save time and don't have to copy and paste every time to test if we have breaking changes.
.... and the PO wants to have this commands as well :)

jesusreal: Ok then... I am just not a fun of having to maintain the same in 2 places. We had several times the issue that the inline code was not up to date, which means it was useless, and IMHO this is mostly solved if you replace the inline code with the scripts, since at the time of releasing Luigi is way more faster to check the validity of the setups. The code is in the scripts anyway... This is my opinion, but I accept what the majority in the team wants to have.

docs/application-setup.md
```diff
@@ -135,20 +135,25 @@ npm run start
 npm install -g live-server
 ````
 
-2. Use the following commands to create a directory for your application, install Luigi, make assets available, and start your local server.
+2. Use the following installer to create a directory for your application, install Luigi, make assets available, and start your local server
```
bszwarc:
```suggestion
2. Use the installer to create a directory for your application, install Luigi, make assets available, and start your local server:
```

docs/application-setup.md
```diff
@@ -9,7 +9,7 @@ Choose the framework to build your application:
 [SAPUI5/OpenUI5](#sapui5)<br>
 [VUE.JS](#vuejs)
 
-## Intitial steps
+## Initial steps
```
bszwarc: I was wondering if we maybe can have something better than `initial steps` Maybe `Basic application setup` 

what do you think?

JohannesDoberer: You are right. Done.

docs/application-setup.md
JohannesDoberer: done

PR #395, opened 2019-02-06
Add generic alert

jesusreal:
**Description**

- Implement generic alert, based on promises and reusing existing alert component for not found paths. Features:
  - Types: 'info'|'success'|'warning'|'error'
  - Optional links to app internal views as part of the alert message
- In angular example app, add example and e2e tests for generic luigi alert
- Add documentation for Luigi client
- Alerts content is sanitised since due to the links we are rendering the content as html, not as a string

**Related issue(s)**


core/src/Alert.html
```diff
@@ -0,0 +1,99 @@
+<div
+  class="fd-alert fd-alert--{settings.type} {settings.dismissButton ? 'fd-alert--dismissible' : ''}"
+  role="alert"
```
JohannesDoberer: fd-alert--info does not work. It should be fd-alert--information as far as I can see.

core/src/Alert.html
```diff
@@ -0,0 +1,99 @@
...
+<script type="text/javascript">
+  import * as Routing from './services/routing.js';
+  import * as RoutingHelpers from './utilities/helpers/routing-helpers';
+
+  export default {
+    oncreate() {
+      const defaultSettings = {
+        text:
+          'I am a Luigi alert with extremely relevant information for the user',
+        type: 'info',
```
JohannesDoberer: bildschirmfoto 2019-02-06 um 11 45 27

core/src/App.html
```diff
@@ -258,18 +256,6 @@
           builderCompatibilityMode: Boolean(window.builderCompatibilityMode),
           isolateAllViews
         };
-        window.addEventListener('popstate', async e => {
```
dariadomagala: The 404 error alert was supposed to be visible until the route change. Currently, it's visible until you close it manually.

client/src/luigi-client.js
```diff
@@ -443,6 +448,55 @@ const LuigiClient = {
...
+       * }
+       * LuigiClient
+       *  .uxManager()
+       *  .showAlert(settings)
+       *  .then(() => {
+       *     // Logic to execute when the alert is dismissed
+       * });
+
+       */
+      showAlert: function showAlert(settings) {
```
dariadomagala: I think we should also add a possibility to hide the alert (for example if someone wants to hide it after the navigation).

core/src/Alert.html
jesusreal: done

core/src/App.html
jesusreal: This is on purpose. SAP Fiori Fundamental is not providing automatic dismiss in their guidelines, so the user will always have to close it by himself

core/src/Alert.html
jesusreal: Default text not being displayed when not text field is provided is now fixed

client/src/luigi-client.js
jesusreal: We will discuss it internally and provide the best solution to let the programmer of the microfrontend close the alert without the user explicitly clicking the dismiss button of the alert.

core/src/Alert.html
```diff
@@ -0,0 +1,110 @@
...
+
+      const { text, links } = this.get().settings;
+      const processedData = this.processTextAndLinks(text, links);
+      this.set({
+        settings: Object.assign(defaultSettings, this.get().settings, {
+          text: processedData.sanitizedText
+        })
+      });
+      processedData.links.forEach(l => {
+        this.addCLickListener(l.url, l.elemId);
```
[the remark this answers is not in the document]
jesusreal: Already pushed that one :)

core/src/Alert.html
```diff
@@ -0,0 +1,113 @@
...
+          success: 'success',
+          warning: 'warning',
+          error: 'error'
+        }
+      };
+    },
+    oncreate() {
+      const defaultSettings = {
+        text:
+          'I am a Luigi alert with extremely relevant information for the user',
```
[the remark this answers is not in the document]
jesusreal: Great point. I will display an info message in the browser console if an alert has no text and will not display anything in the UI. Additionally, I will remove the default values for alerts, since this makes no sense:
```
      const defaultSettings = {
        type: 'info',
        links: undefined
      };
```

core/examples/luigi-sample-angular/src/app/project/project.component.html
```diff
@@ -76,6 +76,19 @@ <h1 class="fd-section__title">LuigiClient uxManager methods</h1>
...
+            <label class="fd-form__label">
+              <input
+                [ngModel]="true"
+                [disabled]="this.alertDismissed === false"
+                type="checkbox"
+                name="text"
+                class="fd-form__control"
+                data-cy="luigi-alert-text"
+              />
+              With text (if alert has not text it is not displayed)
```
dariadomagala: Tiny thing but: has **no** text 😉

jesusreal: both updates are done

client/src/luigi-client.js
```diff
@@ -443,6 +448,52 @@ const LuigiClient = {
           promises.confirmationModal.rejectFn = reject;
         });
         return promises.confirmationModal.promise;
+      },
+
+      /**
+       * Shows an alert.
+       * @param {Object} settings the settings for the alert
+       * @param {string} settings.text the content of the alert. To add a link to the content, you have to setup the link in `links` object. The key(s) in the `links` object must be used in the text to reference the links, wrapped in curly braces with no spaces. If no text is specified, the alert will not be displayed
```
bszwarc:
```suggestion
       * @param {string} settings.text the content of the alert. To add a link to the content, you have to set up the link in the `links` object. The key(s) in the `links` object must be used in the text to reference the links, wrapped in curly brackets with no spaces. If you don't specify any text, the alert is not displayed.
```

I was wondering if we should keep the dot in a parameter ref if we have more sentences to it - usually we don't but maybe it makes sense here.

client/src/luigi-client.js
```diff
@@ -443,6 +448,52 @@ const LuigiClient = {
...
         return promises.confirmationModal.promise;
+      },
+
+      /**
+       * Shows an alert.
+       * @param {Object} settings the settings for the alert
+       * @param {string} settings.text the content of the alert. To add a link to the content, you have to setup the link in `links` object. The key(s) in the `links` object must be used in the text to reference the links, wrapped in curly braces with no spaces. If no text is specified, the alert will not be displayed 
+       * @param {('info'|'success'|'warning'|'error')} settings.type sets the type of the alert
+       * @param {Object} settings.links provides links data
+       * @param {Object} settings.links.LINK_KEY object containing the data for a concrete link.
```
bszwarc:
```suggestion
       * @param {Object} settings.links.LINK_KEY object containing the data for a particular link.
     
```

client/src/luigi-client.js
```diff
@@ -443,6 +448,52 @@ const LuigiClient = {
...
+      },
+
+      /**
+       * Shows an alert.
+       * @param {Object} settings the settings for the alert
+       * @param {string} settings.text the content of the alert. To add a link to the content, you have to setup the link in `links` object. The key(s) in the `links` object must be used in the text to reference the links, wrapped in curly braces with no spaces. If no text is specified, the alert will not be displayed 
+       * @param {('info'|'success'|'warning'|'error')} settings.type sets the type of the alert
+       * @param {Object} settings.links provides links data
+       * @param {Object} settings.links.LINK_KEY object containing the data for a concrete link.  
+       * Read above on how to use the key to for the link to be properly rendered in the alert message.
```
bszwarc:
```suggestion
       * To properly render the link in the alert message refer to the description of the **settings.links.LINK.KEY** parameter.
```

client/src/luigi-client.js
```diff
@@ -443,6 +448,52 @@ const LuigiClient = {
...
+       * @param {string} settings.links.LINK_KEY.text text which will replace the link identifier in the alert content
```
bszwarc:
```suggestion
       * @param {string} settings.links.LINK_KEY.text text which replaces the link identifier in the alert content
```

client/src/luigi-client.js
```diff
@@ -443,6 +448,52 @@ const LuigiClient = {
...
+       * @param {string} settings.links.LINK_KEY.url url to navigate when the link is clicked. Currently only internal links are supported in the form of relative or absolute paths.
```
bszwarc:
```suggestion
       * @param {string} settings.links.LINK_KEY.url url to navigate when you click the link. Currently, only internal links are supported in the form of relative or absolute paths.
```

client/src/luigi-client.js
```diff
@@ -443,6 +448,52 @@ const LuigiClient = {
...
+       * @returns {promise} which is resolved when the alert is dismissed.
```
bszwarc:
```suggestion
       * @returns {promise} which is resolved when the alert is dismissed
```

We can remove the dot here.

client/src/luigi-client.js
jesusreal: Almost, it is `settings.text` instead of `settings.links.LINK.KEY`. I will commit it separately.

Issue #405, opened 2019-02-20
Focus improvements


aartek:
Hi, it'd be great if the menu items would be focusable and selectable from a keyboard (tab, shift+tab, enter). It should be also clear which menu item is currently focused (visible `:focus` state).  Also I think that after selecting a menu item, focus should be set on an opened iframe, to let the user navigate inside the selected ui

PR #412, opened 2019-03-04
#405 Focus improvements

hardl:
#405 Focus improvements

### DIFF
--- a/core/src/navigation/LeftNav.html
+++ b/core/src/navigation/LeftNav.html
@@ -159,12 +159,11 @@
       },
 
       handleKey(event) {
-        let code = event.code;
+        const code = event.code;
         if(code === 'ArrowRight') {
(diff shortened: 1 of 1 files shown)


core/src/navigation/LeftNav.html
```diff
@@ -158,6 +158,17 @@
         this.fire('handleClick', { node });
       },
 
+      handleKey(event) {
+        let code = event.code;
```
jesusreal: Works as expected. Just a couple of cosmetic comments:
- All variables can be defined with const
- Maybe the `win` variable initialization would be simpler with `iframe.contentWindow || (iframe.contentDocument && iframe.contentDocument.window))`
- Actually do we really need `|| (iframe.contentDocument ? iframe.contentDocument.window : null)`? contentDocument is supposed to be like document and document has no focus method.

PR #435, opened 2019-03-25
Fix alert message hidden under content window

y-kkamil:
**Description**

- Moves alert message box on top of content window.


core/src/Alert.html
```diff
@@ -74,4 +83,10 @@
       }
     }
   };
-</script>
\ No newline at end of file
+</script>
+
+<style type="text/scss">
+  .fd-overlay--alert {
+    z-index: 1100;
```
maxmarkus: more z-index, please :-D :-D

PR #438, opened 2019-03-26
Improve login error behavior

maxmarkus:
Redirecting now to logout.html after token issues.
Could be tested in combination with https://github.com/kyma-project/console/pull/720


core/src/providers/auth/openIdConnect.js
```diff
@@ -162,12 +162,12 @@ export class openIdConnect {
             } else {
               window.location.href = window.location.origin;
             }
-            // resolve(true); // not resolving in order to not interrupt window.location.href
           }, 50);
         })
         .catch(err => {
           console.error(err);
-          reject(err);
+          localStorage.clear();
```
pekura: - Is it not too harsch to clear the whole local storage? It would be more polite to just clear keys this auth provider handles.
- Do we clear auth data in non-error logout flows as well?
- Is there a reason to not use session storage for the auth data?

pekura:
The logout page is not a beauty:

Bildschirmfoto 2019-03-27 um 10 19 17

Can we improve that a bit?

core/src/providers/auth/openIdConnect.js
maxmarkus: Yes, you are right. We usually use `localStorage.removeItem('luigi.auth');` upon errors and normal logout. I guess localStorage is used to be able to close the window and still be logged in without going trough sso flow.

pekura: Please doublecheck with Philipp if he is fine with using localStorae here, thanks!

maxmarkus:
We have relied on localStorage so far, changing it to sessionStorage would be breaking change. Maybe we should add a config setting so the user can switch which storage method he wants to use. Should be done in separate ticket.

PR #446, opened 2019-03-28
Modal mfs

JohannesDoberer:


core/src/Modal.html
```diff
@@ -0,0 +1,134 @@
...
+        modalIframe: iframe,
+        modalIframeData: { ...pathData, nodeParams }
+      });
+    } else {
+      await prepareNodeData(component, path);
+    }
+  };
+  const createIframeModal = (viewUrl, componentData, component) => {
+    const elem = document.getElementsByClassName('fd-modal');
+    let sizeInPercent = '80%';
```
parostatkiem: The default value is cool 👌 but please mention it in Luigi API (docu)

core/src/Modal.html
```diff
@@ -0,0 +1,134 @@
...
+    if (component.get().modalSettings.size) {
+      if (component.get().modalSettings.size === 'l') {
+        sizeInPercent = '80%';
+      } else if (component.get().modalSettings.size === 'm') {
+        sizeInPercent = '60%';
+      } else if (component.get().modalSettings.size === 's') {
+        sizeInPercent = '40%';
+      } else {
+        component.get().modalSettings['size'] = size;
```
parostatkiem: Assignment to `get`?

core/src/Modal.html
```diff
@@ -0,0 +1,134 @@
...
+      }
+    }
+    if (!component.get().modalSettings.title) {
+      document.getElementsByClassName('fd-modal__title')[0].style.visibility =
```
parostatkiem: Please render it conditionally in template, not via CSS.

core/src/Modal.html
```diff
@@ -0,0 +1,134 @@
...
+    elem[0].setAttribute(
+      'style',
+      `width:${sizeInPercent};height:${sizeInPercent}`
+    );
+    if (viewUrl) {
+      viewUrl = RoutingHelpers.substituteViewUrl(viewUrl, componentData);
+    }
+    const iframe = document.createElement('iframe');
+    iframe.src = viewUrl;
+    iframe.style.width = '100%';
```
parostatkiem: I't really like to see a class here instead of separate style rules.

core/src/Modal.html
parostatkiem: Also, I'd change the variable name or its values. You double the "percent" thing.

core/src/App.html
```diff
@@ -273,16 +295,33 @@
...
+            ) {
+              let ctx = this.get().modalIframeData.context;
+              ctx.modal = true;
+              const modalConfig = {
+                ...config,
+                iframe: this.get().modalIframe,
+                context: ctx,
+                pathParams: this.get().modalIframeData.pathParams,
+                nodeParams: this.get().modalIframeData.nodeParams,
+                model: true
```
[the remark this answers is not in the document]
JohannesDoberer: It is necessary if you want to use pathParams or nodeParams in a modal.
Try something like this:
`linkManager().openModal('/environments/env1?~test=tets')` or
`linkManager().openModal('/environments/env1')`

client/src/luigi-client.js
```diff
@@ -267,12 +273,24 @@ const LuigiClient = {
...
+       * Opens a microfrontend in a modal
+       * @param {string} path path to be navigated to
+       * @param {Object} modalSettings settings to customize the modal title and size (s=small, m=medium, l=large)
+       */
+      openModal: function(path, modalSettings) {
+        let settings = modalSettings;
+        if (!settings) {
+          settings = {};
+        }
+        this.navigate(path, 0, true, settings);
```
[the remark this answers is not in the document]
JohannesDoberer: 👍

core/src/Modal.html
JohannesDoberer: I also thought about it, but this is a 'special' case. If no title is there e.g. for groups than we have two possibilities. Hide 'undefined' as modal__title or override css for fd-modal__title, because the height is broken.
In this case the conditional case is more pragmatic, for me.
Bildschirmfoto 2019-04-01 um 15 00 58

core/src/Modal.html
```diff
@@ -0,0 +1,134 @@
...
+      path
+    );
+    const params = RoutingHelpers.parseParams(pathUrlRaw.split('?')[1]);
+    const nodeParams = RoutingHelpers.getNodeParams(params);
+    const lastNode = RoutingHelpers.getLastNodeObject(pathData);
+    const modalSettings = component.get().modalSettings;
+    if (!modalSettings.title) {
+      modalSettings.title = lastNode.label;
+    }
+    component.get().isDataPrepared = true;
```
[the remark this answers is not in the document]
JohannesDoberer: good point, I reduced it. Line is not needed anymore.

client/src/luigi-client.js
```diff
@@ -374,7 +392,11 @@ const LuigiClient = {
        * @returns {boolean} indicating if there is a preserved view you can return to.
        */
       hasBack: function hasBack() {
-        return Boolean(currentContext.internal.viewStackSize !== 0);
+        if (currentContext.context.modal) {
```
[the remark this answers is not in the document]
JohannesDoberer: Matter of taste... but fine for me :)

core/src/Modal.html
JohannesDoberer: 👍

core/src/Modal.html
JohannesDoberer: 👍

core/src/App.html
```diff
@@ -218,6 +230,15 @@
     };
   };
 
+  const getDefaultMicrofrontendModalData = () => {
```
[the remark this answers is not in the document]
JohannesDoberer: We don't want a const object here. If this method is callled it should return a 'fresh' object with the possibility to extend it and not mutate the existing one.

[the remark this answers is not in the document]
parostatkiem: You can always use `{...the_object}` in case there are no reference types inside (which is this case) but I'm fine with current approach 👍

core/src/Modal.html
parostatkiem: You mean the title is `"undefined"` instead of `undefined`? If yes, I'd still vote for having `modal.title && modal.title!== 'undefined'` as a condition. If no, please explain it in an other way 😃

PR #448, opened 2019-03-28
Escape some html characters for alert component

### LINKED ISSUE #447: Unescaped character may appear in the Alert component
**Description**

Alert component in luigi core may execute third-party code, as there is no html character escaping present for it and it executes data coming from microfronteds.

**Proposed mitigation**

Add the escaping for problematic html characters.

y-kkamil:
**Description**

- Add html characters encoding for Alert component
- Move sanitisation functions to the new helper file for better readability

**Related issue(s)**
Fixes #447

### DIFF
--- a/core/src/Alert.html
+++ b/core/src/Alert.html
@@ -17,7 +17,11 @@
 </div>
 
 <script type="text/javascript">
-  import { sanitizeHtml } from './utilities/helpers/escaping-helpers';
+  import {
+    sanitizeHtml,
+    sanitizeParam,
+    escapeRegExp
+  } from './utilities/helpers/escaping-helpers';
 
   export default {
     data() {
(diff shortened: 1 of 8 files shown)


y-kkamil:
Oooops.. I found another issue - will fix it in a minute

core/src/Alert.html
```diff
@@ -73,7 +72,8 @@
 
         return Object.entries(links).reduce((acc, [key, content]) => {
           const elemId = key;
-          const processedData = `<a id="${elemId}">${content.text}</a>`;
+          const escapedText = sanitizeHtml(content.text);
+          const processedData = `<a id="${elemId}">${escapedText}</a>`;
```
pekura: I would escape/sanitize the key/elemId as well - it won't unlikely be misused (because it will be escaped in the text before so it is not found when we want to replace it with <a> tag - but it won't work then.

core/src/Alert.html
```diff
@@ -73,7 +72,8 @@
...
           const pattern = new RegExp(`({${key}})`, 'g');
```
pekura: The key can contain characters that regexp interpretes (like '.' for example). The key needs to be passed here regexp-escaped! (See here https://stackoverflow.com/questions/3561493/is-there-a-regexp-escape-function-in-javascript/3561711).

PR #451, opened 2019-04-01
Prevent unlogged rendering

parostatkiem:
**Description**
✔️ Tested with all combinations of `auth.use`, `auth.disableAutoLogin` and login state.

- Add `isHidden` attribute to _Authorization_ component which makes it not render anything, just fire `onCreate` hook (and proceed autologin)
- Render the whole _TopNav_ only if needed. Otherwise, "display" hidden _Authorization_.

**Related issue(s)**

### DIFF
--- a/core/src/navigation/TopNav.html
+++ b/core/src/navigation/TopNav.html
@@ -1,5 +1,5 @@
 <svelte:window on:click="closeAllDropdowns()" on:blur="closeAllDropdowns()" />
-{#if authorizationEnabled && (!autologinEnabled || isLoggedIn) || !authorizationEnabled }
+{#if showTopNav}
   <div class="fd-shellbar {hideNavComponent ? 'hideNavComponent' : ''}">
     <div class="fd-shellbar__group fd-shellbar__group--start">
       <LogoTitle /> 
@@ -223,9 +223,9 @@
     oncreate() {
       this.set({
         authorizationEnabled: LuigiConfig.isAuthorizationEnabled(),
-        autologinEnabled: !(LuigiConfig.getConfigValue(
+        autologinEnabled: !Boolean(LuigiConfig.getConfigValue(
           'auth.disableAutoLogin'
-        ) || false),
+        )),
         isProductSwitcherAvailable: LuigiConfig.getConfigValue(
           'navigation.productSwitcher'
         ),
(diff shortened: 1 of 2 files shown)


core/src/navigation/TopNav.html
```diff
@@ -1,176 +1,181 @@
...
-            class="fd-shellbar__action fd-shellbar__action--hide fd-shellbar__action--collapsible"
-          >
-            <button
-              title="{node.label}"
-              class="fd-button--shell {node === selectedNode ? 'is-selected' : ''} {node.externalLink && node.externalLink.url ? 'fd-global-nav__btn--external' : ''}"
-              aria-controls="0AcWE812"
-              aria-expanded="false"
-              aria-haspopup="true"
-              on:click="handleClick(node)"
+{#if authorizationEnabled && (!autologinEnabled || isLoggedIn) || !authorizationEnabled }
```
jesusreal: I would add the extra parenthesis to avoid anyone reading the code the extra thinking:
```{#if (authorizationEnabled && (!autologinEnabled || isLoggedIn)) || !authorizationEnabled }```
Additionally, I would set a variable for that and use it here and below, to make explicit the relation between this value and the input used for authorization component:
```
{#if showTopNav }
...
...
<Authorization isHidden={!showTopNav}/> // or <Authorization showMarkup={showTopNav}/> which is simpler
```

File: core/src/navigation/TopNav.html [hunk 2]
```diff
@@ -218,6 +223,9 @@
     oncreate() {
       this.set({
         authorizationEnabled: LuigiConfig.isAuthorizationEnabled(),
+        autologinEnabled: !(LuigiConfig.getConfigValue(
```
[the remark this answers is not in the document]
parostatkiem: Ideally would be to use TypeScript :)

core/src/navigation/TopNav.html
parostatkiem: The second thing is impossible because in most cases Authorization is displayed inside TopNav which should be hidden in some of them. So it's logically impossible :)

jesusreal: I meant in the top level if statement, where you used: `<Authorization isHidden={true}/>`. Correct me if I am wrong, but I think it should be possible to use the variable there.

parostatkiem: Well, that would change nothing at all except the fact of having `{!showTopNav}` instead of `{true}`. If this particular <Authorization> is used, it means that `showTopNav` is false. I can change it if you want but for me `{false}` is more clear here - it indicates the fact I mentioned (it is always false here).

jesusreal: I see your point. I am fine keeping it as it is.

PR #461, opened 2019-04-03
Prevent unescaped characters in the Alert component

### LINKED ISSUE #447: Unescaped character may appear in the Alert component

y-kkamil:
**Description**

- Change lifecycle hook to prevent unescaped characters in the Alert component.

**Related issue(s)**
Fixes #447

### DIFF
--- a/core/src/Alert.html
+++ b/core/src/Alert.html
@@ -12,7 +12,7 @@
       aria-controls="j2ALl423"
       data-cy="luigi-alert-dismiss"
     ></button>
-    {@html settings.alertText}
+    {@html dataSanitized ? settings.text: ''}
   </div>
 </div>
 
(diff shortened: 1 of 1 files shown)


y-kkamil:
apparently, the `onestate` hook fires once before `oncreate` - see more: https://svelte.technology/guide#lifecycle-hooks

core/src/Alert.html
```diff
@@ -39,7 +39,7 @@
       const processedData = this.processTextAndLinks(text, links);
       this.set({
         settings: Object.assign(this.get().settings, {
-          text: processedData.sanitizedText
+          alertText: processedData.sanitizedText
```
pekura: What happens if someone initializes the Alert with "settings.alertText" filled with (malicious) text?

y-kkamil: fixed

PR #468, opened 2019-04-04
Check origin of postmessage

dariadomagala:
**Description**

- Check origin of postmessage

**Related issue(s)**

### DIFF
--- a/core/src/App.html
+++ b/core/src/App.html
@@ -295,35 +295,36 @@
           }
         });
         window.addEventListener('message', async e => {
-          if ('luigi.get-context' === e.data.msg) {
-            if (
-              this.get().modalIframe &&
-              e.source === this.get().modalIframe.contentWindow
-            ) {
-              let ctx = this.get().modalIframeData.context;
-              const modalConfig = {
-                ...config,
-                iframe: this.get().modalIframe,
-                context: ctx,
-                pathParams: this.get().modalIframeData.pathParams,
-                nodeParams: this.get().modalIframeData.nodeParams,
-                modal: true
-              };
-              sendContextToClient(this, modalConfig, {});
-            } else if (config.iframe && e.source === config.iframe.contentWindow) {
-              sendContextToClient(this, config, {});
+          if (e.source === config.iframe.contentWindow || (this.get().modalIframe && e.source === this.get().modalIframe.contentWindow)) {
 
-              const loadingIndicatorAutoHideEnabled =
-                GenericHelpers.getConfigValueFromObject(
-                  this.get(),
-                  'currentNode.loadingIndicator.hideAutomatically'
-                ) !== false;
-              if (loadingIndicatorAutoHideEnabled) {
-                this.set({ showLoadingIndicator: false });
+            if ('luigi.get-context' === e.data.msg) {
+              if (
+                this.get().modalIframe &&
+                e.source === this.get().modalIframe.contentWindow
+              ) {
+                let ctx = this.get().modalIframeData.context;
+                const modalConfig = {
+                  ...config,
+                  iframe: this.get().modalIframe,
+                  context: ctx,
+                  pathParams: this.get().modalIframeData.pathParams,
+                  nodeParams: this.get().modalIframeData.nodeParams,
+                  modal: true
+                };
+                sendContextToClient(this, modalConfig, {});
+              } else if (config.iframe && e.source === config.iframe.contentWindow) {
+                sendContextToClient(this, config, {});
+
+                const loadingIndicatorAutoHideEnabled =
+                  GenericHelpers.getConfigValueFromObject(
+                    this.get(),
+                    'currentNode.loadingIndicator.hideAutomatically'
+                  ) !== false;
+                if (loadingIndicatorAutoHideEnabled) {
+                  this.set({ showLoadingIndicator: false });
+                }
               }
             }
-          }
-          if (e.source === config.iframe.contentWindow) {
 
             if ('luigi.show-loading-indicator' === e.data.msg) {
               this.set({ showLoadingIndicator: true });
--- a/core/src/App.html
+++ b/core/src/App.html
@@ -239,6 +239,18 @@
     };
   };
 
+  const checkIfMessageComesFromSameDomain = async (domain) => {
+    const path = Routing.getCurrentPath();
+    const pathData = await Navigation.getNavigationPath(
+      LuigiConfig.getConfigValueAsync('navigation.nodes'),
+      path
+    );
+    const currentNavigationPath = pathData.navigationPath[pathData.navigationPath.length - 1];
+    const currentViewUrl = currentNavigationPath.viewUrl;
+    const sameDomain = currentViewUrl && (currentViewUrl.startsWith(domain) || currentViewUrl.startsWith('/'));
+    return sameDomain;
+  };
+
   export default {
     data() {
       return Object.assign(
@@ -304,14 +316,16 @@
           }
         });
         window.addEventListener('message', async e => {
+          const eventDomain = e.origin;
+          const isSameDomain = await checkIfMessageComesFromSameDomain(eventDomain);
           const allIframes = Iframe.getAllIframes(node);
-          if (allIframes.length > 0) {
-            const isTheSameOrigin = (iframe) => {
+          if (isSameDomain && allIframes.length > 0) {
+            const checkIfMessageComesFromRegisteredIframe = (iframe) => {
               return (e.source === iframe.contentWindow || (this.get().modalIframe && e.source === this.get().modalIframe.contentWindow))
             }
-            const iframe = allIframes.find(isTheSameOrigin);
+            const iframe = allIframes.find(checkIfMessageComesFromRegisteredIframe);
             if (iframe) {
-              iframe.viewDomain = e.origin;
+              iframe.viewDomain = eventDomain;
               if ('luigi.get-context' === e.data.msg) {
                 if (
                   this.get().modalIframe &&
--- a/core/src/App.html
+++ b/core/src/App.html
@@ -301,25 +301,18 @@
           }
         });
         window.addEventListener('message', async e => {
-          const allMessagesSources = [...Iframe.getAllIframes(node, this.get().modalIframe), {contentWindow: window, luigi: {viewUrl: window.location.href}}];
-          const isMessageSource = (source) => source && (e.source === source.contentWindow);
-          const iframe = allMessagesSources.find(isMessageSource);
-          if (!(iframe && iframe.luigi && iframe.luigi.viewUrl)) return;
 
-          const navigateOkMsg = 'luigi.navigate.ok' === e.data.msg;
-          if (navigateOkMsg && !iframe.luigi.nextViewUrl) return;
-          const viewUrl = navigateOkMsg ? iframe.luigi.nextViewUrl : iframe.luigi.viewUrl;
+          const iframe = this.getValidMessageSource(e);
+          if (!iframe) return;
 
-          if (!iFrameHelpers.iframeIsSameDomain(viewUrl, e.origin)) return;
-
-          if (navigateOkMsg) {
+          if ('luigi.navigate.ok' === e.data.msg) {
             iframe.luigi.viewUrl = iframe.luigi.nextViewUrl;
             iframe.luigi.nextViewUrl = '';
             config.navigateOk = true;
           }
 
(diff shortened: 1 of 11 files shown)


core/src/App.html
```diff
@@ -295,181 +295,189 @@
...
-                  'currentNode.loadingIndicator.hideAutomatically'
-                ) !== false;
-              if (loadingIndicatorAutoHideEnabled) {
-                this.set({ showLoadingIndicator: false });
-              }
-            }
-          }
+          this.set({url: e.origin})
+          const allIframes = Iframe.getAllIframes(node);
+          if (allIframes && allIframes.length > 0) {
```
jesusreal: I would suggest avoiding more nesting by doing the following:
```
 if (!allIframes || allIframes.length === 0) { return }
// here code currently inside the if block
```

jesusreal: Additionally, I would enforce  `Iframe.getAllIframes` to return always an array to make checks easier: `if(allIframes.length > 0)`

jesusreal: I assume the forEach block will find zero or one Iframe that will match. Therefore I would use `some` instead and keep the code outside of the loop to avoid another level of nesting. 
Following up on the last 2 comments, I would refactor it to sth like that:
```
const validSource = allIframes.some(() => e.source === iframe.contentWindow || (this.get().modalIframe && e.source === this.get().modalIframe.contentWindow))
if (!validSource) { return }
// here code currently inside the foreach block
```

jesusreal: I saw the following piece of code used 3 times, maybe we can create a variable out of it:
`this.get().modalIframe && e.source === this.get().modalIframe.contentWindow`

dariadomagala: Regarding the 1st and 3rd comment - after `if` block there is still one function that needs to be executed, so returning here wouldn't make sense. I left it in this `if` block.

jesusreal: Correct me if I am wrong, but I think something like that should work:
```
          const isOriginWindow = (iframe) => e.source === iframe.contentWindow;
          const isOriginModal = () => this.get().modalIframe && e.source === this.get().modalIframe.contentWindow;

          const iframe = Iframe.getAllIframes(node)
            .find((iframe) => isOriginWindow(iframe) || isOriginModal());

          if (!iframe) { return };
          // Remaining code
```
Benefits of this:
- more reusability of code than with `isTheSameOrigin()`
- two levels less of indentation for 100+ lines of code ;)

core/src/App.html
```diff
@@ -239,6 +239,19 @@
...
+    const path = Routing.getCurrentPath();
+    const pathData = await Navigation.getNavigationPath(
+      LuigiConfig.getConfigValueAsync('navigation.nodes'),
+      path
+    );
+    if (pathData && pathData.navigationPath && pathData.navigationPath[pathData.navigationPath.length - 1].viewUrl) {
+      const currentViewUrl = pathData.navigationPath[pathData.navigationPath.length - 1].viewUrl;
+      const sameDomain = currentViewUrl && (currentViewUrl.startsWith(domain) || currentViewUrl.startsWith('/'));
+      iframe.trustedDomain = domain;
+    }
```
pekura: Should the trusted domain be removed from the iframe otherwise (else)?

core/src/App.html
```diff
@@ -304,181 +314,194 @@
...
+            return (e.source === iframe.contentWindow || (this.get().modalIframe && e.source === this.get().modalIframe.contentWindow))
           }
+          const iframe = allIframes.find(getTargetIframe);
+          if (iframe && iframe.luigiViewUrl) {
 
-          if ('luigi.show-loading-indicator' === e.data.msg) {
-            this.set({ showLoadingIndicator: true });
-          }
+            if ('luigi.navigate.ok' === e.data.msg) {
+              if (iframe.luigiNextViewUrl) {
```
pekura: Here we should check if the e.origin matches the luigiNextViewUrl (the domain part).

core/src/App.html
```diff
@@ -304,11 +318,27 @@
           }
         });
         window.addEventListener('message', async e => {
+          const allIframes = Iframe.getAllIframes(node);
+          const isOriginWindow = (iframe) => e.source === iframe.contentWindow;
+          const isOriginModal = () => this.get().modalIframe && e.source === this.get().modalIframe.contentWindow;
+          const getTargetIframe = (iframe) => {
+            return (isOriginWindow(iframe) || isOriginModal())
```
pekura: This is not quite right, is it? It will return ANY 'iframe' if isOriginModal(). If isOriginModal then it should return the modal iframe.

pekura:
We need to add     

iframe.luigiViewUrl = viewUrl;

to Modal.html as well (after iframe.src = viewUrl;).

core/src/services/iframe.js
```diff
@@ -91,6 +97,7 @@ export const navigateIframe = (config, component, node) => {
   } else {
     const goBackContext = component.get().goBackContext;
     config.iframe.style.display = 'block';
+    config.iframe.luigiNextViewUrl = viewUrl;
     config.iframe.contentWindow.postMessage(
```
pekura: This postMessage should also have a specific target set. It should be the domain part of the iframe.luigiViewUrl.

core/src/App.html
```diff
@@ -239,6 +241,18 @@
     };
   };
 
+  const checkIfUrlMatchesTheDomain = (viewUrl, domain) => {
+    return viewUrl && (viewUrl.startsWith(domain) || viewUrl.startsWith('/'));
```
pekura: This is too optimisitc :-) In case of a relative path we need to check if the domain == current location domain.

core/src/App.html
```diff
@@ -304,22 +318,34 @@
           }
         });
         window.addEventListener('message', async e => {
+          const allIframes = Iframe.getAllIframes(node, this.get().modalIframe);
+          const isOriginWindow = (iframe) => iframe && (e.source === iframe.contentWindow || e.source === window);
+          const iframe = allIframes.find(isOriginWindow);
+          if (!iframe || !iframe.luigiViewUrl) return;
+
+          if ('luigi.navigate.ok' === e.data.msg) {
```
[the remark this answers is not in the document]
pekura: Because luigi.navigate.ok needs to be processed first to set the luigi.viewUrl properly (based on the luigi.nextViewUrl).

core/src/App.html
```diff
@@ -304,22 +301,34 @@
           }
         });
         window.addEventListener('message', async e => {
+          const allIframes = Iframe.getAllIframes(node, this.get().modalIframe);
+          const isOriginWindow = (iframe) => iframe && (e.source === iframe.contentWindow || e.source === window);
+          const iframe = allIframes.find(isOriginWindow);
```
pekura: In case of a message coming from the current window we get here any/random iframe. I guess this case (message coming not from an iframe but from the core - which means current window) we need a better handling. This makes the whole thing not simpler though.. We could add an fake iframe to the list of frames in the beginning with .contentWindow set to current window and .luigi.viewUrl to the current window's url...

core/src/App.html
```diff
@@ -304,22 +301,34 @@
...
+          }
+
+          if (!iFrameHelpers.iframeIsSameDomain(iframe, e.origin)) return;
+
           if ('luigi.get-context' === e.data.msg) {
-            if (
-              this.get().modalIframe &&
-              e.source === this.get().modalIframe.contentWindow
-            ) {
+            if (isOriginWindow(this.get().modalIframe)) {
```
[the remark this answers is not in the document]
dariadomagala: Isn't it? As I understood it we want to receive messages from all active iframes.

core/src/App.html
```diff
@@ -304,22 +301,34 @@
           }
         });
         window.addEventListener('message', async e => {
+          const allIframes = [...Iframe.getAllIframes(node, this.get().modalIframe), {contentWindow: window, luigi: {viewUrl: window.location.href, trustedDomain: window.location.origin}}];
+          const isOriginWindow = (iframe) => iframe && (e.source === iframe.contentWindow);
```
jesusreal: I would rename `isOriginWindow`to sth like `isMessageSource`

core/src/App.html
```diff
@@ -304,22 +301,34 @@
...
+          }
+
+          if (!iFrameHelpers.iframeIsSameDomain(iframe, e.origin)) return;
+
           if ('luigi.get-context' === e.data.msg) {
-            if (
-              this.get().modalIframe &&
-              e.source === this.get().modalIframe.contentWindow
-            ) {
+            if (isOriginWindow(this.get().modalIframe)) {
```
jesusreal: I think we can use the same `if` statement below around line 377, which currently is:
```
         if (
              this.get().modalIframe &&
              e.source === this.get().modalIframe.contentWindow
            ) {
```

core/src/App.html
```diff
@@ -304,22 +301,34 @@
...
                 iframe: this.get().modalIframe,
-                context: ctx,
+                context: ctx, 
                 pathParams: this.get().modalIframeData.pathParams,
                 nodeParams: this.get().modalIframeData.nodeParams,
                 modal: true
               };
               sendContextToClient(this, modalConfig, {});
-            } else if (config.iframe) {
+            } else if (isOriginWindow(iframe)) {
```
jesusreal: As iframe really contains the source of the message, I would say here we just need `else`.  Above we are already extracting the iframe from all the iframes using `isOriginWindow` so we are sure `iframe` here is the right one and we don't need to check again.

core/src/utilities/helpers/iframe-helpers.js
```diff
@@ -104,3 +104,30 @@ export const getVisibleIframes = () => {
     .call(document.querySelectorAll('iframe'))
     .filter(item => item.style.display !== 'none');
 };
+
+export const getValidIframe = (event, window) => {
+  const iframes = document.querySelectorAll('.iframeContainer iframe');
```
jesusreal: Maybe we should not use `querySelectorAll`anymore for iframes, since now we have modal iframes, which are in a different container. Additionally, do we need to consider here also the Luigi Core window in case the message is coming from Luigi Core?
If we should not use that anymore, then we have to remove it also from 'getVisibleIframes` in this file?

core/src/utilities/helpers/iframe-helpers.js
```diff
@@ -104,3 +104,30 @@ export const getVisibleIframes = () => {
...
+  }
+};
+
+export const iframeIsSameDomain = (iframe, domain) => {
+  const sameDomain = urlMatchesTheDomain(iframe.luigi.viewUrl, domain);
+  iframe.luigi.trustedDomain = sameDomain ? domain : '';
+  return sameDomain;
+};
+
+export const sendPostMessage = (iframe, message) => {
```
jesusreal: I would rename this to sth like 'sendMessageToIframe`, since `Post` is an implementation detail and this method is not valid if Luigi core sends a message to itself.

core/src/utilities/helpers/iframe-helpers.js
pekura: The way we use it it should rather be isValidMessageSource(e) and should return true if either:
e.source === window
or
there is an iframe in the .iframeContainer where iframe.contentWindow === e.source
or
we have modalIframe and modalIframe.contentWindow === e.source.

core/src/services/iframe.js
```diff
@@ -75,35 +80,41 @@ export const navigateIframe = (config, component, node) => {
       config.navigateOk = undefined;
       config.iframe = document.createElement('iframe');
       config.iframe.src = viewUrl;
+      config.iframe.luigi = config.iframe.luigi || {};
+      config.iframe.luigi.viewUrl = viewUrl;
+      const trustedIframeDomain = viewUrl.startsWith('/')
+        ? window.location.origin
+        : IframeHelpers.getLocation(viewUrl).origin;
```
pekura: Please remove .origin!

core/src/services/iframe.js
```diff
@@ -75,35 +80,41 @@ export const navigateIframe = (config, component, node) => {
...
+          const message = ['init', JSON.stringify(componentData.context)];
+          IframeHelpers.sendMessageToIframe(config.iframe, message);
         });
       }
     }
   } else {
     const goBackContext = component.get().goBackContext;
+    const trustedIframeDomain = config.iframe.luigi.viewUrl.startsWith('/')
+      ? window.location.origin
+      : IframeHelpers.getLocation(config.iframe.luigi.viewUrl).origin;
```
pekura: Please remove .origin!

PR #469, opened 2019-04-04
ios scroll

JohannesDoberer:


JohannesDoberer:
testing:
change package.json in sample angular-app:
"start": "concurrently \"npm run buildConfig -- --watch\" \"ng serve --host 0.0.0.0\""

On iphone safari [IP]:4200

PR #479, opened 2019-04-12
Luigi version bump in sample app

pekura:


pekura:
> I don´t know if we should add it to vue app as well, we are not maintaining it anyway...

We would need to take a look at the vue app as it has not been updated in a long time. This we can do but in a separate task I would suggest.

PR #480, opened 2019-04-12
Alert improvements

parostatkiem:
**Description**

- Alerts are queued
- a `closeAfter` property can be provided via Luigi API to (surprise!) close the Alert after the desired time
- Client API part is backwards compatible
- add UI tests for the new features
- update Client API docu
- showing Alert from Core is not backwards-compatible so the only place where it is used needed to be tuned up a bit

**Related issue(s)**


core/src/App.html
```diff
@@ -604,14 +604,37 @@
...
+          console.error(
+            `The alert with id '${
+              settings.id
+            }' already exists in a queue, therefore it won't be displayed `
+          );
+          return Promise.reject();
+        }
+
+        if (settings.closeAfter) {
+          setTimeout(() => {
```
pekura: If the alert gets closed manually before the timeout an error is shown in the console: "An unexisting alert has been dismissed.". There is nothing wrong with closing an alert manually so there should not be any error shown.

parostatkiem: I'll investigate it.

client/src/luigi-client.js
```diff
@@ -1,6 +1,7 @@
+var crypto = require("crypto");
```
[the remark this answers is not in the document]
parostatkiem: Done

Issue #491, opened 2019-04-17
Path routing broken for some micro-frontends


y-kkamil:
**Description**

If the micro-frontend is configured to use path routing instead of hash routing, Luigi will perform full-page iframe reload instead of letting the micro frontend's router take care of navigation.

**Expected result**

Micro frontend SPA performs navigation without page being reloaded.

**Troubleshooting**

Looks like `luigi.navigate` event is performing `window.location.replace` every time navigation event occurs. This is bad :).

kwiatekus:
detection of mf routing method is required.
if path routing used, use history.push api

y-kkamil:
Related: https://github.com/kyma-project/luigi/issues/452
Needs to be fixed, to make https://github.com/kyma-project/console/issues/750 fully working

y-kkamil:
Fixed for Angular/Vue/React routers in #500 . Bear in mind that merging this solution will break catalog/instances views in Console because of https://github.com/kyma-project/console/issues/798

PR #500, opened 2019-04-26
Fix path routing

y-kkamil:
**Description**

- Fixes the problem with reloaded micro frontend view when using path routing

**Related issue(s)**

https://github.com/kyma-project/console/issues/798, which includes the steps to reproduce the issue.


y-kkamil:
Fixes #491

y-kkamil:
Checked for angular/vue/react routers. Please test for any use case you can think of. ;)
Unfortunately, https://github.com/kyma-project/console/issues/798 still persists.

jesusreal:
This ticket is still blocked until https://github.com/kyma-project/console/issues/798  is solved.

y-kkamil:
react-router uses a library called [history](https://github.com/ReactTraining/history/) internally. https://github.com/ReactTraining/history/blob/master/modules/DOMUtils.js#L53 --> this is the line that was causing problems for console. Adding state to `PopStateEvent` makes the expression evaluate to false, which in turn let's the router take care of route change.

PR #501, opened 2019-04-29
Luigi Authentication Events

maxmarkus:
Implemented and created documentation for authentication events:

```
auth: {
  events: {
    onAuthSuccessful: (settings, authData) => {},
    onAuthError: (settings, err) => {}
    onAuthExpired: (settings) => {},
    onLogout: (settings) => {}
  }
}
```
Refactored Luigi Core API out to single classes.


docs/authorization-events.md
```diff
@@ -0,0 +1,31 @@
+# Authorization events
+
+Luigi provides life-cycle events that can be triggered by Luigi internally, or by authorization providers.
+Events are part of the **auth** configuration object and require to be functions, they can be executed asynchronous.
+
+By returning `false` value in those functions, the default behaviour can be disabled and lifecycle execution stops with the function. Be aware that this might lead to blank pages in logged out state, since most of the time there is a redirect to either a logout, login or main page.
+
```
bszwarc: I would put the code sample right after the second sentence. So we have:

`Events are part of the **auth** configuration object and require to be functions, they can be executed asynchronous.`
Then maybe an introductory line for the sample.
`An example events configuration looks as follows:`
<code sample here>

Then I would take the first two sentences from the ## Events section so that we have the description of the first and second parameter (what can it be). At the end, we can have some info on disabling the default behavior by returning false :)  

So basically the ##Events section would only have the description of functions, since the other, more descriptive information would go with the code sample.

What do you think about it?

docs/authorization-configuration.md
```diff
@@ -121,4 +121,26 @@ When Luigi fails to renew the token and then logs out the user, it adds the foll
 - `enabled` means third party cookies are supported by the browser.
 - `not_checked` means that the script was not provided in **thirdPartyCookiesScriptLocation** or it could not be loaded.
 
-The application developer can read these parameters and set a logout page based on them.
\ No newline at end of file
+The application developer can read these parameters and set a logout page based on them.
+
+### Custom Auth Provider
+
+You can write your own Auth Provider that meets your requirements.
```
bszwarc:
```suggestion
You can write your own authentication provider that meets your requirements. 
```

core/src/Authorization.html
```diff
@@ -81,23 +81,42 @@
         isHidden: false
       };
     },
-    oncreate() {
+    async oncreate() {
       if (!LuigiConfig.isAuthorizationEnabled()) {
```
pekura: You should use the isAuthorizationEnabled() from auth (not config, as this is deprecated now).

core/src/core-api/auth.js
```diff
@@ -0,0 +1,31 @@
...
+  isAuthorizationEnabled() {
+    const idpProviderName = config.getConfigValue('auth.use');
+    const idpProviderSettings = config.getConfigValue(
+      `auth.${idpProviderName}`
+    );
+    return !!idpProviderSettings;
+  }
+
+  async handleAuthEvent(eventName, providerInstanceSettings, data, redirectUrl) {
+    const result = await config.executeConfigFnAsync('auth.events.' + eventName, providerInstanceSettings, data);
```
pekura: If the event handler function is not defined this will lead to an error in browser console: 
Bildschirmfoto 2019-05-06 um 11 18 16
As these event handlers are optional this should not lead to an error. We either test first if there is an event handler fn defined and then execute it here or we build this check into the executeConfigFnAsync; maybe with an 'optional' flag so that it fails if the function is required in cases where it is required.

core/src/core-api/auth.js
```diff
@@ -0,0 +1,31 @@
...
+    let redirect = result === undefined || !!result;
```
pekura: Also, it will throw an exception in case of a rejected promise and the code after 'await' won't be executed.

docs/authorization-configuration.md
```diff
@@ -120,4 +120,26 @@ When Luigi fails to renew the token and then logs out the user, it adds the foll
...
+### Implement a Custom Authentication Provider
+
+You can write your own authentication provider that meets your requirements. 
+
+[oAuth2ImplicitGrant.js](src/providers/oAuth2ImplicitGrant.js) is a good starting point if you don't use an external authentication library.
+After authorization is successful on the auth provider's side it redirects back to `Luigi callback.html` **redirect_uri**. The auth provider verifies the authentication data, saves it in  **localStorage** for Luigi, and redirects to the Luigi main page. 
+
+[openIdConnect.js](src/providers/openIdConnect.js) lazy loads the official `oidc-client` library and is a good starting point if you also depend on external authentication libraries.
+
+Make sure to set the following data in your Auth Provider implementation after authentication.
```
bszwarc:
```suggestion
Make sure to set the following data in your Authentication Provider implementation, so that it is used after successful authentication.
```
There was something missing in this one - but verify it, please :) And correct if I understand it wrongly.

maxmarkus:
`onAuthError` is not working anymore due to some changes that got merged in. small refactoring upcoming.

PR #503, opened 2019-04-30
Display info in bottom left corner

parostatkiem:
**Description**

- Display a configurable string at the bottom of LeftNav
- It follows the design ~and is sticky~
- Add e2e test for the new feature
- Update docu of the new luigi-config setting

**Related issue(s)**

### DIFF
--- a/core/src/navigation/LeftNav.html
+++ b/core/src/navigation/LeftNav.html
@@ -139,6 +139,11 @@
     {/if} {/each}
   </nav>
   {/if}
+  {#if isFooterVisible()}
+  <div class="lui-sideNavFooter fd-has-padding-small fd-has-type-minus-1">
+    <span>Version 1.5.5</span>
+  </div>
+  {/if}
 </div>
 
 <script type="text/javascript">
@@ -178,6 +183,11 @@
         return (
           expandedList && expandedList.indexOf(nodes.metaInfo.categoryUid) >= 0
         );
+      },
+      isFooterVisible() {
+        return true;
+        // respect also 'responsiveNavigation' setting in the future.
+        // the footer should also be visible if it's' possible to collapse the Nav
       }
     },
     methods: {
@@ -290,5 +300,13 @@
     &.sap-icon--m::before {
       width: 1.25rem;
     }
+  }
+
+  .lui-sideNavFooter {
+    position: sticky;
+    bottom: 0;
+    left: 0;
+    color: var(--fd-color-text-4);
+    background-color: var(--fd-color-background-2);
   }
 </style>
--- a/core/src/navigation/LeftNav.html
+++ b/core/src/navigation/LeftNav.html
@@ -139,12 +139,10 @@
     {/if} {/each}
   </nav>
   {/if}
-  {#if isFooterVisible(footerText)}
-  <div class="lui-sideNavFooter fd-has-padding-small fd-has-type-minus-1">
-    {#if footerText}
-    <span>{footerText}</span>
-    {/if}
-  </div>
+  {#if footerText}
+  <span
+    class="lui-sideNavFooter fd-has-type-minus-1"
+  >{footerText}</span>
   {/if}
 </div>
 
@@ -245,6 +243,32 @@
     -webkit-box-shadow: 1px 1px 2px 0 rgba(0, 0, 0, 0.05);
     box-shadow: 1px 1px 2px 0 rgba(0, 0, 0, 0.05);
   }
+  $footerHeight: calc(0.85714rem + (2 * 16px));
+  /* text height + top and bottom padding */
+
+  .fd-side-nav {
+    min-height: 100%;
+    margin-bottom: calc(-1 * #{$footerHeight});
+  }
+
+  .fd-side-nav::after {
+    content: '';
+    display: block;
+  }
+
+  .lui-sideNavFooter,
+  .fd-side-nav:after {
+    height: $footerHeight;
+  }
+
+  .lui-sideNavFooter {
+    display: inline-block;
+    width: 100%;
+    color: var(--fd-color-text-4);
+
+    padding: 16px 22px;
+  }
+
   .fd-side-nav__link {
     .sap-icon--action {
       //external-link icon
@@ -304,13 +328,5 @@
     &.sap-icon--m::before {
       width: 1.25rem;
     }
-  }
-
-  .lui-sideNavFooter {
-    position: sticky;
-    bottom: 0;
-    left: 0;
-    color: var(--fd-color-text-4);
-    background-color: var(--fd-color-background-2);
   }
 </style>
--- a/core/src/navigation/LeftNav.html
+++ b/core/src/navigation/LeftNav.html
@@ -141,7 +141,7 @@
   {/if}
   {#if footerText}
   <span
-    class="lui-sideNavFooter fd-has-type-minus-1"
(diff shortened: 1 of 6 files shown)


docs/general-settings.md
```diff
@@ -20,3 +21,4 @@ settings: {
 * **header.title** defines the top left navigation title.
 * **header.favicon** defines the favicon. It rquires a standard favicon file with the `.ico` extension, and 16x16px or 32x32px dimensions.
 * **responsiveNavigation** allows adding a button on the left side of the top navigation. Upon click, the button shows or hides the left navigation. The possible values are `simple` and `simpleMobileOnly`. `simple` displays the button regardless of the browser window´s size, while `simpleMobileOnly` shows the button when the width is lower than `600px`. If you don't specify any value for  **responsiveNavigation**, the button will not be displayed. The same applies when you enable **hideSideNav** for the current active navigation node. 
+* **sideNavFooterText** a string that is displayed in a sticky footer inside Side Navigation. Good place to display the version of your application.
```
maxmarkus: It can also be a function (not async) since we are using `getConfigValue` to fetch it. Do we want to mention this?

core/src/navigation/LeftNav.html
```diff
@@ -291,4 +305,12 @@
       width: 1.25rem;
     }
   }
+
+  .lui-sideNavFooter {
+    position: sticky;
```
maxmarkus: `fixed` instead of `sticky`? It seems to be placed not at the end of the navbar.
screen markus e 2019-04-30 at 15 10 07

core/src/navigation/LeftNav.html
```diff
@@ -139,6 +139,13 @@
     {/if} {/each}
   </nav>
   {/if}
+  {#if isFooterVisible(footerText)}
+  <div class="lui-sideNavFooter fd-has-padding-small fd-has-type-minus-1">
+    {#if footerText}
```
maxmarkus: Do we need the container at all, if `footerText` is undefined? else I would put it before `isFooterVisible`.

core/src/navigation/LeftNav.html
parostatkiem: Reading the conversation on Slack about this, I understood it as `sticky`.

core/src/navigation/LeftNav.html
parostatkiem: I did the implementation (this template + `isFooterVisible` function) with having the future placement of SideNav close button on mind. It's visible on Zeplin. Lydia says it's a concept and likely be introduced in the future. 
image

Furthermore, who knows - maybe we'll put there something else in the future? 
In general, I'm against useless containers (!) but this is a special case in my opinion.

docs/general-settings.md
parostatkiem: Good point, I'll mention it.

parostatkiem: Sadly, it can't be. Checked with arrow and normal function.
image

core/src/navigation/LeftNav.html
parostatkiem: ... but looking at the Zeplin, it's fixed. I'm confused 😕

core/src/navigation/LeftNav.html
```diff
@@ -209,6 +221,9 @@
 
 <style type="text/scss">
   $topNavHeight: 48px;
```
marynaKhromova:
```suggestion
  $topNavHeight: 48px;
  $side-nav-footer--padding: 13px;
  $side-nav-footer--font-size: 12px;
  $side-nav-footer--line-height: 16px;
```

Let's use some variables. Check my comments below.

core/src/navigation/LeftNav.html
```diff
@@ -231,24 +246,46 @@
...
+      }
     }
   }
+
+  .lui-sideNavFooter {
+    display: inline-block;
+    width: 100%;
+    color: var(--fd-color-text-4);
+    /* height: $footerHeight; */
+    padding: 13px 22px;
```
marynaKhromova:
```suggestion
    padding: $side-nav-footer--padding 20px;
```

Let's use a variable here. And the navigation items have 20px padding. I think it's better to have the same here.

core/src/navigation/LeftNav.html
```diff
@@ -231,24 +246,46 @@
...
+      &.fd-side-nav__link--icon {
+        padding-left: 16px;
+      }
+      .fd-side-nav__icon {
+        width: 2em;
+      }
     }
   }
+
+  .lui-sideNavFooter {
```
marynaKhromova:
```suggestion
  .lui-sideNavFooter {
     font-size: $side-nav-footer--font-size;
     line-height: $side-nav-footer--line-height;
```
Let's discuss it. I would go here with the solution to use our sizes (variables) and not rely on class from Fiori Fundamentals. If they change tomorrow `font-size` and `line-height` for class `fd-has-type-minus-1`, then our implementation will be broken. What do you think?

About `lui-sideNavFooter` name. I'm not sure about conventions to the naming. Do we use camel case css classes?

core/src/navigation/LeftNav.html
```diff
@@ -139,6 +139,11 @@
     {/if} {/each}
   </nav>
   {/if}
+  {#if footerText}
+  <span
+    class="lui-sideNavFooter fd-has-type-minus-1"
```
marynaKhromova:
```suggestion
    class="lui-sideNavFooter"
```

In case we decide to use variables. Check my comment below for row `281`

core/src/navigation/LeftNav.html
```diff
@@ -231,24 +246,46 @@
...
+        width: 2em;
+      }
     }
   }
+
+  .lui-sideNavFooter {
+    display: inline-block;
+    width: 100%;
+    color: var(--fd-color-text-4);
+    /* height: $footerHeight; */
```
marynaKhromova:
```suggestion

```

core/src/navigation/LeftNav.html
```diff
@@ -209,6 +221,9 @@
 
 <style type="text/scss">
   $topNavHeight: 48px;
+  $footerHeight: calc(0.85714rem + (2 * 13px) + 3px);
```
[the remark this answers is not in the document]
parostatkiem: I did it this way because I wanted to give a *clear* information how this height is calculated. 
`0.85714rem ` is the _font-size_ value for Fundamental's _minus-1_ font. 
In fact, I could use one fixed magic number and I wouldn't have any scrolling issue ;)

On the other hand, as you noticed, it would be nice to use variables. To make it sensible, I'd have to get them from Fundamental which means importing another SCSS modules. It's pointless in my opinion. Also, declaring those variables manually gives no advantage over a simple comment line which explains everything.

core/src/navigation/LeftNav.html
parostatkiem: I totally agree with the renaming, I'm doing it right now.

About defining everything apart from Fundamental - I think it might produce the opposite result from what you described. 95% of Luigi frontend relies on it. As I heard one day, _we should stick to Fundamental styles_. That's why I based the whole footer on it, doing as little custom styles as possible :)

docs/general-settings.md
klaudiagrz:
```suggestion
* **sideNavFooterText** is a string displayed in a sticky footer inside the Side Navigation. It is a good place to display the version of your application.
```

PR #507, opened 2019-05-03
Improve node and category ordering

pekura:
The nodes are now rendered in the order as specified in the configuration.


core/src/navigation/LeftNav.html
```diff
@@ -1,7 +1,7 @@
 <div class="fd-app__sidebar {hideNavComponent ? 'hideNavComponent' : ''}">
   {#if children && pathData.length > 1}
   <nav class="fd-side-nav" on:keyup="handleKey(event)">
-    {#each Object.entries(children) as [key, nodes]} {#if key === 'undefined'}
+    {#each sortedChildrenEntries as [key, nodes]} {#if key === 'undefined' || key.indexOf('___') === 0}
```
maxmarkus: I would love to see this as a constant `virtualGroupPrefix` or similar.

core/src/utilities/helpers/navigation-helpers.js
```diff
@@ -62,8 +62,11 @@ const orderNodes = nodes => {
   });
 };
 
-export const groupNodesBy = (nodes, property) => {
+export const groupNodesBy = (nodes, property, useVirtualGroups) => {
```
maxmarkus: Is it necessary to add `useVirtualGroups` parameter if it is only used in one file which uses that flag with `true` anyways?

pekura: We are not sure yet if we provide a setting for it or not. I f yes, we want to have the "all-ungrouped-nodes-on-top" functionality there as well.

PR #508, opened 2019-05-03
Improve compatibilty script

maxmarkus:
small improvements on compatibility script
new `--test-only` to re-run test to skip installing and bundling again. similar to /retest


scripts/testCompatibility.sh
```diff
@@ -23,136 +30,204 @@ function showHelp {
...
+  linkLuigi
+  bundleLuigi
+  verifyAndStartWebserver
+  startE2eTestrunner
+fi # --test-only
+
+if [ "true" == "$TESTONLY" ]; then
+  echoe "Running bunded example and e2e tests"
+  verifyAndStartWebserver
+  startE2eTestrunner
```
marynaKhromova: If `verifyAndStartWebserver` and `startE2eTestrunner` are included in both `if` maybe it make sense to move it out and use once.

maxmarkus: Yes, makes sense, I just explicitly put it there to make it readable easier.

PR #526, opened 2019-05-14
Bug in left navigation

marynaKhromova:

- fix the bug
- improve the way to hide navigation
- add `constants.js` file
- add `viewport` metatag to see the mobile navigation

### DIFF
--- a/core/src/App.html
+++ b/core/src/App.html
@@ -1,3 +1,4 @@
+<svelte:window on:resize="onResize()"/>
 <div
   id="app"
   class="{hideNav? 'no-nav' : ''} {hideSideNav? 'no-side-nav' : ''}"
@@ -65,6 +66,7 @@
   import * as iFrameHelpers from './utilities/helpers/iframe-helpers';
 
   const removeQueryParams = str => str.split('?')[0];
+  const mobileBreakpoint = 599;
 
   const isValidBackRoute = (preservedViews, routeHash) => {
     if (preservedViews.length === 0) {
@@ -285,6 +287,7 @@
           builderCompatibilityMode: Boolean(window.builderCompatibilityMode),
           isolateAllViews
         };
+
         window.addEventListener('popstate', async e => {
           const alertQueue = this.get().alerts;
           if (!alertQueue || !alertQueue.length) return;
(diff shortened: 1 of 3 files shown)


core/src/App.html
```diff
@@ -545,9 +547,37 @@
         });
       },
       onResize() {
-        if (window.innerWidth >= CSS_BREAKPOINTS.desktopMinWidth) {
+        const simpleDesktopToMobile = () => {
+          if (this.get().responsiveNavSetting == 'simple') {
+            if (
+              window.innerWidth < CSS_BREAKPOINTS.desktopMinWidth &&
+              this.get().previousBreakpoint >= CSS_BREAKPOINTS.desktopMinWidth
```
jesusreal: May `previousWindowWidth` is a better name than 'previousBreakpoint`

core/src/App.html
```diff
@@ -545,9 +547,37 @@
         });
       },
       onResize() {
-        if (window.innerWidth >= CSS_BREAKPOINTS.desktopMinWidth) {
+        const simpleDesktopToMobile = () => {
+          if (this.get().responsiveNavSetting == 'simple') {
```
jesusreal: Please always use `===` instead of `==`. You will be always be right using `===`

core/src/App.html
```diff
@@ -545,9 +547,37 @@
...
+            if (
+              window.innerWidth < CSS_BREAKPOINTS.desktopMinWidth &&
+              this.get().previousBreakpoint >= CSS_BREAKPOINTS.desktopMinWidth
+            ) {
+              this.closeLeftNav();
```
jesusreal: Correct me if I am wrong, but `closeLeftNav` is closing the _mobile_ left nav. In that case, we could rename the method to sth. like `closeMobileLeftNav` ;)

core/src/App.html
```diff
@@ -545,9 +547,37 @@
...
+
+        if (
+          window.innerWidth >= CSS_BREAKPOINTS.desktopMinWidth &&
+          simpleIsNotSameBreakpoint()
+        ) {
           this.closeLeftNav();
         }
+
+        simpleDesktopToMobile();
+        this.get().previousBreakpoint = window.innerWidth;
```
jesusreal: Please use better ```this.set({ previousBreakpoint: window.innerWidth });```. Otherwise we are doing a mutation of the component state. It is better to let Svelte do the state updates itself, by just telling Svelte what we want to update with `this.set`.

core/src/App.html
marynaKhromova: As discussed, it's not only for mobile navigation but as well for desktop in `simple` mode.

PR #528, opened 2019-05-15
Error message for invalid auth provider

### LINKED ISSUE #510: Invalid auth provider should throw an error
If I accidentally enter a wrong auth provider name `mockAuth2`, no warning or error is shown. 
The checks are not reached due to `isAuthorizationEnabled` returning already before reaching that point in `Authorization.html`.

maxmarkus:
Fixes #510


core/src/Authorization.html
```diff
@@ -194,9 +194,9 @@
             checkAuth();
           })
           .catch(err => {
-            console.error(
-              `Could not instantate ${idpProviderName} provider : ${err}`
-            );
+            const errorMsg = `Error: ${err.message || err}`;
+            console.error(errorMsg, err);
```
jesusreal: I think this can be just `console.error(errorMsg);`

core/src/services/config.js
```diff
@@ -91,9 +95,7 @@ class LuigiConfigManager {
    * @returns {boolean} returns true if authorization is enabled. Otherwise returns false.
    */
   isAuthorizationEnabled() {
-    const idpProviderName = this.getConfigValue('auth.use');
```
jesusreal: This check is now different, before we were checking `auth.use` and the auth settings, now we are checking only `auth.use`. I think we should still check if the auth settings are also added to the Luigi configuration.

core/src/Authorization.html
jesusreal: This part is inside an if block:
```
      if (GenericHelpers.isPromise(idpProviderInstance)) {
```
Looks like we will not display the error if the `idpProviderInstance` is not a promise.

maxmarkus: I updated the implementation to output the second err (object) only in case there is a message found.
If the Provider does not return a Promise, it is up to the provider itself to throw any required error.

core/src/services/config.js
maxmarkus: The implementation was wrong before, since it never reached the checks vor provider existence and validity inside _Authorization_ component, as it returned too early. As soon as there is some `auth.use` set, we can assume authorization should be used.

PR #536, opened 2019-05-16
Enable Luigi core API to instruct Luigi to do actions

pekura:


core/src/main.js
```diff
@@ -8,10 +8,14 @@ const configReadyCallback = () => {
...
   }
 
-  new App({
+  const app = new App({
     target: document.querySelector('body'),
     data: {}
   });
+
+  Luigi.showAlert = settings => {
+    app.showAlert(settings);
```
maxmarkus: I like the general approach of using the app component, but it makes the core api cluttered around the code base. We have a folder `core-api` which should consist of all functionalities.  Is it possible to refactor it like that?

PR #537, opened 2019-05-16
Replace old repo url

maxmarkus:
Replaced to reflect new SAP/luigi.git urls.


scripts/testCompatibility.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # We're considering this file is located in ~/Sites/kyma-project and
```
jesusreal: Please also replace the path in the comment 
```suggestion
# We're considering this file is located in ~/Sites/SAP and
```

maxmarkus: I wanted to leave it like is initially, but on second thought it does not make sense to leave some kyma trace in here. Committing.

PR #542, opened 2019-05-21
Documented luigi.auth.tokenIssued event

maxmarkus:
added documentation for luigi.auth.tokenIssued event for custom provider.


docs/authorization-configuration.md
```diff
@@ -143,3 +143,12 @@ const data = {
 localStorage.setItem('luigi.auth', JSON.stringify(data));
 localStorage.setItem('luigi.newlyAuthorized', true);
 ```
+
+Additionally, if you process authentication data during Luigi runtime (inside the custom provider, similar to openIdConnect provider), you should additionally dispatch the **luigi.auth.tokenIssued** event to update currently opened micro-frontends with the latest authentication data. This is not required when processing authentication outside of Luigi (similar to oAuth2ImplicitGrant provider, that processes the data in callback.html and redirects to Luigi afterwards).
```
bszwarc:
```suggestion
Additionally, if you process authentication data during Luigi runtime (inside the custom provider, similarly to using the`openIdConnect` provider), dispatch the `luigi.auth.tokenIssued` Event to update the currently opened micro frontends with the latest authentication data. This is not required when processing authentication outside Luigi, for example when `oAuth2ImplicitGrant` provider processes the data in `callback.html` and redirects to Luigi afterward.
```

PR #546, opened 2019-05-22
Modernize luigi client

maxmarkus:
Split up Luigi client to separate file per manager and refactored to classes.


docs/luigi-client-api.md
```diff
@@ -209,6 +207,92 @@ LuigiClient.linkManager().goBack(true);
 
 ## uxManager
 
+**Extends LuigiClientBase**
```
jesusreal: Looks like this part is now duplicated

client/src/luigi-client.js
```diff
@@ -71,13 +67,13 @@ function luigiClientInit() {
...
 
-  function setAuthData(eventPayload) {
+  const setAuthData = eventPayload => {
     if (eventPayload) {
       authData = eventPayload;
     }
-  }
+  };
 
   window.addEventListener('message', function messageListener(e) {
```
jesusreal: As discussed, let´s have a unique event listener for `message` event ;)

client/src/luigi-client.js
```diff
@@ -134,20 +113,20 @@ function luigiClientInit() {
...
 
 luigiClientInit();
 
-/** @namespace */
+/** @private */
 const LuigiClient = {
   /**
-   * Use the functions and parameters to define the lifecycle of listeners, navigation nodes, and Event data.
-   * @name lifecycle
+   * Use the functions and parameters to define the Lifecycle of listeners, navigation nodes, and Event data.
```
jesusreal: As discussed, let´s move the lifecycle methods to its own class ;)

client/src/luigi-client.js
```diff
@@ -187,394 +166,46 @@ const LuigiClient = {
...
-        promises.alerts[settings.id].promise = new Promise(resolve => {
-          promises.alerts[settings.id].resolveFn = resolve;
-        });
-        return promises.alerts[settings.id].promise;
-      }
-    };
-  }
+  uxManager: () => uxManagerInstance
 };
 export default LuigiClient;
```
jesusreal: As discussed, we could try to make it a class and export here a unique instance

client/src/uxManager.js
```diff
@@ -0,0 +1,189 @@
+import { LuigiClientBase } from './baseClass';
+import { randomBytes } from 'crypto';
```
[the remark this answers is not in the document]
maxmarkus: crypto is a standard package coming with NodeJs and does not need to be installed. https://nodejs.org/api/crypto.html

client/src/luigi-client.js
maxmarkus: Will be done in separate PR for this ticket, for easier review.

client/src/luigi-client.js
maxmarkus: done

client/src/luigi-client.js
maxmarkus: done

client/src/luigi-client.js
jesusreal: PR: https://github.com/SAP/luigi/pull/559

client/src/linkManager.js
```diff
@@ -0,0 +1,220 @@
+import { LuigiClientBase } from './baseClass';
+
+/**
+ * The Link Manager allows you to navigate to another route. Use it instead of an internal router to:
+  - Route inside micro front-ends.
```
bszwarc:
```suggestion
  - Provide routing inside micro front-ends.
```

PR #547, opened 2019-05-23
Limit iframe privileges

### LINKED ISSUE #40: Limit privileges to microfrontends
**Description**

A parent frame is holding several nested iframes (microfrontends). Communication between components is built on top of postMessage API. In regards to postMessage API, communication is only possible between microfrontends and parent frame. Custom routing logic is implemented to provide further capabilities as required. 

IdTokens are being transferred to microfrontends in order for them to authenticate against backend services when required. 

**Risk** 

Malicious code in microfrontends could run with unnecesarily high privileges.

**Suggested Mitigation**

Evaluate the possibility of making microfrontend iframes sandboxed and define a restrictive by default policy allowing only strictly required privileges. 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe

maxmarkus:
Fixes #40


maxmarkus:
TODO: Verify against kyma console.

maxmarkus:
> allow-scripts and allow-same-origin

Unfortunately our micro frontends fail once I remove one of the two tags.

maxmarkus:
Testing by Daria: Well, I was clicking around console with this PR and it seems to work fine :smile:

PR #556, opened 2019-05-27
NPM release pipeline

maxmarkus:
Publishing tested similar to our real package with the following package (will be unpublished after acceptance)
npm package: https://www.npmjs.com/package/markus-one-core
got released by: https://travis-ci.org/maxmarkus/luigi/jobs/537816105

Read issue description on how it is implemented and can be tested.


marynaKhromova:
npm publish job should run only after the other jobs have passed successfully.

scripts/npmPublish.sh
```diff
@@ -0,0 +1,100 @@
...
+}
+
+function checkGithubRelease {
+  cd $BASE_DIR/../client
+  NAME=$(node -p "require('./package.json').name")
+  CLIENT_VERSION=$(node -p "require('./package.json').version")
+  cd $BASE_DIR/../core/public
+  CORE_VERSION=$(node -p "require('./package.json').version")
+
+  if [ "CORE_VERSION" != "CLIENT_VERSION" ]; then
```
jesusreal: Corrrect me if I´m wrong, but here we are comparing strings instead of the values from the variables

scripts/npmPublish.sh
```diff
@@ -0,0 +1,100 @@
...
+
+function removeNpmToken {
+  if [ "$TRAVIS" = "true" ]; then
+    # setup token when running in travis
+    echo "" > ~/.npmrc
+  fi
+}
+
+
+checkGithubRelease
```
jesusreal: Maybe we can give this function a more generic name like `prepublishChecks`

PR #559, opened 2019-05-29
Modernize Luigi Client - Refactor event listeners

maxmarkus:
Review after https://github.com/SAP/luigi/pull/546


client/src/helpers.js
```diff
@@ -0,0 +1,62 @@
...
+class Helpers {
+  /** @private */
+  constructor() {
+    this.listeners = [];
+
+    window.addEventListener(
+      'message',
+      function(evt) {
+        this.listeners
+          .filter(listener => listener.name == evt.data.msg)
```
marynaKhromova: I think it would be better to use here `===` instead `==`.

client/src/helpers.js
```diff
@@ -0,0 +1,62 @@
...
+      }.bind(this)
+    );
+  }
+
+  /**
+   * Registers a post message listener
+   * Don't forget to remove the event listener at the end of
+   * your eventFn if you do not need it anymore.
+   * @private
+   * @param name string event name
```
marynaKhromova: as I remember `@param` should be written as a type like `{string}`, `{function}` etc.

maxmarkus: Since it is a private function it would not cause any issues, but I agree it should be same as exposed docu.

PR #564, opened 2019-06-03
Alert unit tests

maxmarkus:
Moved processTextAndLinks to helpers, refactored escaping helpers to class and added unit test.


core/test/utilities/helpers/escaping-helpers.spec.js
```diff
@@ -21,4 +29,65 @@ describe('Iframe-helpers', () => {
...
+        url: `javascript:alert('Wufff1!')`
+      },
+      works: {
+        text: 'Linktext',
+        url: `javascript:alert('Wufff2!')`
+      }
+    };
+    const uniqueID = '1234567890';
+
+    sinon.stub(EscapingHelpers, 'sanitizeHtml').callsFake(input => {
```
jesusreal: We need to mock `sanitizeParam` and `escapeRegExp` as well

core/test/utilities/helpers/escaping-helpers.spec.js
```diff
@@ -1,13 +1,21 @@
 const chai = require('chai');
 const assert = chai.assert;
 const sinon = require('sinon');
-const EscapingHelpers = require('../../../src/utilities/helpers/escaping-helpers');
+const EscapingHelpers = require('../../../src/utilities/helpers/escaping-helpers')
```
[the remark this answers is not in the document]
maxmarkus: Indeed it works, I used to remember that it never worked with es6 import. Maybe because of those older node versions, lets see if it breaks some time again.

core/test/utilities/helpers/escaping-helpers.spec.js
```diff
@@ -1,13 +1,21 @@
...
+  .EscapingHelpers;
 
 describe('Iframe-helpers', () => {
   it('sanitizeHtml', () => {
```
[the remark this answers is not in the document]
maxmarkus: I don't like that `<br>` approach at all, since you could also write it as `<br />` and it wont be covered then.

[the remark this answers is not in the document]
jesusreal: True, please ask Philipp if we need a small follow-up for that

core/test/utilities/helpers/escaping-helpers.spec.js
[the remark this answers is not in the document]
jesusreal: Please remember to change this one ;)

core/test/utilities/helpers/escaping-helpers.spec.js
```diff
@@ -16,9 +24,90 @@ describe('Iframe-helpers', () => {
...
-    const escapedRegexp = EscapingHelpers.escapeRegExp(regex);
-    assert.equal(escapedRegexp, '\\/\\*\\/');
+  it('escapeKeyForRegexp', () => {
+    const key = 'some/*/thing';
+    const escapedRegexp = EscapingHelpers.escapeKeyForRegexp(key);
+    assert.equal(escapedRegexp, 'some\\/\\*\\/thing');
+  });
+
+  describe('processTextAndLinks', () => {
+    it('without links', () => {
```
jesusreal: please also create stubs for external function in this testcase, maybe you can add them to a beforeEach block :)

core/test/utilities/helpers/escaping-helpers.spec.js
```diff
@@ -16,9 +24,90 @@ describe('Iframe-helpers', () => {
...
+      const text = `This is text <img src="http://url.to.file.which/not.exist" onerror=alert(document.cookie);><IMG SRC=j&#X41vascript:alert('test2')>`;
+      const uniqueID = '1234567890';
```
jesusreal: Looks like this test does not need `uniqueID`, if so please remove it

core/test/utilities/helpers/escaping-helpers.spec.js
jesusreal: And maybe it is a good idea to have an additional testcase to check that the stub functions are called with proper values

maxmarkus: done

PR #570, opened 2019-06-04
Move prepush hooks to travis pipeline

maxmarkus:
- pre-push hooks removed
- travis validates, generates and checks for documentation changes if there is a **docu** in package.json scripts
- else travis validates documentation only (applies to core currently)


scripts/docuCheck.sh
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bash
+
+set -e # exit on errors
+
+BASE_DIR="$( cd "$(dirname "$0")" ; pwd -P )"
+
+echoe() {
```
jesusreal: Maybe is a good idea to have this function in a separate file to avoid duplication, as we are using it already in a couple of scripts and we will probably use it more in the future.

scripts/docuCheck.sh
```diff
@@ -0,0 +1,68 @@
...
+# add all folders that are containing documentation steps
+while read LINE; do
+    checkDocu "${LINE}"
+done <<HERE
+  client
+  core
+HERE
+}
+
+installPrerequisites
```
jesusreal: Looking at the logs, would be great to add _echo_ commands signalling when each step begins

scripts/docuCheck.sh
```diff
@@ -0,0 +1,68 @@
...
+validateMdChanges() {
+  # verify that there are no changes in md files
+  MD_FILE_CHANGES=`git status | grep '.md' | wc -l`
+  if [[ $MD_FILE_CHANGES != *"0"* ]]; then
+    echoe "Unsaved documentation changes found"
+    echo "The following files need to be commited and pushed again:"
+    git status | grep '.md'
+    exit 1
+  fi
+}
```
jesusreal: An _echo_ command here reporting that the validation was successful would be great ;)

.travis.yml
```diff
@@ -8,6 +8,11 @@ notifications:
     secure: wiOchVgoiKNUikjCeIhsuv3Y5jolFud6ercdZdO3+JigQqkmoAC95f2jKTYEdplJWyJr8JY6z0+UJNceHyr5YJx/l1wpm6Zq+3H4dq/V0yUXH1uhw2eaTzweX2vjfw0bzewEdc1CGl4kSfdyTajnbKgt+mtnrAV9lPqiXYSOOJTu5BCFomi7u2GhAz1JjBK6P4Ar2jXm+oc81nYjj40P8bZFlA7Rjm0hNN42MMkKvnD8OHztL3EwezKHXlQW7fN4eOrhPOMbT3NRU72g9Nir+lfoKJlch1zPoHXQ7DRp41uGyHV/qfvfRLzXwxftZK6kdvzMe0eI0i1aIm0R6AE0bLphB2o/klJzyQpqSiQgETTH0qJN+3px2kddGrw7Me+UNC/1zZrt1MJfWf0h6WjXTsPDJf3ajLsn8OoIeBTRZbAb9as3UWQZcknuuMf8oGUzZkrZNMWYqo3py0+qWm4wSbWXWVUCgVAIYA6oEADwq37z59HySGkyHI4gEPzxsYODfCuvO4pJX1h4vEyH7w2IigwhDAPq5G4Vdez9ZFS96p/LSCF+iH8yWMy294u8wk5ofWUutFFv0JC45LQBXu51a2TeQkalwVn4DiuxxHE6jZieREw75YFjra4jpWsyziWkzshGT75NCra6cKAbNa4T8n6Pm3ogjEPeeWAO3psgSdo=
 jobs:
   include:
+    #----- DOCU -----
```
jesusreal: Maybe is a good idea to add caching so that _documentation_ package is not installed every time we do a new push to a branch. What do you think?

scripts/docuCheck.sh
maxmarkus: done

scripts/docuCheck.sh
maxmarkus: done

scripts/docuCheck.sh
maxmarkus: done

.travis.yml
maxmarkus: done

scripts/docuCheck.sh
```diff
@@ -0,0 +1,76 @@
...
+
+  cd $BASE_DIR/tmp-docu
+  DOCU_VERSION=`cat $BASE_DIR/../client/package.json | jq --raw-output ".devDependencies.documentation"`
+  # create local package json to make it cachable
+  echo "{\"dependencies\": {\"documentation\": \"$DOCU_VERSION\"}}" > package.json
+  npm i
+
+  # link documentation binary if it does not exist
+  NODE_BIN_FOLDER="node_modules/.bin/"
+  for FOLDER in "${LUIGI_FOLDERS[@]}"
```
jesusreal: Here there is a slash missing in _$FOLDER$NODE_BIN_FOLDER_ which creates additional folders and prevents caching for working properly. I would additionally propose to remove some duplication in the paths usage. Here is my proposal (you can replace it with the current implementation and see the actual changes ;) ):
``` 
  # link documentation binary if it does not exist
  for FOLDER in "${LUIGI_FOLDERS[@]}"
  do
    local NODE_BIN_FOLDER=$BASE_DIR/../$FOLDER/node_modules/.bin/
    if [ ! -e $NODE_BIN_FOLDER/documentation ]; then
      echo "Linking documentation binary for $FOLDER"
      mkdir -p $NODE_BIN_FOLDER
      ln -s $BASE_DIR/tmp-docu/$NODE_BIN_FOLDER/documentation $NODE_BIN_FOLDER/documentation
    fi
  done
```

scripts/docuCheck.sh
```diff
@@ -0,0 +1,76 @@
...
+BASE_DIR="$( cd "$(dirname "$0")" ; pwd -P )"
+
+declare -a LUIGI_FOLDERS=(
+  "client"
+  "core"
+)
+
+source $BASE_DIR/shared/bashHelpers.sh
+
+installPrerequisites() {
```
jesusreal: I would decoupled completely Luigi and Core documentation in this step. Fo that I would run the complete logic of this function once for Client and once for Core, to prevent difficult to find bugs in the future due to coupling created among Luigi and Core documentation. What do you think?

Issue #216, opened 2018-11-14
Improve default child mechanism 


dariadomagala:
**Actual behaviour**
If the user will go to the node with no view he is redirected to the default child node. It's ok until the user wants to go back using the browsers 'go back' mechanism. In that case, user goes back to the node with no view and is automatically redirected to the default child node again. 

**Expected behaviour**
When user uses the browsers 'go back' mechanism, he is always redirected to the previously seen node (so in this case one before the node with no view). 

**Suggestion**
If it's possible, change the view and the url but without the actual navigation.

PR #595, opened 2019-06-19
Add goBack support for normal navigation

maxmarkus:
goBack works now also if you are navigating normally and want to go back to the last route.
- not working when default child routing kicks in, until #216 is fixed
- goBackContext is not supported when using normal navigation, works only with preserved view


pekura:
#### Implementation details

- View group preloading is triggered after navigation has changed and the view is rendered (after the microfrontend view sends confirmation to luigi core that the navigation was successful ('luigi.navigation.ok' event) or context was requested ('luigi.get-context' event). This won't affect user experience when clicking on navigation node and it is likely that after that for some time there will be no further interaction with the UI.
- Only one view group gets preloaded the first time. The subsequent preloading events will decide the batch size (how many view groups get preloaded at once in parallel) based on measured loading time of previously preloaded view groups (if this information is available). The batch size can be 1 to 3. It gets reevaluated each time. 
- If at the time view groups preloading should be triggered there are view groups being loaded the preloading gets skipped completely till next navigation event happens. These throttling measures are there to minimize impact on user perceived responsiveness of the application.

#### How to test
You need at least on view group configured in the Luigi configuration, better more that one. In navigation section of Luigi config you need to specify it like this:

```
  viewGroupSettings = {
    ananas: {
      preloadUrl: '/ananasapp.html#/preload',
    },
    tets: {
      preloadUrl: '/tetsapp.html#/preload',
    },
    tets2: {
      preloadUrl: '/tets2app.html#/preload',
    }
}
```

Preload URL should be rather empty page that includes LuigiClient so that it can react on 'luigi.navigat.ok' event properly!

You also need to mark some navigation nodes to belong to a view group like this:

```
   {
     category: { label: 'Settings', icon: 'action-settings' },
     pathSegment: 'settings',
     label: 'Project Settings',
     viewUrl: '/sampleapp.html#/projects/' + projectId + '/settings',
    icon: 'settings',
    viewGroup: 'ananas'
   },

```

Than you can click around in the application and watch iframes with preloaded viewgroups getting preloaded (there are also debug messages to be seen in the console).

client/luigi-client.d.ts
```diff
@@ -136,8 +136,8 @@ export declare interface LinkManager {
   fromContext: (navigationContext: string) => this;
 
   /**
-   * Discards the active view and navigates back to the last visited view (preserved view), if a preserved view was set before.
-   * @param {any} goBackValue data that is passed in the `goBackContext` field to the last visited view
+   * Discards the active view and navigates back to the last visited view. Works with preserved views, and also acts as the substitute of the browser **back** button. **goBackContext** is only available when using preserved views.
+   * @param {any} goBackValue data that is passed in the `goBackContext` field to the last visited view, when using preserved views.
```
bszwarc:
```suggestion
   * @param {any} goBackValue data that is passed in the **goBackContext** field to the last visited view when using preserved views.
```

PR #618, opened 2019-06-27
Dropdown functionality for top right nav nodes

JohannesDoberer:


core/src/navigation/TopNav.html
```diff
@@ -19,8 +19,53 @@
...
+      {#each children as node ,i}
       {#if !node.hideFromNav}
+      {#if node.isCat}
+      <div
+        class="fd-shellbar__action fd-shellbar__action--hide fd-shellbar__action--collapsible"
+      >
+        <div class="fd-product-menu">
+          <div class="fd-popover fd-popover--right">
+            <div class="fd-popover__control" on:click="event.stopPropagation()">
+              <button
```
maxmarkus: The Button has some weird padding on the right, guess this is also the reason that the popover is positioned to far right.
screen markus e 2019-07-01 at 12 42 50

docs/navigation-configuration.md
```diff
@@ -27,6 +27,9 @@ A navigation path is any existing path in the navigation tree. It connects the f
 
 If you set the **hideSideNav** property to `true`, the left navigation disappears when you click the affected node. It is set to `false` by default.
 
+If you want to group some navigation nodes and seperate these nodes into one parent node you can use the **category** property. The grouped navigation nodes are rendered in a dropdown. The **category** property should have a **label** and an **icon**.
```
```suggestion
If you want to group some navigation nodes into a separate parent node, you can use the **category** property. The grouped navigation nodes are rendered in a dropdown. The **category** property needs a **label** and an **icon**.
```

PR #623, opened 2019-07-01
Node category documentation

maxmarkus:
Added missing docu for node category configuration documentation.


docs/navigation-parameters-reference.md
```diff
@@ -27,7 +27,12 @@ Luigi.setConfig({
         children: [node, node, node],
         hideFromNav: false,
         isolateView: false,
-        icon: 'settings'
+        icon: 'settings',
+        category: {
+          label: 'General',
+          icon: 'general'
+        },
```
JohannesDoberer: Maybe a comment with an 'or'?
```
category: {
  label: 'General',
  icon: 'general'
},//OR 
category: 'General'
```

maxmarkus: Yea, also thought about this, but we didn't have such case until now. Will add one

maxmarkus: done

docs/navigation-parameters-reference.md
```diff
@@ -116,6 +121,9 @@ The node parameters are as follows:
...
 - **icon** is the name of an icon from the [OpenUI](https://openui5.hana.ondemand.com/1.40.10/iconExplorer.html) or a custom link (relative or absolute) to an image displayed next to the Node label in the side navigation or instead of the label in the top navigation.
 - **hideSideNav** if set to `true`, the left navigation disappears when you click the affected node. It is set to `false` by default.
+- **category** defines a group of views that is separated with a headline and icon. At least one of the nodes of a group should be defined as an Object with `label` and `icon` properties. For all other nodes it is valid to set **category** as string with the `label` value. 
+   - **label** is a string that represents the title of the category
+   - **icon** is the name of an icon from the [OpenUI](https://openui5.hana.ondemand.com/1.40.10/iconExplorer.html) or a custom link (relative or absolute) to an image displayed next to the Node label in the side navigation or instead of the label in the top navigation.
```
jesusreal: Maybe we could explain that for OpenUI icons we already add `sap-icon--` so that the user only needs to specify the remaining part of the icon name.

jesusreal: Not sure if it is needed, but we could also say that categories can be used on both left and top navigation to group nodes

maxmarkus: good points, added those infos. guess ppl assume automatically it works on top and left nav, so maybe it was more a missing implementation so far

docs/navigation-parameters-reference.md
```diff
@@ -114,8 +119,11 @@ The node parameters are as follows:
...
 - **viewGroup** defines a group of views in the same domain sharing a common security context. This improves performance through reusing the frame. Use **viewGroup** only for the views that use path routing internally.
-- **icon** is the name of an icon from the [OpenUI](https://openui5.hana.ondemand.com/1.40.10/iconExplorer.html) or a custom link (relative or absolute) to an image displayed next to the Node label in the side navigation or instead of the label in the top navigation.
+- **icon** is the name of an icon, without the `sap-icon--` prefix, from the [OpenUI](https://openui5.hana.ondemand.com/1.40.10/iconExplorer.html) or a custom link (relative or absolute) to an image displayed next to the Node label in the side navigation or instead of the label in the top navigation.
```
```suggestion
- **icon** is the name of an icon, without the `sap-icon--` prefix. Its source may be [OpenUI](https://openui5.hana.ondemand.com/1.40.10/iconExplorer.html) or a custom link (relative or absolute) to an image. The icon is displayed next to the node label in the side navigation or instead of the label in the top navigation.

File: docs/navigation-parameters-reference.md [hunk 4]
```diff
@@ -114,8 +119,11 @@ The node parameters are as follows:
...
 - **hideSideNav** if set to `true`, the left navigation disappears when you click the affected node. It is set to `false` by default.
+- **category** defines a group of views that is separated with a headline and icon. At least one of the nodes of a group should be defined as an Object with `label` and `icon` properties. For all other nodes it is valid to set **category** as string with the `label` value.
```
```suggestion
- **category** defines a group of views separated with a headline and an icon. You should define at least one node in a group should as an Object with **label** and **icon** properties. For all other nodes, you can set **category** as a string with the `label` value. 
```

docs/navigation-parameters-reference.md
```diff
@@ -114,8 +119,11 @@ The node parameters are as follows:
...
+- **icon** is the name of an icon, without the `sap-icon--` prefix. Its source may be [OpenUI](https://openui5.hana.ondemand.com/1.40.10/iconExplorer.html) or a custom link (relative or absolute) to an image. The icon is displayed next to the node label in the side navigation or instead of the label in the top navigation.
 - **hideSideNav** if set to `true`, the left navigation disappears when you click the affected node. It is set to `false` by default.
+- **category** defines a group of views separated with a headline and an icon. You should define at least one node in a group should as an Object with **label** and **icon** properties. For all other nodes, you can set **category** as a string with the `label` value. 
+   - **label** is a string that represents the title of the category
+   - **icon** is the name of an icon, without the `sap-icon--` prefix, from the [OpenUI](https://openui5.hana.ondemand.com/1.40.10/iconExplorer.html) or a custom link (relative or absolute) to an image displayed next to the label in the side navigation or instead of the label in the top navigation. In case you accidentially define different icons in a category group, only the first one is used.
```
```suggestion
   -  **icon** is the name of an icon, without the `sap-icon--` prefix. Its source may be [OpenUI](https://openui5.hana.ondemand.com/1.40.10/iconExplorer.html) or a custom link (relative or absolute) to an image. The icon is displayed next to the node label in the side navigation or instead of the label in the top navigation. In case you accidentally define different icons in a category group, only the first one is used.
```

docs/navigation-parameters-reference.md
```diff
@@ -114,8 +119,12 @@ The node parameters are as follows:
...
 - **hideSideNav** if set to `true`, the left navigation disappears when you click the affected node. It is set to `false` by default.
+- **category** defines a group of views separated with a headline and an icon. You should define at least one node in a group should as an Object with **label** and **icon** properties. For all other nodes, you can set **category** as a string with the `label` value. 
+   - **label** is a string that represents the title of the category
+   -  **icon** is the name of an icon, without the `sap-icon--` prefix. Its source may be [OpenUI](https://openui5.hana.ondemand.com/1.40.10/iconExplorer.html) or a custom link (relative or absolute) to an image. The icon is displayed next to the node label in the side navigation or instead of the label in the top navigation. In case you accidentally define different icons in a category group, only the first one is used.
+   - **collapsible** if set to `true`, category items are hidden initially and can be expanded on click. Multiple collapsible categories look like an accordion.
```
```suggestion
   - **collapsible** if set to `true`, category items are hidden at first. To expand them, click the main category node.
```

PR #626, opened 2019-07-03
Open modal from navigation node

JohannesDoberer:


core/examples/luigi-sample-angular/e2e/tests/navigation.spec.js
```diff
@@ -199,5 +199,43 @@ describe('Navigation', () => {
...
+      cy.get('iframe', { timeout: 1000 }).then(ifr => {
+        iframeBody = ifr.contents().find('body');
+      });
+      // projects page
+      cy.get('.fd-shellbar')
+        .contains('Projects')
+        .click();
+
+      //projects page
+      cy.get('.fd-app__sidebar').should('contain', 'Project One');
```
jesusreal: I would say these two checks for project one and two are not necessary, as well as this one below:
``` 
      cy.get('.fd-app__sidebar').should('not.contain', 'Project Two');
```

core/examples/luigi-sample-angular/e2e/tests/navigation.spec.js
```diff
@@ -199,5 +199,43 @@ describe('Navigation', () => {
       cy.get('.no-side-nav').should('exist');
       cy.get('.fd-app__sidebar').should('not.be.visible');
     });
+
+    it('Open navigation node in a modal', () => {
+      let iframeBody;
+      cy.get('iframe', { timeout: 1000 }).then(ifr => {
```
jesusreal: Looks like these timeout is not needed (run the test a couple of times locally)

core/src/App.html
```diff
@@ -577,7 +577,21 @@
       handleNavClick(node) {
         this.getUnsavedChangesModalPromise().then(() => {
           this.closeLeftNav();
-          Routing.handleRouteClick(node, this);
+          if (node.openNodeInModal) {
+            const route = RoutingHelpers.buildRoute(
```
jesusreal: wouldn't it be better to call here `Routing.getNodePath`?

core/src/App.html
```diff
@@ -577,7 +577,21 @@
...
+              node,
+              `/${node.pathSegment}`
+            );
+            if (node.openNodeInModal)
+              this.openViewInModal(
+                route,
+                typeof node.openNodeInModal === 'boolean'
```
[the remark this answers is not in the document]
JohannesDoberer: Yes because it is cool. Also Pete liked it.. and it is documented. See docu..
`Optionally you can set the **openNodeInModal** property to true to use default title and size.`

core/src/App.html
```diff
@@ -577,7 +577,21 @@
       handleNavClick(node) {
         this.getUnsavedChangesModalPromise().then(() => {
           this.closeLeftNav();
-          Routing.handleRouteClick(node, this);
+          if (node.openNodeInModal) {
+            const route = RoutingHelpers.buildRoute(
+              node,
+              `/${node.pathSegment}`
+            );
+            if (node.openNodeInModal)
```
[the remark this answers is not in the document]
JohannesDoberer: good point! thx

core/src/App.html
JohannesDoberer: Mmmh... take a look in the Routing.handleRouteClick function, which will be called in the else part you will find the same call. So there is nothing changed in the 'flow'. That's why i took this way.

core/examples/luigi-sample-angular/e2e/tests/navigation.spec.js
JohannesDoberer: fine for me :)

core/examples/luigi-sample-angular/e2e/tests/navigation.spec.js
JohannesDoberer: ok

core/src/App.html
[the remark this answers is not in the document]
jesusreal: Ok, I missed that part in the docu

core/src/App.html
jesusreal: Makes sense, thanks for the clarification

docs/navigation-parameters-reference.md
```diff
@@ -116,6 +116,9 @@ The node parameters are as follows:
 - **viewGroup** defines a group of views in the same domain sharing a common security context. This improves performance through reusing the frame. Use **viewGroup** only for the views that use path routing internally.
 - **icon** is the name of an icon from the [OpenUI](https://openui5.hana.ondemand.com/1.40.10/iconExplorer.html) or a custom link (relative or absolute) to an image displayed next to the Node label in the side navigation or instead of the label in the top navigation.
 - **hideSideNav** if set to `true`, the left navigation disappears when you click the affected node. It is set to `false` by default.
+- **openNodeInModal** is an object which configures the view to be opened in a modal. You can specify the modal's title and size. If you don't specify the title, it is the node label. If there is no node label, the title remains empty. The default size of the modal is `l`, which means 80%. You can also use `m` (60%) and `s` (40%) to set the modal size.<br/> Optionally you can set the **openNodeInModal** property to `true` to use default title and size.
```
```suggestion
- **openNodeInModal** configures the settings of the view which opens in a modal. You can specify the modal's title and size. If you don't specify the title, the node label is used. If there is no node label, the title remains empty. The default size of the modal is `l`, which means 80%. You can also use `m` (60%) and `s` (40%) to set the modal size.<br/> Optionally you can set the **openNodeInModal** property to `true` to use default title and size.
```

PR #627, opened 2019-07-04
Document view groups

jesusreal:
**Description**

- Add documentation and parameters reference for viewgroups, including caching and preloading


docs/navigation-configuration.md
```diff
@@ -268,3 +268,15 @@ Luigi.setConfig({
     ]
   }
 });
+```
+
+## View Groups
+You can use the view groups feature to override the default iframes management policy. Imagine you have in your app the following microfrontend views: `http://mysite.com/a#e` and  `http://mysite.com/b#f`. Due to hash routing they don´t belong to the same domain, since the domain is the path until the `#` and therefore they will be rendered in different iframes by default. Nevertheless both views might have the **same origin** (in the example `mysite.com`) and belong to the **same microfrontend** and you will want to render them in the same iframe. The view groups feature makes that possible just by setting the `viewGroup` parameter in the most upper nodes, as their children nodes are automatically considered part of the same view group.
```
pekura: The term 'domain' usually refers to the host name only (mysite.com in this case). Luigi's default policy to determine if to render the previous and next view in the same iframe compares the whole url with protocol, domain name, port and even the path till # character - that is true, but I would not refer to it as 'domain', this could be confusing.

docs/navigation-parameters-reference.md
```diff
@@ -16,29 +16,72 @@ Luigi.setConfig({
...
-        viewUrl: 'https://admin.mydomain.com/settings',
-        viewGroup: 'settingsGroup',
-        // optional
-        children: [node, node, node],
-        hideFromNav: false,
-        isolateView: false,
-        icon: 'settings'
+  viewGroupSettings: {
+    main: {
+      preloadUrl: 'https://my-site.com/index.html#/overview',
```
pekura: The example preloadUrls should not lead to the impression that you should use any of your usual views. The preload URL view should be small, visually empty but must have LuigiClient included and must be fine with getting an empty context (e.g. without a token). Maybe this should be mentioned below more clearly.

docs/navigation-configuration.md
jesusreal: thanks, done

docs/navigation-parameters-reference.md
jesusreal: You are right, I added that info

docs/navigation-configuration.md
```diff
@@ -268,3 +268,15 @@ Luigi.setConfig({
     ]
   }
 });
+```
+
+## View Groups
+You can use the view groups feature to override the default iframes management policy. Imagine you have in your app the following microfrontend views: `http://mysite.com/a#e` and  `http://mysite.com/b#f`. Due to hash routing and the fact the path until _#_ differs, they will be rendered in different iframes by default. Nevertheless both views might have the **same origin** (in the example `mysite.com`) and belong to the **same microfrontend** and you will want to render them in the same iframe. The view groups feature makes that possible just by setting the `viewGroup` parameter in the most upper nodes, as their children nodes are automatically considered part of the same view group. 
+
+View group nodes are always rendered in their own iframe. Nodes not belonging to any view group will default to the same-origin iframe rendering policy.
```
maxmarkus: in their own iframe .. means in a view group iframe?

docs/navigation-configuration.md
```diff
@@ -268,3 +268,15 @@ Luigi.setConfig({
...
+
+The view groups feature also offers caching. Caching of view groups is provided out of the box. Everytime you navigate to another view group, either a new iframe is created or if the iframe already exists, it is reused. In both cases, the iframe you are navigating from is not destroyed, but hidden and available to be used again. If you navigate back to the first iframe, and the view in that frame should be updated (e.g. in the second iframe you just added a new entry to a table that should be displayed in the first iframe), you need to set a preload url to any view from the view group to ensure that the view is refreshed when you navigate back to it. 
+
+We offer also preloading for view groups. You just need to setup which url you want to preload, and Luigi will preload the view after some user interactions when the browser is most likely to be idle. This option is active by default, but can be deactivated via a [configuration flag](navigation-parameters-reference.md#node-parameters).
+
+For more information about how to set caching with view refresh and preloading for view groups, please read [here](navigation-parameters-reference.md#node-parameters).
```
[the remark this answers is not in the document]
jesusreal: done via regular commit

docs/navigation-configuration.md
```diff
@@ -268,3 +268,15 @@ Luigi.setConfig({
...
+
+Nodes belonging to the same view group are always rendered in their own viewgroup iframe. Nodes not belonging to any view group will default to the same-origin iframe rendering policy. 
+
+The view groups feature also offers caching. Caching of view groups is provided out of the box. Everytime you navigate to another view group, either a new iframe is created or if the iframe already exists, it is reused. In both cases, the iframe you are navigating from is not destroyed, but hidden and available to be used again. If you navigate back to the first iframe, and the view in that frame should be updated (e.g. in the second iframe you just added a new entry to a table that should be displayed in the first iframe), you need to set a preload url to any view from the view group to ensure that the view is refreshed when you navigate back to it.
```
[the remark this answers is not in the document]
jesusreal: That is great but I didn't apply it completely because the example part was confusing for me. Please check again.

docs/navigation-configuration.md
```diff
@@ -268,3 +268,15 @@ Luigi.setConfig({
     ]
   }
 });
+```
+
+## View Groups
+You can use the view groups feature to override the default iframes management policy. Imagine you have in your app the following microfrontend views: `http://mysite.com/a#e` and  `http://mysite.com/b#f`. Due to hash routing and the fact the path until _#_ differs, they will be rendered in different iframes by default. Nevertheless both views might have the **same origin** (in the example `mysite.com`) and belong to the **same microfrontend** and you will want to render them in the same iframe. The view groups feature makes that possible just by setting the `viewGroup` parameter in the most upper nodes, as their children nodes are automatically considered part of the same view group.
```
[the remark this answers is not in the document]
jesusreal: I commited your suggestion, just replaced "micro front-end" with "micro frontend".

PR #639, opened 2019-07-09
alias fn for getEventData

JohannesDoberer:


client/luigi-client.d.ts
```diff
@@ -259,6 +259,14 @@ export type getToken = () => AuthData['accessToken'];
 export function getEventData(): Context;
 export type getEventData = () => Context;
 
+/**
+ * Returns the context object. It invokes the getEventData() function.
```
jesusreal: Maybe it is more elegant to say here and in the docu that it is an alias of the getEventData function instead of saying that it invokes it?

docs/luigi-client-api.md
```diff
@@ -50,6 +50,12 @@ Returns the context object. Typically it is not required as the [addContextUpdat
 
 Returns **[Object](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object)** current context data
 
+### getContext
+
+Returns the context object. It is an alias function for getEventData().
```
bszwarc:
```suggestion
Returns the context object. It is an alias function for `getEventData()`.
```

PR #648, opened 2019-07-11
Generic node actions

jesusreal:
**Description**

- Add generic node action
- Write documentation, unit and e2e tests for new feature
- Refactor unit tests in [this commit](https://github.com/SAP/luigi/commit/3b158bb2b6b127ea68602337be4d937dfb289d80)
- For e2e, refactor how _iframeBody_ is fetched in [this commit](https://github.com/SAP/luigi/commit/d57dc5644d6ba5400e62939e2669bdd515a154a0)


docs/navigation-parameters-reference.md
```diff
@@ -190,6 +190,7 @@ The node parameters are as follows:
 - **openNodeInModal** configures the settings of the view which opens in a modal. You can specify the modal's title and size. If you don't specify the title, the node label is used. If there is no node label, the title remains empty. The default size of the modal is `l`, which means 80%. You can also use `m` (60%) and `s` (40%) to set the modal size.<br/> Optionally you can set the **openNodeInModal** property to `true` to use default title and size.
   -   **title** modal title. By default, it is the node label. If there is no label, it is left empty
   -   **size** **(`"l"` \| `"m"` \| `"s"`)** size of the modal (optional, default `"l"`)
+- **onNodeActivation** is an optional method that is executed when a request to navigate to the node occurs. The method receives as input parameter the node object as described in the configuration. This method can return synchronously or asynchronously. If the method returns boolean `false`, the navigation is not triggered, otherwise navigation happens as usual.
```
pekura: We usually use the term 'function' instead of method throughout the documentation. For the sake of consistency we should call this one function as well.

core/examples/luigi-sample-angular/e2e/tests/navigation.spec.js
```diff
@@ -88,6 +86,53 @@ describe('Navigation', () => {
     });
   });
 
+  describe('Node activation hook', () => {
```
pekura: This is a cool scenario. But I think we should include clicking an actual node in the top or left navigation to activate the on-activation function as this is the main use case.

docs/navigation-parameters-reference.md
jesusreal: Good point! I replaced the four occurrences of `method` with `function`

core/examples/luigi-sample-angular/e2e/tests/navigation.spec.js
jesusreal: Really good point, I have updated the example app and the e2e tests.

core/examples/luigi-sample-angular/src/luigi-config/extended/projectDetailNav.js
```diff
@@ -64,6 +64,39 @@ export const projectDetailNavStructure = projectId => [
     viewUrl: '/sampleapp.html#/projects/' + projectId + '/developers',
     icon: '/assets/favicon-sap.ico'
   },
+  {
```
pekura: I like your choice of icons :-).

docs/navigation-parameters-reference.md
```diff
@@ -194,6 +194,7 @@ The node parameters are as follows:
 - **openNodeInModal** configures the settings of the view which opens in a modal. You can specify the modal's title and size. If you don't specify the title, the node label is used. If there is no node label, the title remains empty. The default size of the modal is `l`, which means 80%. You can also use `m` (60%) and `s` (40%) to set the modal size.<br/> Optionally you can set the **openNodeInModal** property to `true` to use default title and size.
   -   **title** modal title. By default, it is the node label. If there is no label, it is left empty
   -   **size** **(`"l"` \| `"m"` \| `"s"`)** size of the modal (optional, default `"l"`)
+- **onNodeActivation** is an optional function that is executed when a request to navigate to the node occurs. The function receives as input parameter the node object as described in the configuration. This function can return synchronously or asynchronously. If the function returns boolean `false`, the navigation is not triggered, otherwise navigation happens as usual.
```
```suggestion
- **onNodeActivation** is an optional function executed when a request to navigate to the node occurs. As an input parameter, the function receives the node object as described in the configuration. This function can return results synchronously or asynchronously. If the function returns boolean `false`, the navigation is not triggered, otherwise, navigation renders as usual.
```

core/examples/luigi-sample-angular/src/luigi-config/extended/projectDetailNav.js
jesusreal: Thanks :)

PR #650, opened 2019-07-12
Split View microfrontends

maxmarkus:
Added ability to open a view in a split screen.


client/src/linkManager.js
```diff
@@ -75,6 +94,110 @@ export class linkManager extends LuigiClientBase {
...
+   * @param {Object} splitViewSettings opens a view in a split view. Use these settings to configure the split view's behaviour
+   * @param {string} splitViewSettings.title split view title. By default, it is the node label. If there is no label, it is left empty
+   * @param {number} [splitViewSettings.size=40] size of the split view in percent
+   * @example
+   * LuigiClient.linkManager().openAsSplitView('projects/pr1/logs', {title: 'Logs', size: 40});
+   */
+  openAsSplitView(path, splitViewSettings) {
+    this.navigate(path, 0, true, undefined, splitViewSettings || {});
+    const sendSplitViewAction = (action, data) => {
+      window.parent.postMessage({
```
jesusreal: Please note this will need to be updated once [_Post message target handling in Luigi Client_](https://github.com/SAP/luigi/pull/651) is merged

core/src/services/split-view.js
```diff
@@ -0,0 +1,228 @@
...
+  RoutingHelpers
+} from '../utilities/helpers';
+
+class SplitViewSvcClass {
+  constructor() {
+    this.splitViewValues;
+    this.storedSplitViewValues;
+  }
+
+  getSplitViewContainer() {
```
jesusreal: As we are calling all these methods with `SplitViewSvc` maybe we don't need to add 'SplitView` to the method names. What do you think?

client/src/splitViewHandle.js
```diff
@@ -0,0 +1,177 @@
...
+   * @private
+   */
+  constructor(settings) {
+    super();
+
+    this.splitView = {
+      exists: true,
+      size: 0,
+      isCollapsed: false,
+      isExpanded: true
```
jesusreal: Ad discussed, if this object is only for internal use, maybe we don't need `isExpanded`, and as `isCollapsed` is not a function we could call it just `collapsed`

docs/luigi-client-api.md
```diff
@@ -211,6 +238,141 @@ LuigiClient.linkManager().goBack({ foo: 'bar' });
...
+splitViewHandle.setSize(60);
+```
+
+### on
+
+Registers a listener for split view events
+
+#### Parameters
+
+-   `key` **enum** a set of predefined events: expand, collapse, resize, close
```
jesusreal: Correct me if I am wrong, but the key is actually a string from an enum, not an enum itself

core/src/services/iframe.js
```diff
@@ -34,11 +34,15 @@ class IframeClass {
     return container && container.length > 0 ? container[0] : undefined;
   }
 
-  getAllIframes(modalIframe) {
+  getAllIframes(iframeTypeArr = []) {
```
jesusreal: Maybe we also need to update LuigiElements.getMicrofrontendIframes(). Probably we can get rid of `getAllIframes` and use also internally Luigi Core api. But then we might need to extend the elements and provide functions like: `getAllIMicrofrontendIframes`, `getMainViewMicrofrontendIframes`, `getSplitViewMicrofrontendIframes` and `getModalMicrofrontendIframes`.

core/src/services/iframe.js
```diff
@@ -34,11 +34,15 @@ class IframeClass {
...
     const iframes = Array.from(
       document.querySelectorAll('.iframeContainer iframe')
     );
-    if (modalIframe) iframes.push(modalIframe);
+    iframeTypeArr
```
jesusreal: What about any of the following variants? 
```
return iframes.concat(iframeTypeArr.filter(Boolean));
``` 
or 
``` 
iframes.push(...iframeTypeArr.filter(Boolean));
return iframes;
```

core/src/App.html
```diff
@@ -87,7 +107,7 @@
     return paths.includes(removeQueryParams(routePath));
   };
 
-  const enableRouting = (component, node, config) => {
+  const enableRouting = (component, node, config, callback) => {
```
jesusreal: Maybe it is overkill to pass the callback and we could just execute the callback code inline?

client/src/linkManager.js
```diff
@@ -37,16 +38,27 @@ export class linkManager extends LuigiClientBase {
...
    * LuigiClient.linkManager().navigate('users/groups/stakeholders')
    * LuigiClient.linkManager().navigate('/settings', null, true) // preserve view
    */
-  navigate(path, sessionId, preserveView, modalSettings) {
+  navigate(path, sessionId, preserveView, modalSettings, splitViewSettings) {
     if (this.options.errorSkipNavigation) {
       this.options.errorSkipNavigation = false;
       return;
     }
+    if ([modalSettings, splitViewSettings].filter(obj => !!obj).length > 1) {
```
jesusreal: Maybe `if (modalSettings && splitViewSettings)` or if you want to be more strict `if (Boolean(modalSettings) && Boolean(splitViewSettings))` do the same and they are more simple ;)

client/src/linkManager.js
```diff
@@ -75,6 +88,23 @@ export class linkManager extends LuigiClientBase {
...
+   * @param {string} path navigation path
+   * @param {Object} splitViewSettings opens a view in a split view. Use these settings to configure the split view's behaviour
+   * @param {string} splitViewSettings.title split view title. By default, it is the node label. If there is no label, it is left empty
+   * @param {number} [splitViewSettings.size=40] size of the split view in percent
+   * @returns {Object} instance of SplitView
+   * @see {@link splitView} for further documentation about the returned instance
+   * @example
+   * const splitViewHandle = LuigiClient.linkManager().openAsSplitView('projects/pr1/logs', {title: 'Logs', size: 40});
+   */
+  openAsSplitView(path, splitViewSettings) {
```
jesusreal: I would vote to default `splitViewSettings` to `{}` so that we avoid any future runtime error for our customers in case of creating the split view handle with `undefined` splitViewSettings, and then we can remove the `or` in the next line ;)

docs/luigi-client-api.md
maxmarkus: done

client/src/splitViewHandle.js
maxmarkus: done

core/src/services/iframe.js
maxmarkus: Guess we do not want to expose so many features in Core Api yet.

core/src/services/iframe.js
maxmarkus: done

core/src/App.html
maxmarkus: done

client/src/linkManager.js
maxmarkus: done

client/src/linkManager.js
maxmarkus: done

client/src/linkManager.js
```diff
@@ -37,16 +38,27 @@ export class linkManager extends LuigiClientBase {
    * @param {Object} modalSettings opens a view in a modal. Use these settings to configure the modal's title and size
    * @param {string} modalSettings.title modal title. By default, it is the node label. If there is no label, it is left empty
    * @param {('l'|'m'|'s')} [modalSettings.size="l"] size of the modal
+   * @param {Object} splitViewSettings opens a view in a split view. Use these settings to configure the split view's behaviour
+   * @param {string} splitViewSettings.title split view title. By default, it is the node label. If there is no label, it is left empty
+   * @param {number} [splitViewSettings.size=40] size of the split view in percent
+   * @param {boolean} [splitViewSettings.collapsed=false] create split view but leave it closed initially
+   * @returns {Object} handle that allows functions to control the splitView: collapse, expand, setSize, onCollapse, onExpand, onResize, exists, getSize, isCollapsed
```
jesusreal: Look like this `returns` does not belong here

scripts/npmPrePublish.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
```
jesusreal: As discussed, please remove this file, since it is not related to the task ;)

client/src/linkManager.js
```diff
@@ -37,16 +38,27 @@ export class linkManager extends LuigiClientBase {
    * @param {Object} modalSettings opens a view in a modal. Use these settings to configure the modal's title and size
    * @param {string} modalSettings.title modal title. By default, it is the node label. If there is no label, it is left empty
    * @param {('l'|'m'|'s')} [modalSettings.size="l"] size of the modal
+   * @param {Object} splitViewSettings opens a view in a split view. Use these settings to configure the split view's behaviour
+   * @param {string} splitViewSettings.title split view title. By default, it is the node label. If there is no label, it is left empty
+   * @param {number} [splitViewSettings.size=40] size of the split view in percent
```
jesusreal: Maybe we want to indicate that 'size' refers to the height

core/src/services/split-view.js
```diff
@@ -0,0 +1,215 @@
...
+            displayed: false,
+            collapsed: false
+          });
+          Iframe.getIframeContainer().style.paddingBottom = '';
+          this.sendClientEvent('close');
+        });
+    }
+  }
+
+  sendClientEvent(name, data) {
```
jesusreal: shall we check if `name` has a proper value and in case not display a console warning instead of sending the message?

core/examples/luigi-sample-angular/src/app/project/project.component.ts
```diff
@@ -57,6 +59,15 @@ export class ProjectComponent implements OnInit, OnDestroy {
     if (this.cudListener) {
       const removed = removeContextUpdateListener(this.cudListener);
     }
+    this.unsubscribeSplitViewHandlers();
+  }
+
+  private unsubscribeSplitViewHandlers() {
```
jesusreal: I think it would make sense to offer this out of the box when split view close happens, then we can remove some code from this component and in the implementation, the code under `window.onunload` can be moved to the `closed.ok` listener ;)

core/src/App.html
```diff
@@ -641,7 +670,9 @@
                 }
                 resolve();
               },
-              () => {}
+              e => {
+                console.error('E', e);
```
jesusreal: In case we want to keep the `console.error`, can we say "Error in handleNavigation" or sth similar instead of "E"?

client/src/linkManager.js
maxmarkus: done

client/src/linkManager.js
maxmarkus: true, done

core/src/SplitView.html
```diff
@@ -0,0 +1,320 @@
...
+<script>
+  import { LuigiElements } from './core-api';
+  import { Iframe, SplitViewSvc } from './services';
+  import {
+    GenericHelpers,
+    IframeHelpers,
+    RoutingHelpers
+  } from './utilities/helpers';
+
+  let rightContentHeight,
```
jesusreal: Maybe it is a good idea to wrap all these variables under an object

client/luigi-client.d.ts
```diff
@@ -207,6 +230,21 @@ export declare interface LinkManager {
...
    */
   openAsModal: (nodepath: string, modalSettings?: ModalSettings) => void;
+
+  /**
+   * Opens a view in a splitted view. You can specify the split view's title and size. If you don't specify the title, it is the node label. If there is no node label, the title remains empty.  The default size of the split view is `l`, which means 80%. You can also use `m` (60%) and `s` (40%) to set the split view size. Optionally, use it in combination with any of the navigation functions.
+   * @memberof linkManager
+   * @param {string} path navigation path
+   * @param {Object} splitViewSettings opens a view in a split view. Use these settings to configure the split view's behaviour
+   * @param {string} splitViewSettings.title split view title. By default, it is the node label. If there is no label, it is left empty
+   * @param {number} [splitViewSettings.size=40] size of the split view in percent
```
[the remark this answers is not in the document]
maxmarkus: oh yea, done.

client/luigi-client.d.ts
```diff
@@ -207,6 +230,21 @@ export declare interface LinkManager {
    * LuigiClient.linkManager().openAsModal('projects/pr1/users', {title:'Users', size:'m'});
    */
   openAsModal: (nodepath: string, modalSettings?: ModalSettings) => void;
+
+  /**
+   * Opens a view in a splitted view. You can specify the split view's title and size. If you don't specify the title, it is the node label. If there is no node label, the title remains empty.  The default size of the split view is `l`, which means 80%. You can also use `m` (60%) and `s` (40%) to set the split view size. Optionally, use it in combination with any of the navigation functions.
```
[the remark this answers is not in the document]
maxmarkus: done

client/luigi-client.d.ts
```diff
@@ -207,6 +230,21 @@ export declare interface LinkManager {
...
+   * @memberof linkManager
+   * @param {string} path navigation path
+   * @param {Object} splitViewSettings opens a view in a split view. Use these settings to configure the split view's behaviour
+   * @param {string} splitViewSettings.title split view title. By default, it is the node label. If there is no label, it is left empty
+   * @param {number} [splitViewSettings.size=40] size of the split view in percent
+   * @example
+   * LuigiClient.linkManager().openAsSplitView('projects/pr1/logs', {title: 'Logs', size: 40});
+   */
+  openAsSplitView: (
+    nodepath: string,
```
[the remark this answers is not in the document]
maxmarkus: done

client/src/linkManager.js
```diff
@@ -75,6 +88,23 @@ export class linkManager extends LuigiClientBase {
     this.navigate(path, 0, true, modalSettings || {});
   }
 
+  /**
+   * Opens a view in a splitted view. You can specify the split view's title and size. If you don't specify the title, it is the node label. If there is no node label, the title remains empty.  The default size of the split view is `l`, which means 80%. You can also use `m` (60%) and `s` (40%) to set the split view size. Optionally, use it in combination with any of the navigation functions.
```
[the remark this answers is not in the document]
maxmarkus: done

client/src/linkManager.js
```diff
@@ -75,6 +88,23 @@ export class linkManager extends LuigiClientBase {
...
+   * @memberof linkManager
+   * @param {string} path navigation path
+   * @param {Object} splitViewSettings opens a view in a split view. Use these settings to configure the split view's behaviour
+   * @param {string} splitViewSettings.title split view title. By default, it is the node label. If there is no label, it is left empty
+   * @param {number} [splitViewSettings.size=40] size of the split view in percent
```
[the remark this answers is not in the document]
maxmarkus: Its an integer, description is wrong.

core/src/App.html
```diff
@@ -286,7 +311,10 @@
         document.body.classList.add('lui-semiCollapsible');
       }
     },
-    onstate({ changed, current, previous }) {},
+    onstate({ changed, current, previous }) {
+      if (changed.mfSplitView) {
```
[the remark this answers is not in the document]
maxmarkus: This happens when you clean up console.log automagically :D

core/src/App.html
```diff
@@ -741,12 +772,18 @@
...
             this.set({
               mfModal: {
-                isDisplayed: false
+                displayed: false
               }
             });
           });
         }
       },
+      open(nodepath, settings) {
```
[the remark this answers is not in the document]
maxmarkus: Jesus asked me to remove the `SplitView` from there since you always use with the service prefix, eg `SplitViewSvc.open()`. Do you still want to see the change?

core/src/utilities/helpers/generic-helpers.js
```diff
@@ -222,6 +223,18 @@ class GenericHelpersClass {
     }
     return processedString;
   }
+
+  getRightContentHeight() {
```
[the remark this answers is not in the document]
maxmarkus: I like _ getContentAreaHeight_.

client/src/splitViewHandle.js
```diff
@@ -0,0 +1,176 @@
...
+export class splitViewHandle extends LuigiClientBase {
+  /**
+   * @private
+   */
+  constructor(settings) {
+    super();
+
+    this.splitView = {
+      exists: true,
+      size: 0,
```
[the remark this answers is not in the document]
maxmarkus: done

client/src/splitViewHandle.js
```diff
@@ -0,0 +1,176 @@
...
+          this.splitView.size = e.data.data;
+        }
+      ),
+      helpers.addEventListener(
+        `luigi-client.navigation.splitview.collapse`,
+        () => {
+          this.splitView.collapsed = true;
+        }
+      ),
+      helpers.addEventListener(
```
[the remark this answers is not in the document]
maxmarkus: On close the view is already destroyed, or at least a race condition. I will give it a try again. Did it with window.onclose additionally, which does not always trigger, but in most cases.

core/examples/luigi-sample-angular/e2e/tests/split-view-microfrontend.spec.js
```diff
@@ -0,0 +1,105 @@
...
+        cy.expectPathToBe('/projects/pr2');
+
+        cy.get('button')
+          .contains('expand')
+          .scrollIntoView();
+        const buttons = ['expand', 'collapse', 'setSize', 'close'];
+        // button states, same order as buttons array
+        const tests = [
+          {
+            checkStates: [false, true, true, true],
```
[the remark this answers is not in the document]
maxmarkus: done

core/src/App.html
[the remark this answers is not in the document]
pekura: This would make sense:

```
      openSplitView(nodepath, settings) {
        SplitViewSvc.open(this, nodepath, settings);
      },
      closeSplitView() {
        SplitViewSvc.close(this);
      },
```

What do you think?

core/examples/luigi-sample-angular/e2e/tests/split-view-microfrontend.spec.js
```diff
@@ -0,0 +1,105 @@
+describe('SplitView', () => {
+  beforeEach(() => {
+    cy.visit('/');
```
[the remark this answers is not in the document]
maxmarkus: done

core/src/App.html
maxmarkus: was just a test, guess we don't need it

core/src/SplitView.html
maxmarkus: done

client/src/splitViewHandle.js
```diff
@@ -0,0 +1,169 @@
...
+    };
+
+    Object.assign(this, settings);
+
+    this.splitView.listeners = [
+      helpers.addEventListener(`luigi.navigation.splitview.internal`, e => {
+        // console.log('received new internal data', e.data.data);
+        Object.assign(this.splitView, e.data.data);
+      }),
+      helpers.addEventListener(`luigi.navigation.splitview.resize`, e => {
```
jesusreal: Correct me if I am wrong, but I think we don't need the listeners for collapse and expand, since the fields are already updated via the `luigi.navigation.splitview.internal` handler.

jesusreal: Probably items for follow up tasks, writing them here so that we don't forget:
- When collapsing the split view, the resize event is triggered.
- After resizing one, clicking (without shift) on the resize bar sends the resize event.

core/src/services/split-view.js
```diff
@@ -0,0 +1,215 @@
...
+import {
+  GenericHelpers,
+  IframeHelpers,
+  RoutingHelpers
+} from '../utilities/helpers';
+
+class SplitViewSvcClass {
+  constructor() {
+    this.splitViewValues;
+    this.storedSplitViewValues;
```
jesusreal: Probably an item for follow up task. Are we sure we need to have both `splitViewValues` and `storedSplitViewValues`? I refactored it to use only `splitViewValues` and it seems to work.

client/src/splitViewHandle.js
jesusreal: Also for the listeners we keep (except `internal`), we need to add the '.ok' suffix to the message name

client/src/splitViewHandle.js
```diff
@@ -0,0 +1,176 @@
...
+   * @example
+   * splitViewHandle.removeEventListener(listenerId);
+   */
+  removeEventListener(id) {
+    return helpers.removeEventListener(id);
+  }
+  // TODO: check if we want to provide removeListener, alterantively as last param in the callback or do not provide it
+  // removeListener: (id) => { return helpers.removeEventListener(id); },
+  /**
+   * Collapses the split view
```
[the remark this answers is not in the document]
maxmarkus: done

core/src/App.html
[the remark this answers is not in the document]
maxmarkus: Oh, its in App.html, of course it should be different there. Thanks

core/src/services/split-view.js
maxmarkus: Correct, initially thought there needs to be a "reset" .. like if you press ESC while dragging, so the old value stays. But since we do not provide such functionality, its ok to combine them.

core/src/SplitView.html
maxmarkus: done

core/examples/luigi-sample-angular/src/app/project/project.component.ts
jesusreal: Discussed, we will do it as described above, since each time a new split view is created the user gets a new instance of `splitViewHandle`.

core/src/SplitView.html
```diff
@@ -0,0 +1,333 @@
...
+  export default {
+    oncreate() {
+      this.onResize();
+      window.addEventListener('message', messageHandler.bind(this));
+    },
+    onupdate({ changed, current, previous }) {
+      getNode(this, this.get().nodepath);
+    },
+    ondestroy() {
+      window.removeEventListener('message', messageHandler.bind(this));
```
maxmarkus: Event listener removal does not work, tbd in follow up ticket.

client/src/splitViewHandle.js
```diff
@@ -0,0 +1,176 @@
...
+    return !this.splitView.collapsed;
+  }
+
+  /**
+   * Closes and destroys the split view
+   * @memberof splitView
+   * @example
+   * splitViewHandle.close();
+   */
+  close() {
```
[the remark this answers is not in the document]
maxmarkus: done

core/examples/luigi-sample-angular/src/app/project/project.component.ts
maxmarkus: done

client/src/splitViewHandle.js
```diff
@@ -0,0 +1,168 @@
+import { LuigiClientBase } from './baseClass';
+import { helpers } from './helpers';
+
+/**
+ * Split view 
+  Allows to open a micro-frontend in a split screen in the lower part of the content area. Open it by calling `const splitViewHandle = LuigiClient.linkManager().openAsSplitView`. 
+  At a given time, you can open only one split view. It closes automatically when you navigate to a different route.
+  When you call `handle.collapse()`, the split view gets destroyed. You can create a new frame by calling `handle.expand()`.
```
bszwarc:
```suggestion
  When you call `handle.collapse()`, the split view gets destroyed. It recreates when you use `handle.expand()`.
```

PR #635, opened 2019-07-08
I18n manage current locale in luigi

pekura:

PR #651, opened 2019-07-12
Post message target handling in Luigi Client

JohannesDoberer:


README.md
```diff
@@ -64,6 +64,8 @@ For security reasons, follow these guidelines when developing a micro frontend:
 - Make the micro frontend accessible only through HTTPS.
 - Add Content Security Policies (CSPs).
 - Make the Access-Control-Allow-Origin HTTP header as restrictive as possible.
+- Maintain a whitelist with trusted domains and compare it with the origin of luigi core app. The origin will be passed when you call the init listener in your micro frontend. Stop further processing in case of a non-matching origin.
```
bszwarc:
```suggestion
- Maintain a whitelist with trusted domains and compare it with the origin of Luigi core application. The origin will be passed when you call the init listener in your micro frontend. Stop further processing if the origin does not match.
```

client/src/lifecycleManager.js
```diff
@@ -29,16 +29,18 @@ class LifecycleManager extends LuigiClientBase {
 
     /**
      * Iterates over an object and executes all top-level functions
-     * with a given payload.
+     * with a given payload and the Luigi core domain
      * @private
      */
-    const _callAllFns = (objWithFns, payload) => {
+    const _callAllFns = (objWithFns, ...names) => {
```
jesusreal: There is one call to this function which is not sending the origin (`luigi.navigate` listener). Anyway what about calling here `Helpers.getLuigiCoreDomain()` instead of having the origin as input parameter?

client/src/helpers.js
```diff
@@ -60,6 +61,20 @@ class Helpers {
...
+    return this.origin;
+  }
+
+  setLuigiCoreDomain(origin) {
+    if (origin) {
+      this.origin = origin;
+    }
+  }
+
+  sendPostMessageToLuigiCore(msg) {
```
jesusreal: There are still two places where we are not using this method and still sending (`luigi.navigate` listener), please update those calls.

client/src/lifecycleManager.js
```diff
@@ -129,15 +131,15 @@ class LifecycleManager extends LuigiClientBase {
   }
 
   /**
-   * Registers a listener called with the context object as soon as Luigi is instantiated. Defer your application bootstrap if you depend on authentication data coming from Luigi.
+   * Registers a listener called with the context object and the Luigi core domain as soon as Luigi is instantiated. Defer your application bootstrap if you depend on authentication data coming from Luigi.
```
jesusreal: Should we also call the context update listeners with the Luigi Core domain for consistency?

jesusreal:
```suggestion
   * Registers a listener called with the context object and the Luigi Core domain as soon as Luigi is instantiated. Defer your application bootstrap if you depend on authentication data coming from Luigi.
```

client/src/lifecycleManager.js
```diff
@@ -29,16 +29,18 @@ class LifecycleManager extends LuigiClientBase {
 
     /**
      * Iterates over an object and executes all top-level functions
-     * with a given payload.
+     * with a given payload and the Luigi core domain
      * @private
      */
-    const _callAllFns = (objWithFns, payload) => {
```
jesusreal: Please note that once [this pr](https://github.com/SAP/luigi/pull/635/files#diff-18560a5dd187d564cad88d5a36754bce) is merged we will need to update the changes done to this function.

client/src/lifecycleManager.js
JohannesDoberer: talked with Philipp, we don't need this in luigi.navigate.

client/src/lifecycleManager.js
JohannesDoberer: do we need it in context update listeners? I don't think so.

client/src/lifecycleManager.js
jesusreal: Ok cool. Is there any special reason why we are we using "...names" here? Maybe we could use `(objWithFns, payload, luigiCoreOrigin)`

JohannesDoberer: good point

PR #678, opened 2019-07-24
Client permissions available in the client

JohannesDoberer:


core/examples/luigi-sample-angular/src/luigi-config/extended/projectDetailNav.js
```diff
@@ -64,6 +64,12 @@ export const projectDetailNavStructure = projectId => [
     viewUrl: '/sampleapp.html#/projects/' + projectId + '/developers',
     icon: '/assets/favicon-sap.ico'
   },
+  {
+    pathSegment: 'clientPermissionsTets',
+    label: 'ClientPermissionsTets',
+    viewUrl: '/sampleapp.html#/projects/pr1',
+    hideFromNav: true
```
jesusreal: Do you think we should expose the node so that everyone can see this feature in action?

client/src/lifecycleManager.js
```diff
@@ -235,5 +235,14 @@ class LifecycleManager extends LuigiClientBase {
   getPathParams() {
     return this.currentContext.pathParams;
   }
+
+  /**
+   * Returns the current client permissions as specified on the navigation node. (See [Node parameters](navigation-parameters-reference.md).)
```
bszwarc:
```suggestion
   * Returns the current client permissions as specified in the navigation node. For details, see [Node parameters](navigation-parameters-reference.md).
```

client/src/lifecycleManager.js
```diff
@@ -235,5 +235,14 @@ class LifecycleManager extends LuigiClientBase {
...
+   * @returns {Object} client permissions as specified on the navigation node.
```
bszwarc:
```suggestion
   * @returns {Object} client permissions as specified in the navigation node.
```

core/examples/luigi-sample-angular/e2e/tests/luigi-client-ux-manager-features.spec.js
```diff
@@ -336,6 +337,21 @@ describe('Luigi client ux manager features', () => {
           .find('[data-cy=luigi-current-locale]')
           .should('contain', "'pl_PL'");
       });
+      it('clientPermissions: check if set localization in client is disabled', () => {
```
jesusreal: Let´s use the new custom command we added recently and make the test shorter ;)
```
it('clientPermissions: check if set localization in client is disabled', () => {
        cy.visit('/projects/pr1/clientPermissionsTets')
          .getIframeBody()
          .then(function($body) {
            cy.wrap($body)
              .find('[data-cy=luigi-input-locale]')
              .should('be.disabled');

            cy.wrap($body)
              .find('[data-cy=set-current-locale]')
              .should('be.disabled');
          });
      });
```

core/examples/luigi-sample-angular/src/luigi-config/extended/projectDetailNav.js
JohannesDoberer: We have it in action in project 1 in currentLocale. It's just true. But we can discuss it in scrum.

jesusreal: True. ok for me to leave it as it is.

PR #697, opened 2019-07-30
Callback documentation for LuigiClient.addInitListener

maxmarkus:
Added documentation for LuigiClient.addInitListener callback documentation.


client/src/lifecycleManager.js
```diff
@@ -154,6 +154,12 @@ class LifecycleManager extends LuigiClientBase {
     }
     return id;
   }
+  /**
+   * Callback of the addInitListener
+   * @callback Lifecycle~initListenerCallback
+   * @param {Object} context current context data
+   * @param {string} origin luigi core origin
```
jesusreal: Maybe we could use _Luigi Core_ instead ;)

website/landingpage/public/styles.css
```diff
@@ -0,0 +1,23 @@
+body {
```
jesusreal: This looks like a merge issue or so, please check

client/src/lifecycleManager.js
```diff
@@ -107,10 +107,10 @@ class LifecycleManager extends LuigiClientBase {
    * @private
    * @memberof Lifecycle
    */
-  _callAllFns(objWithFns, payload) {
+  _callAllFns(objWithFns, payload, origin) {
```
jesusreal: This method is being called by the `_notifyUpdate` without _origin_, which means we are calling the listener handler in this case with origin _undefined_ I guess that is fine since _undefined_ is the same as no parameter.

maxmarkus: yes, this is intentionally

PR #705, opened 2019-07-31
Core api for getting translation for a specified key for given locale

JohannesDoberer:


core/src/TopNavDropDown.html
```diff
@@ -119,6 +120,9 @@ <h5 class="fd-action-bar__title fd-has-type-1">{node.label}</h5>
     helpers: {
       hasOpenUIicon(node) {
         return NavigationHelpers.isOpenUIiconName(node.icon);
+      },
+      getNodeLabel(node) {
```
marynaKhromova: I've noticed we define the same function in 2 more components. Maybe it makes sense to find the central place for it to avoid duplicates.

JohannesDoberer: First issue (click twice will fixed in a different issue)

JohannesDoberer: second issue: imo it is fine, because we already have centralized `NavigationHelpers.isOpenUIiconName(node.icon)` and hasOpenUIicon is fine... we need it as a **delegate** function.

docs/general-settings.md
```diff
@@ -27,3 +34,4 @@ You can set the following values:
   * `semiCollapsible` displays the arrow button at the bottom of the left side navigation. Once you click the button, the navigation shows up or collapses.<br>
 If you don't specify any value for  **responsiveNavigation**, the buttons remain hidden. The same applies when you enable **hideSideNav** for the currently active navigation node. 
 * **sideNavFooterText** is a string displayed in a sticky footer inside the side navigation. It is a good place to display the version of your application.
+* `customTranslationImplementation` **[function](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/function)** is executed when Luigi Application will be initialized. You can provide a custom function that translates text for the specified key in the current locale or in the specified one.
```
mmitoraj:
```suggestion
* **customTranslationImplementation** executes during the Luigi Application initialization. It allows you to provide a custom function that translates text for the specified key in the current locale or in the specified one.
```

docs/luigi-core-api.md
```diff
@@ -276,3 +276,14 @@ Unregisters a listener for locale changes.
 **Meta**
 
 -   **since**: 0.5.3
+
+#### getTranslation
+
+Gets translated text for the specified key in the current locale or in the specified one.
+Property values for token replacement in the localization key will be taken from the specified interpolations object.
```
mmitoraj:
```suggestion
Property values for token replacement in the localization key are taken from the specified interpolations object.
```

docs/luigi-core-api.md
```diff
@@ -276,3 +276,14 @@ Unregisters a listener for locale changes.
...
+
+##### Parameters
+
+-   `key` **[string](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)** key to be translated
+-   `interpolations` **[Object](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object)** objects with properties that will be used for token replacements in the localization key
```
mmitoraj:
```suggestion
-   `interpolations` **[Object](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object)** objects with properties  used for token replacement in the localization key
```

docs/luigi-core-api.md
```diff
@@ -276,3 +276,14 @@ Unregisters a listener for locale changes.
...
+-   `locale` **locale** optional locale to get the translation for; default is the current locale
```
mmitoraj:
```suggestion
-   `locale` **locale** optional locale for which you can get the translation. The default is the current locale.
```

core/src/TopNavDropDown.html
marynaKhromova: ok, thanks for the explanation 👍

docs/luigi-core-api.md
```diff
@@ -292,3 +292,14 @@ Unregisters a listener for locale changes.
...
+
+#### getTranslation
+
+Gets translated text for the specified key in the current locale or in the specified one.
+Property values for token replacement in the localization key are taken from the specified interpolations object.
+
+##### Parameters
+
+-   `key` **[string](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)** key to be translated
+-   `interpolations` **[Object](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object)** objects with properties  used for token replacement in the localization key
```
mmitoraj:
```suggestion
-   `interpolations` **[Object](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object)** objects with properties used for token replacement in the localization key
```

PR #722, opened 2019-08-06
Fix broken micro frontend path routing

maxmarkus:
Proposal for fixing `isSameDomain()` check, which was not working when mf is using path based routing.


maxmarkus:
Not broken, it is implemented as it should be: 
- mf with _hash_ based routing works as is
- mf with _path_ based routing needs to specify a *viewGroup* parent nodes to have proper microfrontend routing without reload

PR #745, opened 2019-08-13
Extend iframe sandbox rules

JohannesDoberer:


core/test/utilities/helpers/iframe-helpers.spec.js
```diff
@@ -19,8 +20,9 @@ describe('Iframe-helpers', () => {
       },
       get: () => lastObj
     };
-
-    sinon.stub(LuigiConfig, 'getConfigValue');
+    sinon
+      .stub(LuigiConfig, 'getConfigValue')
+      .callsFake(() => customSandboxRules);
```
maxmarkus: You could just use a simple `.returns(customSandboxRules)`.

Issue #720, opened 2019-08-05
Fix UI compatibility tests execution


jesusreal:
Currently they are failing, we want to be able to use them again.

maxmarkus:
Fixed in #750  by using `cy.getIframeBody()` only, which now relies on `.iframe()` command to wait for iframe body. Also gracefully behave on uncaught runtime errors.

For testing against versions same or older 0.5.4, it is recommended to either update it with the code from the linked PR, or using e2e:open which is a bit more precise and does not fail so often.

Issue #714, opened 2019-08-02
Improve CI pipeline execution time


jesusreal:
Currently our CI pipeline is taking too long to run, we would like to apply some improvements on that area. These are some of the aspect we want to take a look:
- Caching
- No execution of NPM publish job while not being on master or a branch starting with "release-v" (i.e., should not be executed for prs)
- Parallelization is out of scope, as it does not work in current Cypress

**Timeboxed to 1.5 days**

PR #750, opened 2019-08-16
Pipeline execution time

maxmarkus:
Improvements for testrunner, using more lightweight and stable webserver which seems to cause some issues with Cypress. 
Improves #714 and #720


maxmarkus:
I lowered the parallel e2e tests to 2 until we have OSS approval from Cypress. Currently all tets are executed in all integration instances because we are out of runtime minutes for Cypress Dashboard runs.

maxmarkus:
Working now as expected. 
- improved caching
- only single e2e step until we have parallel solution
- npm step only running if target branch matches
  - standard barnch https://travis-ci.org/SAP/luigi/builds/575694558 
  - how it will look at release branch https://travis-ci.org/SAP/luigi/builds/575701489

PR #754, opened 2019-08-19
Ui test for active node

JohannesDoberer:


core/examples/luigi-sample-angular/src/app/project/developers/developers.component.ts
```diff
@@ -1,12 +1,40 @@
...
 @Component({
   selector: 'app-developers',
   templateUrl: './developers.component.html',
   styleUrls: ['./developers.component.css']
 })
-export class DevelopersComponent implements OnInit {
-  constructor() {}
+export class DevelopersComponent implements OnInit, OnDestroy {
+  private initListener;
+  private updateListener;
```
jesusreal: Correct me if I am wrong, but it looks like we don't need the update listener

core/examples/luigi-sample-angular/src/app/project/developers/developers.component.ts
```diff
@@ -1,12 +1,40 @@
...
+  updateVisitors() {
+    let visitors = window['visitors'] || 0;
+    window['visitors'] = ++visitors;
+    return visitors;
+  }
+
+  ngOnInit() {
+    this.initListener = addInitListener((context, origin) => {
+      this.visitors = this.updateVisitors();
+      console.log('visitorsInit ', this.visitors);
```
jesusreal: Maybe we don't need this console log

core/examples/luigi-sample-angular/e2e/tests/navigation.spec.js
```diff
@@ -33,6 +37,63 @@ describe('Navigation', () => {
     cy.get('.fd-app__sidebar').should('contain', 'Second Child');
   });
 
+  it('Check if active node is selected', () => {
```
jesusreal: I would suggest to use `cy.visit` instead of navigating manually and refactor a little bit to make the test shorter and more readable. For example, we don't need to use `should('exist')` after `find` because `find` will fail itself if it does not find the element. Here is my suggestion:
```
  it('Check if active node is selected', () => {
    cy.visit('/projects');
    cy.get('.fd-shellbar')
      .contains('Projects')
      .should('have.class', 'is-selected');

    cy.visit('projects/pr1');
    cy.get('.fd-side-nav__subitem')
      .contains('Project Settings')
      .click()
      .should('have.class', 'is-selected');
  });

  it('Check if active node reloads page', () => {
    cy.visit('/projects/pr1/developers');
    cy.getIframeBody().then($iframeBody => {
      cy.wrap($iframeBody)
        .should('contain', 'Developers content')
        .find('[title="visitors: 1"]')
      cy.get('.fd-app__sidebar')
        .contains('Project Settings')
        .click();
      cy.get('.fd-app__sidebar')
        .contains('Developers')
        .click();
      cy.wrap($iframeBody)
        .find('[title="visitors: 2"]')
    });
    cy.get('.fd-app__sidebar')
      .contains('Developers')
      .click();
    cy.getIframeBody().then($iframeBody => {
      cy.wrap($iframeBody)
        .find('[title="visitors: 1"]')
    });
  });
```

JohannesDoberer: 👍

core/examples/luigi-sample-angular/e2e/tests/navigation.spec.js
```diff
@@ -1,7 +1,11 @@
 describe('Navigation', () => {
+  let $iframeBody;
   beforeEach(() => {
     cy.visit('/');
     cy.login('tets@email.com', 'tets');
+    cy.getIframeBody().then(result => {
```
jesusreal: Please remove this code, since it is not needed

PR #760, opened 2019-08-26
Prevent luigi.auth.tokenIssued infinite loop

maxmarkus:
Prevents possible infinite loop for `luigi.auth.tokenIssued` event.


client/src/lifecycleManager.js
```diff
@@ -69,7 +69,7 @@ class LifecycleManager extends LuigiClientBase {
         this._notifyInit(e.origin);
       });
 
-      helpers.addEventListener('luigi.auth.tokenIssued', e => {
+      helpers.addEventListener('luigi-client.auth.tokenIssued', e => {
```
pekura: Doesn't this break backwards compatibility if only client OR core gets updated?

PR #763, opened 2019-08-26
Fix pipeline client caching

maxmarkus:
Moves bundled Luigi client to its own public folder.


client/package.json
```diff
@@ -1,7 +1,8 @@
 {
-  "name": "@kyma-project/luigi-client",
+  "name": "luigi-client-private",
   "version": "0.6.1",
-  "description": "Javascript library supporting consumers of the Luigi framework",
+  "private": true,
+  "description": "luigi-client development version, only the public folder is being published to npm as a package",
```
jesusreal: I would remove `keywords` and `publishConfig` from this file, as that is only relevant for publishing. 
Additionally, I would move `main` to `public/package.json`

PR #787, opened 2019-08-30
Initial app loading indicator

maxmarkus:
Add possibility to show a loading indicator:
- until configuration is ready
- until the first microfrontend is ready by calling `Luigi.afterInit()`


docs/general-settings.md
```diff
@@ -37,11 +39,13 @@ You can set the following values:
...
+```javascript
+{
+  getTranslation: (key, interpolations, locale) => {
+    // should return translation of the 'key' in the 'locale' or current locale
+  }
+}
+```
 > **NOTE:** You can translate Luigi internal messages by providing translation for [these keys](../core/src/utilities/defaultLuigiTranslationTable.js).
 * **customSandboxRules** is an array of custom rules for the content in the iframe. You can extend the [Luigi default sandbox rules](https://github.com/SAP/luigi/blob/af1deebb392dcec6490f72576e32eb5853a894bc/core/src/utilities/helpers/iframe-helpers.js#L140) by adding further rules.
+* **appLoadingIndicator.hideAutomatically** allows to disable the auto hiding of the app loading indicator, which is enabled by default in case the app loading indicator is being used. Take a look at the [App loading indicator](luigi-ux-features.md#app-loading-indicator) section on how to use this feature.
```
alexandra-simeonova:
```suggestion
* **appLoadingIndicator.hideAutomatically** allows you to disable automatic hiding of the app loading indicator, which is enabled by default in case the app loading indicator is being used. Take a look at the [App loading indicator](luigi-ux-features.md#app-loading-indicator) section on how to use this feature.
```

docs/luigi-ux-features.md
```diff
@@ -21,3 +21,35 @@ You can quickly adjust the Luigi application to improve user experience on mobil
...
+<div luigi-app-loading-indicator>
+  <div class="fd-spinner">
+    <div></div>
+  </div>
+</div>
+```
+
+By default, the loading indicator is removed after `Luigi.setConfig({})` has been executed. 
+
+Alternatively, to keep the loading indicator until the first micro frontend is usable, follow the following steps:
```
alexandra-simeonova:
```suggestion
Alternatively, to keep the loading indicator until the first micro frontend is usable, follow these steps:
```

docs/luigi-ux-features.md
```diff
@@ -21,3 +21,35 @@ You can quickly adjust the Luigi application to improve user experience on mobil
...
+
+1. Set app loading indicator `hideAutomatically` to `false`
```
alexandra-simeonova:
```suggestion
1. Set the app loading indicator parameter `hideAutomatically` to `false`
```

docs/luigi-ux-features.md
```diff
@@ -21,3 +21,35 @@ You can quickly adjust the Luigi application to improve user experience on mobil
...
+  settings: {
+    appLoadingIndicator: {
+      hideAutomatically: false
+    }
+  }
+  ...
+}
+```
+
+2. Call [`Luigi.ux().hideAppLoadingIndicator()`](./luigi-core-api.md#hideAppLoadingIndicator) in the Luigi Core once your initial micro frontend has finished loading to remove the loading indicator. You can for example use the [custom messages](./communication.md#custom-messages) feature for the micro frontend using Luigi Client to communicate the Core when this function should be executed.
```
alexandra-simeonova:
```suggestion
2. Call [`Luigi.ux().hideAppLoadingIndicator()`](./luigi-core-api.md#hideAppLoadingIndicator) in the Luigi Core once your initial micro frontend has finished loading to remove the loading indicator. You can, for example, use the [custom messages](./communication.md#custom-messages) feature to allow the Luigi Client micro frontend to communicate with the Core when this function should be executed.
```

core/src/core-api/ux.js
```diff
@@ -0,0 +1,26 @@
...
+ * @namespace UX
+ */
+class LuigiUX {
+  /**
+   * Hides the app loading indicator.
+   * @memberof UX
+   * @since 0.6.4
+   */
+  hideAppLoadingIndicator() {
+    const appLoadingIndicator = document.querySelector(
```
pekura: Why don't we just render the loading spinner and overlay in the place where luigi itself is rendered?

maxmarkus: Because Luigi itself is not rendered yet when we already show the loading spinner. We would need to create and set the spinner programmatically, which might take a while under some circumstances - we had it in place initially, but Philipp wanted to have it the way it is implemented now - simplification.

docs/luigi-ux-features.md
maxmarkus: done

core/src/core-api/ux.js
maxmarkus: At least we are consistent with the naming `luigi-app-*` selectors.

pekura: 👍

PR #795, opened 2019-09-04
Add checks for bundlesizes in CI pipeline

zarkosimic:
**Description**

- Bundlesize checks added in CI pipeline for both Luigi Core and Luigi Client

Here is an [example](https://travis-ci.org/zarkosimic/luigi/jobs/580303359#L894) of the failed build for a luigi-client which triggered by reducing the bundle size value of the luigi-client.js file.

Related to the example above, here is an [example](https://travis-ci.org/zarkosimic/luigi/jobs/580309879#L905) of the failed build for a luigi-core where the limit size for luigi.js file was changed.

[Successful build](https://travis-ci.org/zarkosimic/luigi/jobs/580314820) with correct bundle size checks.


core/package.json
```diff
@@ -51,27 +51,28 @@
...
     "test:watch": "npm run test -- --watch",
-    "bundlesize": "npm run bundle && bundlesize"
+    "bundlesize": "npm run bundle && bundlesize",
+    "bundlesizeOnly": "bundlesize"
   },
   "bundlesize": [
     {
       "path": "./public/luigi.js",
-      "maxSize": "100 kB",
+      "maxSize": "350 kB",
```
maxmarkus: Is this raise really necessary? I guess you tried with `npm run bundle-develop` .. please check with `npm run bundle`. Bundlesize in CI is always using production build. Also for other sizes, I guess they are fine as they have bene.

zarkosimic: I've tried with `npm run bundle` and the result for luigi.js is around 297KB. So, I gave it a little bit more space, counting it would be enough for the next few months. Any suggestions are welcome. 

With `npm run bundle-develop`, the result for luigi.js is 1.9MB.

maxmarkus: Well, it has been a long time since I really looked at the bundlesize and thought we always enforced it and it stays below 100kb. It looks like we have built tons of new stuff which brought in code size ;-) So let's leave it for now thanks.

zarkosimic:
CI result: 
https://travis-ci.org/zarkosimic/luigi/jobs/581609821#L885

Issue #809, opened 2019-09-09
Improve docs readme file


alexandra-simeonova:
Docs readme file: https://github.com/SAP/luigi/tree/master/docs

Changes proposed:
- Add a link to the Luigi application example 
- Add a link to the development guidelines 
- Elaborate on the difference between Luigi Core and Luigi Client with 1-2 sentences.

Reason:
I think a lot of people go straight to "docs" when looking for information. It is also linked at the top of the Luigi landing page. It might be useful to include a little more info there.

PR #813, opened 2019-09-09
Improve Luigi docs readme file

alexandra-simeonova:
**Description**

-Add a link to the Luigi application example
-Add a link to the development guidelines
-Elaborate on the difference between Luigi Core and Luigi Client with 1-2 sentences.

Reason:
I think a lot of people go straight to "docs" when looking for information. It is also linked at the top of the Luigi landing page. It might be useful to include a little more info there.

See also #809


docs/README.md
```diff
@@ -18,11 +18,17 @@ Set up and configure the main application.
...
-* [UX Manager](luigi-client-api.md#uxmanager) helps you to manage the appearance features in Luigi, such as the behavior of backdrop or loading indictors.
+* [UX Manager](luigi-client-api.md#uxmanager) helps you manage the appearance features in Luigi, such as the behavior of backdrop or loading indictors.
 
-## Common
-* [Communication](communication.md) describes how to configure features related to communication between Luigi Core and Client.
+## Examples
+
+Follow the Luigi [application examples](core/examples) for an in-depth look at Luigi capabilities. 
+
+## Development Guidelines
```
bszwarc:
```suggestion
## Development guidelines
```
I would use lowercase for the second word - https://github.com/kyma-project/community/blob/master/guidelines/content-guidelines/04-formatting.md#headings

docs/README.md
```diff
@@ -18,11 +18,17 @@ Set up and configure the main application.
...
 * [Lifecycle](luigi-client-api.md#lifecycle) gives you details on the lifecycle of listeners, navigation nodes, and Event data.
 * [Link Manager](luigi-client-api.md#linkmanager) allows you to navigate to a specific route. 
-* [UX Manager](luigi-client-api.md#uxmanager) helps you to manage the appearance features in Luigi, such as the behavior of backdrop or loading indictors.
+* [UX Manager](luigi-client-api.md#uxmanager) helps you manage the appearance features in Luigi, such as the behavior of backdrop or loading indictors.
 
-## Common
-* [Communication](communication.md) describes how to configure features related to communication between Luigi Core and Client.
+## Examples
+
+Follow the Luigi [application examples](core/examples) for an in-depth look at Luigi capabilities.
```
bszwarc:
```suggestion
Follow the Luigi [application examples](../core/examples/README.md) for an in-depth look at Luigi capabilities. 
```
This link won't work, because examples are not in the `docs` directory - it worked in the main README, as `core/examples` is located in the main Luigi directory. To make it work, you can either use the absolute path (https://github.com/SAP/luigi/tree/master/core/examples) or a relative one `(../core/examples)` to go one level up and then point to the right directory. :)
I pointed to the readme.md file directly :) 
To check the links, the best way is to view this file as rendered Markdown (using the `...` > View File in the right top corner (under Add your review). Of course, sometimes they won't work - if you are, for example, linking to a doc which you add in the same PR - but if the doc exists already it should be easy to check if the link is not broken.

PR #832, opened 2019-09-16
Set userinfo without auth, new navigation.profile.staticUserInfoFn

maxmarkus:
Added `navigation.profile.staticUserInfoFn` setting which allows to set a static or async function to retrieve and return an object with `name` and `email` properties.

You can test it by disabling the `use = "mockAuth"` line in luigi-config/extended/ auth.js and defining your settings in navigation.js profile configuration.


docs/navigation-parameters-reference.md
```diff
@@ -265,6 +265,7 @@ The profile section is a configurable drop-down list available in the top naviga
   - **externalLink** is an object which indicates that the node links to an external URL. If this parameter is defined, the **link** parameter is ignored. It has the following properties:
     - **sameWindow** defines if the external URL is opened in the current tab or in a new one. The default value for this parameter is `false`.
     - **url** is the external URL that the link leads to.
+- **staticUserInfoFn** is a function that is used to retrieve a user's name and email to simulate a login. It can be used with authorization is disabled and also gets called if the defined IDP provider does not have either `settings.userInfoFn` defined or does not provide a `userInfo` function internally. It can be asynchronous and should return an object with `name`, `email` and `picture` properties.
```
```suggestion
- **staticUserInfoFn** is a function that is used to retrieve a user's name and email to simulate a login. It can be used when authorization is disabled and also gets called if the defined IDP provider does not have `settings.userInfoFn` defined or does not provide a `userInfo` function internally. It can be asynchronous and should return an object with `name`, `email` and `picture` properties.
```

alexandra-simeonova: Alternatively: "It can be used with disabled authorization "/ "It can be used if authorization is disabled"?

PR #881, opened 2019-10-09
Add and describe "allowRules" in the general settings

zarkosimic:

- Add `allowRules` to the example in the general settings
- Add a description for `allowRules`


docs/general-settings.md
```diff
@@ -50,4 +51,5 @@ If you don't specify any value for  **responsiveNavigation**, the buttons remain
 ```
 > **NOTE:** You can translate Luigi internal messages by providing translation for [these keys](../core/src/utilities/defaultLuigiTranslationTable.js).
 * **customSandboxRules** is an array of custom rules for the content in the iframe. You can extend the [Luigi default sandbox rules](https://github.com/SAP/luigi/blob/af1deebb392dcec6490f72576e32eb5853a894bc/core/src/utilities/helpers/iframe-helpers.js#L140) by adding further rules.
+* **allowRules** is an array of rules for the content in the iframe, managed by "allow" attribute. You can use one or more rules by adding them in the array(e.g. `allowRules: ['microphone', 'camera']`). Be aware that this mechanism requires the browser to support [Feature Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Feature_Policy).
```
alexandra-simeonova:
```suggestion
* **allowRules** is an array of rules for the content in the iframe, managed by the HTML **allow** attribute. You can use one or more rules by adding them to the array, for example `allowRules: ['microphone', 'camera']`. Be aware that this mechanism requires the browser to support [Feature Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Feature_Policy).
```

alexandra-simeonova: Good job! I second Johannes' comment that maybe we shouldn't have quotes. I also wanted to specify that the allow attribute is not a Luigi attribute, hence adding the HTML. Lastly, Barbara told me that parentheses should be avoided in documentation. 😄

PR #901, opened 2019-10-17
Inactive lifecycle hook

maxmarkus:
Adds `addInactiveListener` and `removeInactiveListener` to client api for receiving notification upon micro frontend inactivity.

A listener is already in the Angular example. It can be tested by using *preserveView* linkManager feature and by clicking to and from a *viewGrouped* navigation item (Settings gear icon).


client/src/lifecycleManager.js
```diff
@@ -201,6 +215,34 @@ class LifecycleManager extends LuigiClientBase {
     return false;
   }
 
+  /**
+   * Registers a listener called upon microfrontend inactivity. This happens when a another micro frontend gets shown while keeping the old one cached.
+   * Gets called when:
+   * - navigating with **preserveView**
+   * - navigating from or to a **viewGroup**
+   * Does not get called when navigating normally, or when `openAsModal` or `openAsSplitView` are used.
```
alexandra-simeonova: Line 223 does not display correctly in the .md file (it is concatenated with the last bullet point). Maybe adding a new empty line before it would fix this?

PR #923, opened 2019-10-23
Refactor authentication

maxmarkus:
Refactored hardcoded `localStorage.. 'luigi.auth'` implementation to custom store which is exposed to Core API. Added new `auth.storage` configuration setting to select between `localStorage`, `sessionStorage` or `none` for simple in-memory storage. Default is `localStorage` to be backwards compatible.


docs/luigi-core-api.md
```diff
@@ -255,6 +256,94 @@ Luigi.auth().isAuthorizationEnabled();
 
 Returns **[boolean](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Boolean)** true if authorization is enabled. Otherwise returns false.
 
+### AuthorizationStore
+
+Authorization Storage helpers, to be used in your custom authorization provider.
+Read more about [custom authorization providers]\(authorization-configuration.md#Implement a custom authorization provider).
```
alexandra-simeonova:
```suggestion
Read more about [custom authorization providers](authorization-configuration.md#implement-a-custom-authorization-provider).
```

core/test/services/auth-store.spec.js
```diff
@@ -0,0 +1,242 @@
...
+      //     }
+      //   };
+      // }
+      // global['localStorage'] = mockStorage();
+      // global['sessionStorage'] = mockStorage();
+    });
+    beforeEach(() => {
+      AuthStoreSvc._setStore.restore();
+      AuthStoreSvc._getStore.restore();
+      // global['sessionStorage'] = {
```
JohannesDoberer: If this is comment out tets are failing locally. 
In line 74 
> Cannot read property 'getItem' of undefined

PR #925, opened 2019-10-24
provide possibility to add alt attribute to the <img> tag

marynaKhromova:

- add alt attribute to the logo
- add alt attribute to the img tags in top and left navigation
- fix for the height of a product swither icon when it's 
- a small html fix for product switcher on mobile 
- docu changes


marynaKhromova:
> I found couple of places where the alt attribute is missing or not working. I created a patch for those places:
> 
> [patch.diff.zip](https://github.com/SAP/luigi/files/3782378/patch.diff.zip)
> 
> Please take a look and apply if you agree. Thank you!

Thank you, Peter! really good one 👍 the changes are done.

PR #933, opened 2019-10-25
landingpage improvements 

marynaKhromova:

- fix a bug for the footer in safari
- fix a bug in FF for the footer background
- minor content improvements for about page


marynaKhromova:
> Please lowercase the image file names.
> And would it make sense to link the _How to get started_ link to docs/getting-started.md instead of the direct link to core setup?

Thank you Markus :) The link has been changed. The names of images were already with lowercase, I just removed the old images from the public folder.

PR #943, opened 2019-11-04
Unit tests - part 1

zarkosimic:
Increase tests coverage


core/test/services/navigation.spec.js
```diff
@@ -319,6 +458,231 @@ describe('Navigation', function() {
...
+          pathSegment: 'projects',
+          children: [
+            {
+              pathSegment: 'settings1'
+            }
+          ]
+        }
+      ];
+      const result = Navigation.getNodes(children, pathData);
+      expect(result).to.be.deep.equal([]);
```
maxmarkus: Can we use the same assert command please in this file? `assert.deepEqual(`

core/test/utilities/helpers/async-helpers.spec.js
```diff
@@ -5,11 +5,64 @@ const assert = chai.assert;
...
+      AsyncHelpers.waitForKeyExistency(obj, 'client').then(result => {
+        assert.equal(result, true);
+        done();
+      });
+      setTimeout(() => {
+        obj.client = 'yes';
+      }, 500);
+      clock.tick(1000);
+    });
+    it('returns message if key exists after 20000ms', done => {
```
maxmarkus: we can remove that test, the "if key does not exist" is valid for both cases. maybe rename the one below to "if key does not exist after timeout"

PR #944, opened 2019-11-04
Luigi example react

JohannesDoberer:


core/examples/luigi-example-react/README.md
```diff
@@ -0,0 +1,38 @@
...
+3. Open it in your browser by going to [http://localhost:3000/](http://localhost:3000/).
+
+
+### Build
+
+```
+npm run build
+```
+
+Builds the app for production to the `build` folder.<br />
```
alexandra-simeonova: Maybe it's better if we follow the same format as the Vue/Angular example? The text there is:

```
### Build
`npm run build`

The build compiles and minimizes the source files for production usage. The build generates a dist folder which you can serve using a web server.
```

PR #952, opened 2019-11-06
Cm fix

JohannesDoberer:


website/docs/src/luigi-config/extended/navigation.js
```diff
@@ -4,20 +4,20 @@ if (location.port == '4000') {
...
-        .then(function(json) {
-          return json.map((child) => {
-            child.viewUrl = child.viewUrl.replace('__BASE_URL__', baseUrl);
-            return child;
-          });
-        }).catch(function(err) {
-          console.error(`Error: ${err}`);
-        })
+    .then((res) => {
+      res.setHeader("Strict-Transport-Security", "max-age=31536000; includeSubDomains; preload");
```
maxmarkus: Issue of the runtime error:
This `res` is not an express/nodejs request handle, its an fetch response object. `fetch` itself needs to be configured to use the header: 
```
I pushed the fix for it.
return fetch('/navigation-children.json', {
      headers: {
        'Strict-Transport-Security': 'max-age=31536000; includeSubDomains; preload'
      }
    })
    .then((res) => {
      return res.json();
    })
```

Issue #955, opened 2019-11-06
Improve application-setup.md


alexandra-simeonova:
Improve the application setup document with better formatting. Before, it was not clear whether the "Basic setup" steps actually had to be performed by the user. The goal is to make the meaning of this section clearer.

PR #965, opened 2019-11-13
Improve application-setup.md

alexandra-simeonova:
Improve the application setup document with better formatting. Before, it was not clear whether the "Basic setup" steps actually had to be performed by the user. The goal is to make the meaning of this section clearer. 

See: #955


docs/application-setup.md
```diff
@@ -1,49 +1,37 @@
...
 
-Choose the framework to build your application:
-
-[Application setup without a framework](#noframework)  
-[Angular 6](#angular6)  
-[SAPUI5/OpenUI5](#sapui5)  
-[VUE.JS](#vuejs)
-
-## Basic application setup
+All those steps are condensed in a single installer script which sets up a Luigi application on your system. Keep in mind that if you have a single page application, you need to transfer it to a two-page application. Your application should run to make sure the existing router doesn't interfere.
```
bszwarc: The last sentence is a bit confusing to me: does it mean that the application needs to be running to check if there is no interference with the router or that you should run it to see if there is no such interference?

alexandra-simeonova: It's the second one :) I changed it to "You should run your application to make sure the existing router doesn't interfere."

bszwarc: Then I would put it this way - it's more direct and clear.

PR #983, opened 2019-11-21
Display the docu aligned in the middle

maxmarkus:
Just a poc to show how the docu would show middle-aligned. 

Q: Hey, the scrollbar is somewhere in the middle of the screen, can you do something about it?
A: Yes, we can move the scrollbar also to the very right side of the window, especially for our windows users. This requires some extra steps, but is also almost ready in a stash.


maxmarkus:
Decided not to change the current state.

PR #984, opened 2019-11-21
change documentation link

marynaKhromova:
I've updated the documentation link on landingpage.


website/landingpage/public/about.html
```diff
@@ -37,8 +37,7 @@
               </li>
               <li class="item">
                 <a
```
JohannesDoberer: https://github.com/SAP/luigi/blob/df60c86d09f2f46c2c1abfbc108c13c8612bfdd2/website/landingpage/public/about.html#L99
Do we need to change this as well?

marynaKhromova: good catch! thank you :) done.

PR #992, opened 2019-11-27
Edge browser back

maxmarkus:
Fixes browser back issue in MS Edge and IE11.


core/package.json
```diff
@@ -52,7 +52,7 @@
     "bundle-ie11": "webpack --mode=production --display-error-details --config webpack-ie11.config.js",
     "bundle-develop": "npm run bundle-develop-evergreen",
     "bundle-develop-evergreen": "npm run bundle-evergreen -- -d --watch",
-    "bundle-develop-ie11": "webpack --display-error-details --config webpack-ie11.config.js --debug --devtool cheap-source-map --output-pathinfo --watch",
+    "bundle-develop-ie11": "MINIFY=false webpack --display-error-details --config webpack-ie11.config.js --debug --devtool source-map --output-pathinfo --watch",
```
pekura: I think this won‘t work for ie11 as the non-cheap source-map uses eval and the code in the eval block is not transpiled.

PR #994, opened 2019-11-29
handle init in luigi client without init handshake

JohannesDoberer:


JohannesDoberer:
For testing unzip and run these two folders on different webserver. May you have to change the port in the index files.
[https://app.zenhub.com/files/139590701/4fc8897b-242d-4f60-b927-7eedf5f1527a/download](https://app.zenhub.com/files/139590701/4fc8897b-242d-4f60-b927-7eedf5f1527a/download)

PR #1001, opened 2019-12-03
Add en e2e test scenario for goBack and passing data

ndricimrr:
**Description**

Added an E2E scenario for goback and passing data 

- ...
- ...
- ...

**Related issue(s)**


core/examples/luigi-sample-angular/e2e/tests/luigi-client-preserved-view-params.spec.js
```diff
@@ -0,0 +1,74 @@
+Cypress.env('RETRIES', 1);
```
maxmarkus: Please include the whole file into `luigi-client-link-manager-features.spec.js`, all files should be kind of equally long and optimally consume more then 30 seconds to test, else the parallelization is slower then if we would run it all together. You can take a look at the current test runs and see that this spec file takes the same time as the cleanup between test runs on that machine. https://dashboard.cypress.io/projects/czq7qc/runs

PR #1002, opened 2019-12-03
Implicit structural nodes

maxmarkus:
Replace the environments node `extended/navigation.js:51`
```
{
      hideFromNav: true,
      pathSegment: 'environments/:environmentId',
      viewUrl: '/sampleapp.html#/environments/:environmentId'
},
```


docs/navigation-configuration.md
```diff
@@ -93,11 +93,32 @@ navigation: {
...
 	- `link: '/home/projects'` is correct
 	- `link: '/projects'`is not correct, since `/projects` is not the root node
 * **externalLink** - takes you to an external website. You must define the **url** of the website and whether it should be opened in a new window (`sameWindow: false`) or the same window (`sameWindow: true`).
+
+You can also specify a nested node as `project/:id`. It can only be used standalone, with no other nodes having the same or partially overlapping pathSegments (eg. `projects` and `projects/:id` do not work together).
+
+```
+// structural node example
+{
+  viewUrl: '//mf.url/projects/:id',
```
alexandra-simeonova:
```suggestion
  viewUrl: '//example.com/projects/:id',
```

docs/navigation-configuration.md
```diff
@@ -93,11 +93,32 @@ navigation: {
...
+  pathSegment: `project/:id`,
+  context: { id: ':id' }
+}
+// gets internally expanded to this node structure
+{
+  viewUrl: 'project',
+  children: [{
+    viewUrl: '//mf.url/projects/:id',
```
alexandra-simeonova:
```suggestion
    viewUrl: '//example.com/projects/:id',
```

core/src/navigation/services/navigation.js
```diff
@@ -87,33 +88,60 @@ class NavigationClass {
...
+
+  getExpandStructuralPathSegment(node) {
     // Checking for pathSegment to exclude virtual root node
-    if (node && node.pathSegment && node.children) {
-      node.children.forEach(child => {
-        child.parent = node;
-      });
+    if (node && node.pathSegment && node.pathSegment.indexOf('/') !== -1) {
+      const segs = node.pathSegment.split('/');
+      const clonedNode = merge({}, node);
```
pekura: We can get rid of the 'lodash.merge' dependency here by doing just this: 

`const clonedNode = { ...node };`

core/src/utilities/helpers/generic-helpers.js
```diff
@@ -59,6 +61,14 @@ class GenericHelpersClass {
     return this.deepMerge(target, ...sources);
   }
 
+  /**
```
pekura: It looks like this method is not used.

core/src/navigation/services/navigation.js
maxmarkus: Unfortunately this is not working.
I need to deep clone properly without keeping references. Try this in the console
```
const node = { label: 'a' };
const context = { key: 'value' };
node.context = context;

const copy = { ...node };
context.key = 'changed!';
// copy context.key should still be 'value' if its a deep clone.
console.log(copy.context.key)
```

docs/navigation-configuration.md
maxmarkus: Maybe we should think of a separation of luigi url and microfrontend url. Thats why i put there mf.url instead of example.com, because example.com could also be the luigi core domain, and might confuse people.

core/src/navigation/services/navigation.js
pekura: I understand, but why do we need a deep copy here? 

From

```
{
	pathSegment: 'environments/:environmentId'
	viewUrl: 'sampleapp.html/#envs/:environmentId',
	context: {
		key: 'value'
	}
}
```

we do

```
{
	pathSegment: '/environments'
	children: [
		{
			pathSegment: ':environmentId'
			viewUrl: 'sampleapp.html/#envs/:environmentId',
			context: {
				key: 'value'
			}
		}
	]
}
```

Yes, the context is the same object (and other stuff like children - if the node would have children -) so this is not a pure non-mutable operation but in the current state we reuse the configuration node tree anyway so this is not a big deal. We use the contexts of navigation nodes as they are. In this case we just "attach" context, children etc to the newly created node.   Well, it is not ideal atm but the non-deep copy of node attributes would be closer to the current "architecture" :-). 
In the future we might go with a solution where we do not touch the navigation tree at all and build our own representation based on the configuration but that would be a bigger refactoring.
We can discuss it in an after-scrum.

PR #1008, opened 2019-12-05
custom item renderer for options in context switcher

JohannesDoberer:


core/src/navigation/ContextSwitcher.html
```diff
@@ -49,10 +49,14 @@
               on:click="{() => goToOption(node, selectedOption)}"
               data-testid="{getTestId(node)}"
             >
+              {#if customOptionsRenderer}
+              {@html customOptionsRenderer(node)}
+              {:else}
               <a
                 class="fd-menu__item {node === selectedLabel ? 'is-selected' : ''}"
```
maxmarkus: How is the "selected" reflected for custom renderers? Guess we should add that `selectedLabel` as a parameter for `customOptionsRenderer`.

core/examples/luigi-sample-angular/src/luigi-config/extended/settings.js
```diff
@@ -14,7 +14,6 @@ class Settings {
   responsiveNavigation = 'simpleMobileOnly'; // Options: simple | simpleMobileOnly | semiCollapsible
   sideNavFooterText = `Luigi Client: ${version || 'unknown'}`;
   thirdPartyCookieCheck = {
-    thirdPartyCookieDetection: true,
```
maxmarkus: Since you commented out the `thirdPartyCookieScriptLocation` it might make sense to also just comment out that line?

core/src/navigation/ContextSwitcher.html
JohannesDoberer: 👍 good catch!

docs/navigation-parameters-reference.md
```diff
@@ -262,6 +262,11 @@ The context switcher is a drop-down list available in the top navigation bar. It
   - **label** defines the context element label. If not defined, the **pathValue** is passed to **fallbackLabelResolver** to set its value. The default value is **pathValue**, if **fallbackLabelResolver** is not defined.
   - **pathValue** defines the context element path that is appended to **parentNodePath** and reflects a **pathSegment**.
 
+### customOptionsRenderer
+- **type**: function
+- **parameter**: option, isSelected
+- **description**: possibility to render code inside an `<li>` element. The list element has already registered a navigation function to the options path. The function gets an option as parameter. It is recommended to use this function carefully because it is possible to inject javascript code.
```
```suggestion
- **description**: enables you to add custom elements to the context switcher by rendering code inside a `<li>` element. The function takes an **option** object as a parameter. It is recommended to use this function carefully because it is possible to inject JavaScript code.
```

PR #1014, opened 2019-12-09
Disable context switcher drop down caret when there is only one option

pekura:


docs/navigation-parameters-reference.md
```diff
@@ -282,6 +282,10 @@ The context switcher is a drop-down list available in the top navigation bar. It
 - **type**: boolean
 - **description**: if set to `true`, the sub-path is preserved on context switch.
 
+### alwaysShowDropdown
+- **type**: boolean
+- **description**: if set to `false`, the dropdown is not shown on click if there is only one option and no actions
+- **default**: true
```
```suggestion
- **default**: `true`
```

docs/navigation-parameters-reference.md
```diff
@@ -282,6 +282,10 @@ The context switcher is a drop-down list available in the top navigation bar. It
 - **type**: boolean
 - **description**: if set to `true`, the sub-path is preserved on context switch.
 
+### alwaysShowDropdown
+- **type**: boolean
+- **description**: if set to `false`, the dropdown is not shown on click if there is only one option and no actions
```
```suggestion
- **description**: if set to `false`, the drop-down is not shown on click if there is only one option and no actions.
```

Issue #1016, opened 2019-12-11
Improve Luigi API documentation 


alexandra-simeonova:
There are some things that can be improved about the Luigi (Core and Client) API documentation, such as:

* Some functions are missing from Luigi Core API (e.g. showAlert)
* We can look into options to improve the structure as it is just one long list (e.g. putting links at the top or a navigation structure on the documentation page)

PR #1025, opened 2019-12-17
Improve API documentation

alexandra-simeonova:
**Description**

- add overviews and structure elements
- add a few missing functions, specifically the UX part of Luigi Core API

**Related issue(s)**
#1016


core/src/core-api/i18n.js
```diff
@@ -30,7 +30,7 @@ class LuigiI18NManager {
 
   /**
    * Gets the current locale.
-   * @returns {string} current locale
+   * @returns {string}: current locale
```
maxmarkus: Type definitions cannot be followed by a colon. If you run `lerna run docu` you see that it is broken. Please change all occurrences in this file.

docs/luigi-core-api.md
```diff
@@ -644,3 +654,80 @@ Hides the app loading indicator.
...
 
 -   **since**: 0.6.4
+
+#### showAlert
+
+Shows an alert.
+
+##### Parameters
+
+-   `settings` **[Object](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object)** the settings for the alert
```
bszwarc: Some of the sentences do have dots at the end some don't - but I think it may be this rendering problem you told me about. Also, it's up to you to have / not have them as long as they are consistent :)

docs/luigi-core-api.md
```diff
@@ -644,3 +654,80 @@ Hides the app loading indicator.
...
+
+-   **since**: 0.6.4
+
+#### showConfirmationModal
+
+Shows a confirmation modal.
+
+##### Parameters
+
+-   `settings` **[Object](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object)** the settings of the confirmation modal. If you don't provide any value for any of the fields, a default value is used
```
bszwarc:
```suggestion
-   `settings` **[Object](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object)** the settings of the confirmation modal. If you do not provide any value for any of the fields, a default value is used
```
we usually don't (;)) use contractions - but it also depends on the type of doc. I would avoid them to be consistent.

docs/luigi-client-api.md
```diff
@@ -13,47 +13,57 @@
...
-# Luigi Client API
+# Luigi Client
+
+This document outlines the features provided by the Luigi Client API. It covers these topics:
+
+-   [Lifecycle](#lifecycle) - functions that define the lifecycle of different Luigi elements
+-   [Callbacks](#lifecycle˜initlistenercallback) - callback functions for initListener and customMessageListener
+-   [Link manager](#linkmanager) - you can use the linkManager instead of an internal router
+-   [Split view](#splitview) - allows you to open a micro frontend in the lower part of the content area in a "split screen" view
+-   [uxManager](#uxmanager) - functions related to appearance and user interface
```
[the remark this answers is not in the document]
alexandra-simeonova: Since Luigi is almost exclusively about user interface, I'd say the first one. Maybe we don't need the word "appearance" here. Thank you 👍

docs/luigi-core-api.md
alexandra-simeonova: Yes, I tried to add the dots but I get an error from the script which renders the docu 🙁

alexandra-simeonova: So, I removed the dots for consistency.

PR #1028, opened 2019-12-18
Add navigation functionality to core api

JohannesDoberer:


docs/luigi-core-api.md
```diff
@@ -377,6 +377,133 @@ Refreshes top navigation badge counters by rendering the navigation again.
...
+```
+
+#### openAsModal
+
+Opens a view in a modal. You can specify the modal's title and size. If you don't specify the title, it is the node label. If there is no node label, the title remains empty.  The default size of the modal is `l`, which means 80%. You can also use `m` (60%) and `s` (40%) to set the modal size. Optionally, use it in combination with any of the navigation functions.
+
+##### Parameters
+
+-   `path` **[string](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)** navigation path
+-   `modalSettings` **[Object](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object)?** opens a view in a modal. Use these settings to configure the modal's title and size
```
alexandra-simeonova:
```suggestion
-   `modalSettings` **[Object](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object)** opens a view in a modal. Use these settings to configure the modal's title and size
```

I don't think there should be a "?" after "Object" here. It doesn't seem to be in `navigation.js.` either. 🤔

PR #1034, opened 2020-01-07
Add valid href to navigation links

maxmarkus:
Added configurable option to show href links in the navigation.


ndricimrr:
I seem to be getting a weird behaviour with the ContextSwitcher. The href shows as 'undefined' for Environment 1,2,3 until 9. They work fine if opened in the same tab but not in new ones.
Looks like this for me:
`<a href="/undefined" class="fd-menu__item is-selected svelte-1ldh2pm" title="Environment 1">Environment 1</a>` 
Other than that all works fine

PR #1052, opened 2020-01-15
Store update should remove cache objects from nodes

maxmarkus:
navigation.nodes should not contain cache objects after receiving an update, eg. through `Luigi.configChanged('navigation')`.


maxmarkus:
Due to side effects and inability to do a simple fix, we instead document the best practices in #1062

PR #1076, opened 2020-01-31
Add splitview functionality

ndricimrr:
**Description**

- The function `openAsSplitview` is implemented.
- In addition the `navigate` function does support splitview as well.

**Status Update:**
Ready for review

**Related issue(s)**


core/src/core-api/splitViewHandle.js
```diff
@@ -0,0 +1,178 @@
...
+import { LuigiCoreAPIBase } from './baseClass';
+
+/**
+ * Split view 
+  Allows to open a micro frontend in a split screen in the lower part of the content area. Open it by calling `const splitViewHandle = Luigi.navigate().openAsSplitView`. 
+  At a given time, you can open only one split view. It closes automatically when you navigate to a different route.
+  When you call `handle.collapse()`, the split view gets destroyed. It recreates when you use `handle.expand()`.
+  `openAsSplitView` returns an instance of the split view handle. The functions, actions, and event handlers listed below allow you to control and manage the split view.
+  * @name splitView
+  * @since 0.6.0
```
maxmarkus: Please update all `@since` occurrences, because it will get into care api earliest with `0.7.4`. 

Just thinking about this, since the implementer does not know when it gets merged/released, any defined number might be wrong. We should add something like `* @since NEXTRELEASE`  .. and releasemeister will update it with the latest release, once it gets released.

core/src/main.js
```diff
@@ -70,6 +70,34 @@ const configReadyCallback = () => {
         return app.$$.ctx.showModal(settings);
       };
 
+      Luigi.openAsSplitView = (path , settings)  => {
```
maxmarkus: Would it make sense to group it a bit?  like 
```
Luigi.splitView = {
  openAsSplitView: (path , settings) => app.$$.ctx.openSplitView(path, settings),
  closeSplitView: ...
}
```
So its not so cluttered. And you could skip the brackets, so its one line per command.

core/src/core-api/splitViewHandle.js
ndricimrr: I've updated `@since` occurrences on the new functions I've added according to your suggestion. This comment also made me realise that this particular file was redundant and I had forgotten to delete it :) . (which also fixed unit test coverage issue)

core/src/main.js
ndricimrr: nice suggestion, thanks

core/src/core-api/navigation.js
```diff
@@ -49,6 +53,88 @@ class LuigiNavigationManager {
...
+   * @since NEXTRELEASE
+   * @example
+   * Luigi.navigation().closeSplitView();
+   */
+  closeSplitView() {
+    return Luigi.splitView.closeSplitView();
+  }
+
+  /**
+   * Collapse split view opened with openAsSplitView(path, splitViewSettings)
```
alexandra-simeonova:
```suggestion
   * Collapses split view opened with openAsSplitView(path, splitViewSettings)
```

For consistency with the other descriptions

core/src/core-api/navigation.js
```diff
@@ -49,6 +53,88 @@ class LuigiNavigationManager {
...
+   * @since NEXTRELEASE
+   * @example
+   * Luigi.navigation().expandSplitView();
+   */
+  expandSplitView() {
+    return Luigi.splitView.expandSplitView();
+  }
+
+  /**
+   * Returns true if splitview is collapsed
```
alexandra-simeonova:
```suggestion
   * Returns `true` if split view is collapsed
```

core/src/core-api/navigation.js
```diff
@@ -49,6 +53,88 @@ class LuigiNavigationManager {
...
+   * @since NEXTRELEASE
+   * @example
+   * Luigi.navigation().isSplitViewCollapsed();
+   */
+  isSplitViewCollapsed() {
+    return Luigi.splitView.isSplitViewCollapsed();
+  }
+
+  /**
+   * Returns true if splitview is expanded
```
alexandra-simeonova:
```suggestion
   * Returns `true` if split view is expanded
```

core/src/core-api/navigation.js
```diff
@@ -49,6 +53,88 @@ class LuigiNavigationManager {
...
+   * @since NEXTRELEASE
+   * @example
+   * Luigi.navigation().isSplitViewExpanded();
+   */
+  isSplitViewExpanded() {
+    return Luigi.splitView.isSplitViewExpanded();
+  }
+
+  /**
+   * Returns true if splitview is closed
```
alexandra-simeonova:
```suggestion
   * Returns `true` if split view is closed
```

core/src/core-api/navigation.js
```diff
@@ -49,6 +58,26 @@ class LuigiNavigationManager {
...
+  /**
+   * Opens a view in a split view. You can specify the split view's title and size. If you don't specify the title, it is the node label. If there is no node label, the title remains empty. The default size of the split view is 40, which means 40% height of the split view.
+   * @memberof LuigiNavigation
+   * @param {string} path navigation path
+   * @param {Object} splitViewSettings opens a view in a split view. Use these settings to configure the split view's behaviour
+   * @param {string} splitViewSettings.title split view title. By default, it is the node label. If there is no label, it is left empty
+   * @param {number} [splitViewSettings.size=40] height of the split view in percent
+   * @param {boolean} [splitViewSettings.collapsed=false] opens split view in collapsed state
+   * @returns {Object} an instance of the SplitView. It provides functions to control its behavior.
+   * @see {@link https://docs.luigi-project.io/docs/luigi-client-api?section=splitview|SplitView Client} for further documentation. These methods are implemented for Luigi Core: `close`, `collapse`, `expand`, `isCollapsed`, `isExpanded`, `exists`
```
alexandra-simeonova:
```suggestion
   * @see {@link https://docs.luigi-project.io/docs/luigi-client-api?section=splitview|SplitView Client} for further documentation. These methods from the Client SplitView are also implemented for Luigi Core: `close`, `collapse`, `expand`, `isCollapsed`, `isExpanded`, `exists`
```

alexandra-simeonova: Not a big deal, but it might be good to repeat this to avoid confusion 🤔

PR #1083, opened 2020-02-04
Open profile items in a modal window

zarkosimic:
- Adds the possibility to open profile items in a modal window.

Steps to reproduce: 
Go to `core/examples/luigi-sample-angular/src/luigi-config/extended/navigation.js` find **settings** node and replace with this one `{
      pathSegment: 'settings',
      hideFromNav: true,
      hideSideNav: true,
      label: 'Settings',
      viewUrl: '/sampleapp.html#/settings'
    }`.
In the same file, find **getProfileItems** and in **items** array add this node `{
        label: "Settings",
        icon: "action-settings",
        link: "/settings",
        openNodeInModal: {
          title: "Settings M",
          size: "m"
        }
      }`.

Before changes: 

image

After changes: 

image


core/src/Authorization.html
```diff
@@ -351,7 +352,15 @@
 
   export function onActionClick(item) {
     getUnsavedChangesModalPromise().then(() => {
-      Routing.navigateToLink(item);
+      if (
+        typeof item.openNodeInModal === 'object' &&
+        item.openNodeInModal !== null
+      ) {
+        const route = RoutingHelpers.buildRoute(item, `/${item.link}`);
```
JohannesDoberer: `route` is in my case `//settings`. I would remove the slash.

PR #1096, opened 2020-02-11
Document how to set auth storage type

### LINKED ISSUE #1092: document how to set auth storage type
There is documentation how to get the auth storage type https://docs.luigi-project.io/docs/luigi-core-api?section=authorization but none how to set it ;)

should be here: https://docs.luigi-project.io/docs/authorization-configuration?section=general-authorization-options

alexandra-simeonova:
**Description**

- documents how to set auth storage type 

**Related issue(s)**
Fixes #1092


docs/authorization-configuration.md
```diff
@@ -43,7 +43,11 @@ auth: {
 }
 ```
 
-The **use** key defines the active authorization provider and the **disableAutoLogin** key allows you to disable the automatic login flow that is provided by default.
+- **use** defines the active authorization provider.
+- **disableAutoLogin** allows you to disable the automatic login flow that is provided by default.
+
+Optionally, you can also add:
+- **auth.storage** parameter which allows you to set the storage type.
```
maxmarkus: should only be _storage_
```suggestion
- **storage** parameter which allows you to set the storage type.
```

PR #1103, opened 2020-02-14
Update version of Fundamental Library Styles

marynaKhromova:
- update Fundamental Library Styles
- adjust tabs
- adjust app switcher and contextSwitcher
- fix e2e tests
- remove (where it's possible) custom styles with --sap variables


core/src/Alerts.html
```diff
@@ -117,19 +117,14 @@
...
-      border-bottom-style: solid;
-      border-bottom-width: 1px;
-      border-bottom-color: #0a6ed1;
-      border-bottom-color: var(--sapLinkColor, #0a6ed1);
       -webkit-transition: all 125ms ease-in;
       transition: all 125ms ease-in;
       text-decoration: none;
 
       &:hover {
+        text-decoration: underline;
```
ndricimrr: The alert text links didn't seem to become underlined on hover.  
Attaching a picture here. Tried other browsers as well.
hover

marynaKhromova: thanks, I've pushed some changes for it.

PR #1129, opened 2020-03-04
OIDC profile interceptor

### LINKED ISSUE #1125: Add interceptor for OIDC profile storage
If an oidc server returns a `profile` object in the jwt, we are storing it. Due to security constraints it might be necessary to remove certain values like email from the profile. 

It should get implemented OIDC specific and available as `auth.openIdConnect.profileStorageInterceptor`. 

Example usage:
```
profileStorageInterceptor: (profile) => {
  profile.email = undefined;
  return profile;
}
```

maxmarkus:
Fixes #1125

To verify, set the `auth.use` to `openIdConnect` and start the mock server as mentioned in the config https://github.com/SAP/luigi/pull/1129/files#diff-2bd534401de4dfa73c5f410a930c8a79R40


docs/authorization-configuration.md
```diff
@@ -97,6 +98,7 @@ auth: {
 - **accessTokenExpiringNotificationTime** is the number of seconds before an access token expires and triggers silent token refresh. The default value is `60` seconds.
 - **thirdPartyCookiesScriptLocation** is the URL to the page containing third-party cookies support check. For details, see [Third-party cookies and silent token refresh section](#Third-party-cookies-and-silent-token-refresh).
 - **userInfoFn** provides a function to get user information. It returns a promise of a **userinfo** object which can contain **name**, **email** and **picture** (value is a URL to the image). **Name** or **email** are displayed in the profile drop-down menu and the user’s profile picture is displayed in the top navigation.
+- **profileStorageInterceptorFn** provides a function to mutate the profile values of the [JSON Web Token (JWT)](https://en.wikipedia.org/wiki/JSON_Web_Token) before it gets stored in Luigi. It allows the removal of values like **email** to comply with data privacy restrictions. Since it is async, you could use it to enrich the profile data, but it should not get mixed up with **userInfoFn** which is specifically designed to define user information.
```
alexandra-simeonova:
```suggestion
- **profileStorageInterceptorFn** provides a function to mutate the profile values of the [JSON Web Token (JWT)](https://tools.ietf.org/html/rfc7519) before it gets stored in Luigi. It allows the removal of values like **email** to comply with data privacy restrictions. Since it is async, you could use it to enrich the profile data, but it should not get mixed up with **userInfoFn** which is specifically designed to define user information.
```

The description looks good 🙂 One small point is I'm not sure if it's a good idea to use a Wikipedia link as it has a "bad reputation" since anyone can edit. At the same time there aren't many "official" sources to use... One I could find was: https://tools.ietf.org/html/rfc7519

PR #1130, opened 2020-03-05
Add hint to core functionality called early

ndricimrr:
**Description**
Certain Luigi functionality ( ex: auth() ) requires Luigi to be initialised, yet setConfig() might take some time causing such functionality to be called too early.

- Adds console warning when Luigi functionality gets called before the Luigi configuration is initialized.

In order to simulate the behaviour: 

Add the following line : 
`Luigi.auth().store.getAuthData()`

into  _luigi-sample-angular/src/luigi-config/extended/main.js_  right after `setConfig() `is called. 

Then wrap the setConfig call with the timeout fn: 
`setTimeout( ()=>{ setConfig(...); }, 0); ` 
in order to force setConfig() to lose the race. This is then going to produce a console warning

**Related issue(s)**


core/src/core-api/auth.js
```diff
@@ -66,6 +66,7 @@ class LuigiAuth {
    * @property {string} idToken - id token, used for renewing authentication
    */
   get store() {
+    if(!LuigiConfig.initialized)  console.warn("Luigi Core is not initialized yet. Consider moving your code to the luigiAfterInit lifecycle hook. Documentation: https://docs.luigi-project.io/docs/lifecycle-hooks");
```
maxmarkus: We prefer not to have those shorthands :-) and max line length should be taken into account. 
Something similar to:
```
    if(!LuigiConfig.initialized) {
      console.warn(`Luigi Core is not initialized yet. Consider moving your code to the luigiAfterInit lifecycle hook. 
      Documentation: https://docs.luigi-project.io/docs/lifecycle-hooks`);
    }
```

PR #1145, opened 2020-03-11
Route change loses context in contextswitcher

JohannesDoberer:


JohannesDoberer:
The function `getConfigValueFromObjectAsync` in **core/src/utilities/helpers/async-helpers.js** expects three params. In the function `getDefaultChildNode` in **core/src/utilities/helpers/routing-helpers.js**  the `getConfigValueFromObjectAsync` fn is called with 2 params (context is missing).

PR #1156, opened 2020-03-13
Fix auth layer runtime error

maxmarkus:
If auth is not defined, we can step out of auth layer init.


core/src/services/auth-layer.js
```diff
@@ -35,6 +31,10 @@ class AuthLayerSvcClass {
 
   async init() {
     const idpProviderName = LuigiConfig.getConfigValue('auth.use');
+    if (!idpProviderName) {
+      // No Authentication active
```
ndricimrr: Nice fix 👍  what do you say about adding a console log here in case it happens. or is it too redundant

PR #1158, opened 2020-03-13
Aligned splitview and centered arrow

ndricimrr:
**Description**

- Fix Splitview Alignment

**Related issue(s)**

### DIFF
--- a/core/src/SplitView.html
+++ b/core/src/SplitView.html
@@ -307,8 +307,9 @@
 
 <style type="text/scss">
   $topNavHeight: 44px;
-  $leftNavWidth: 320px;
+  $leftNavWidth: 15em;
   $collapsedSplitviewHeight: 38px;
+  $collapsedSplitViewContainerHeight: 49px;
   $desktopMinWidth: 600px;
   $colorNeutral4: #89919a;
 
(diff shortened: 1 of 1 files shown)


core/src/SplitView.html
```diff
@@ -307,8 +307,9 @@ <h1 class="fd-splitView__title">{splitViewSettings.title}</h1>
 
 <style type="text/scss">
   $topNavHeight: 44px;
-  $leftNavWidth: 320px;
+  $leftNavWidth: 15em;
   $collapsedSplitviewHeight: 38px;
+  $collapsedSplitViewContainerHeight: 49px;
```
marynaKhromova: I would recommend removing these changes with height and instead, just add `margin-top: 0;` to the `fd-splitView__title`

core/src/SplitView.html
```diff
@@ -329,7 +330,7 @@ <h1 class="fd-splitView__title">{splitViewSettings.title}</h1>
   }
 
   :global(.lui-collapsed.splitViewContainer) {
-    height: $collapsedSplitviewHeight;
+    height: $collapsedSplitViewContainerHeight;
```
marynaKhromova: I would recommend removing these changes, see the comment above.

PR #1188, opened 2020-03-23
Redesign angular example according to fundamental library styles

marynaKhromova:


marynaKhromova:
> some minor stuff regarding some panel paddings

@ndricimrr Thanks for your feedback, a good catch 👍 I've adjusted the parts you mentioned.

marynaKhromova:
This PR is closed, because of a new one https://github.com/SAP/luigi/pull/1207

PR #1213, opened 2020-03-27
Fix OIDC storage: none

### LINKED ISSUE #1182: OIDC storageType:none broken - provider should not redirect to root for hash removal
OIDC provider is removing the hash based access_token stuff by redirecting to / path.
consider using history.replace to avoid redirect, as it loses the stored data of the memory stored in window object.

maxmarkus:
Fixes #1182


maxmarkus:
Works as confirmed by @kwiatekus , thanks for reporting!

PR #1225, opened 2020-04-01
fix svelte example

kohlerm:
The svelte example does not work for me. 

- there are some issues with the webpack 
- Hello does not set a value for the message. Is this intended?
- Some dependencies missing  in the local package file?


core/examples/luigi-example-svelte/src/routes/Home.svelte
```diff
@@ -10,7 +10,7 @@
   import LuigiClient from '@kyma-project/luigi-client';
   import { onMount } from 'svelte';
 
-  export let message;
+  export let message='Hello world'
```
zarkosimic, line 13: **Note**: _This message would be only visible if micro frontend runs without Luigi._

PR #1270, opened 2020-04-16
Refactor Svelte example app

zarkosimic:
Fixes #1270 

- Fixed bug when build files
- Updated prod scripts
- Updated Luigi to v.1.0.1
- Updated Fundamental-styles to v.0.8.1
- Refactored


core/examples/luigi-example-svelte/package.json
```diff
@@ -10,27 +10,28 @@
...
   },
   "scripts": {
-    "build:prod": "webpack --mode=production --display-error-details",
+    "build:prod": "NODE_ENV=production webpack --display-error-details",
     "build:dev": "webpack",
-    "dev": "concurrently \"npm run build:dev -- --watch\" \"npm run serve:dev\" \"npm run serve:luigi\"",
-    "serve": "concurrently \"npm run serve:dev\" \"npm run serve:luigi\"",
-    "serve:dev": "sirv public --single --dev --port 5000",
-    "serve:luigi": "cd public && ws --spa mf.html --port 5001"
+    "dev": "concurrently \"npm run build:dev --watch\" \"npm run serve:luigi\" \"npm run serve:microfrontend\"",
```
maxmarkus, line 22: > [0] npm run build:dev --watch exited with code 0

--watch is not watching anymore, is there a reason for the change?
```suggestion
    "dev": "concurrently \"npm run build:dev -- --watch\" \"npm run serve:luigi\" \"npm run serve:microfrontend\"",
```

zarkosimic: Actually an idea was to remove `--watch` as there is no sense to watch file as we have no `hot reload` option running and in the case of changes we have to run `npm run dev` again.

Also, after changes when `--watch` is enabled, it was resulting in my case with 404 error when tried to reload page after changes. Sometimes I had to run `npm run dev` 2 times.

JohannesDoberer: got the same bahaviour

maxmarkus: I also have seen this, caused by Cleanup webpack plugin. I have removed it since its not ultimately required for the example app.

PR #1304, opened 2020-04-27
Add possibility to unload Luigi

### LINKED ISSUE #1126: Reinitialising Luigi  HTML container dynamically
**Description**

My Luigi container is part of an angular application with a login/logout process. On login, a user initialise luigi and has successfully access to the app. On logout it destroy the HTML container from the DOM to show the login form. 
If the user login a second time, we display the luigi container again but it no longer start. On page reload, luigi load properly.

I am looking for a way to reinitialise the HTML container, losing the previous instance would actually be acceptable but I believe Luigi should be able to reload.

**Expected result**

Luigi should appear on the second load.

**Actual result**

Luigi do not appear at all.

**Steps to reproduce**

You can access the [stackblitz angular-luigi-routing](https://stackblitz.com/edit/angular-luigi-routing) I just created which display a luigi container within two angular pages. 
My use case is exactly the one when switching to the login page then back to the luigi page using the navigation menu on top of the app.

**Troubleshooting**

Starting Luigi using `Luigi.setConfig`, on second load I could see the initialise value is set to true and so I guess skip the binding with the DOM. If it is false, it works.

maxmarkus:
Adds `Luigi.unload()` method to unload Luigi Core application, which then can be re-created with `Luigi.config(...)`. It also checks for `unload()` method in the active auth provider to unload logic which might break stuff (eg. setInterval).

Fixes #1126


client/src/helpers.js
```diff
@@ -22,8 +22,6 @@ class Helpers {
     }.bind(this);
 
     window.addEventListener('message', helperListener);
-    window.onunload = () =>
-      window.removeEventListener('message', helperListener);
```
maxmarkus, line 26: Moved to event-listener-helpers.js.

docs/luigi-core-api.md
```diff
@@ -155,6 +155,10 @@ Returns **[boolean](https://developer.mozilla.org/docs/Web/JavaScript/Reference/
 
 -   **deprecated**: now located in Luigi.auth() instead of Luigi
 
+#### unload
+
+Unloads the current Luigi instance, which can be initialized later again by using `Luigi.config().init()`
```
alexandra-simeonova, line 160:
```suggestion
Unloads the current Luigi instance, which can be initialized again later by using `Luigi.config().init()`.
```

PR #1306, opened 2020-04-28
Search component

### LINKED ISSUE #1259: Add search component
Add search component to the shellbar with open/close option.
It should be configurable and part of the root item of Luigi config like
`Luigi.setConfig({
  auth,
  navigation,
  routing,
  settings,
  communication,
  lifecycleHooks,
  globalSearch
});`

JohannesDoberer:
docSearch result needs to be styled if we want to have this.

Fixes #1259


test/e2e-test-application/e2e/tests/1-angular/navigation.spec.js
```diff
@@ -573,4 +573,60 @@ describe('Navigation', () => {
       });
     });
   });
+  describe('GlobalSearch', () => {
+    context('Desktop', () => {
+      it('GlobalSearch Desktop', () => {
+        cy.get('.fd-shellbar__action--desktop .luigi-search__input').should(
```
marynaKhromova, line 579: Maybe it's better to rely on custom classes instead of classes from Fundamentals (the same applies to  `globalSearch.js`).

PR #1309, opened 2020-04-29
Fix input validation in fiddle idpmock

### LINKED ISSUE #1290: Input validation in fiddle idpmock
in website\fiddle\public\auth\idpmock\logout.html

post_logout_redirect_uri can be used to execute javascript, e.g. https://fiddle.luigi-project.io/auth/idpmock/logout.html?post_logout_redirect_uri=javascript:alert(22)

redirectTo should be validated as follows:
*  Provide proper input validation
* do not accept javascript:
* do not redirect to arbitrary domains. (since this is only a sample for fiddle, restrict to fiddle.luigi-project.com domain)

Reference: SecVal 7103.16

zarkosimic:
Fixes #1290


website/fiddle/public/auth/idpmock/logout.html
```diff
@@ -9,10 +9,19 @@
 	Logging out...
 	<script>
 		window.onload = function() {
-    const redirectTo = decodeURIComponent(
+    let redirectTo = decodeURIComponent(
       window.location.href.match(/post_logout_redirect_uri=(.*?)(&|$)/)[1]
     );
-    window.location.href = redirectTo;
+    const disableJs = redirectTo.replace(/javascript:/gi, '');
```
maxmarkus, line 15: I am not sure if this is enough, since there are tons of possibilities how one could write `javascript` according to https://owasp.org/www-community/xss-filter-evasion-cheatsheet  .. But since we are checking if it starts with _window.location.origin_ I couldn't find a possibility to sneak anything in. Just thinking if we remove that check at all just not to raise wrong safety feelings.

JohannesDoberer: I would say it's fine.
I tested:
`http://localhost:8080/auth/idpmock/logout.html?post_logout_redirect_uri=javascript:alert(&quot;XSS&quot;)
http://localhost:8080/auth/idpmock/logout.html?post_logout_redirect_uri=JaVaScRiPt:alert('XSS')
http://localhost:8080/auth/idpmock/logout.html?post_logout_redirect_uri=javascript:alert('XSS');
http://localhost:8080/auth/idpmock/logout.html?post_logout_redirect_uri=javascript:alert("RSnake says, 'XSS'")`
All of them aren't working.

We can think if we apply the last point in the issue description and replace `window.location.origin` with the fiddle url but I think we can also do the implemented version.

PR #1344, opened 2020-05-18
Add pre commit hook

### LINKED ISSUE #1285: Add pre-commit hook to prevent committing of specific strings
It happened that luigi-client-api.md got pushed with a merge conflict. 
Add a pre-commit script that checks the code for merge commits, which are marked with `<<<<<<<<`, `>>>>>>>>` and `=========` (i did not check how many characters those string have).

Additionally it should check that `window.addEventListener` is not used in other files than EventListenerHelper class.

ndricimrr:
**Description**
Fixes #1285


scripts/hooks/prevent-illegal-characters.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
maxmarkus, line 1:
```suggestion
#!/usr/bin/env bash
```
Just a best practice. I started changing it in other scripts.

scripts/hooks/prevent-illegal-characters.sh
```diff
@@ -0,0 +1,43 @@
...
+    done
+}
+
+function check_addEventListener_wrong_usage() { 
+    # string containing number of changes already staged for commit
+    illegal_files=$(git diff  --cached --diff-filter=ACMR \
+                    -G 'window.addEventListener' --name-only  -- core \
+                    ':(exclude)scripts/hooks/prevent-illegal-characters.sh' \
+                    ':(exclude)core/src/utilities/helpers/event-listener-helpers.js')
+    echo $illegal_files
```
maxmarkus, line 32:
```suggestion
```
Is it a leftover?

ndricimrr: ah damn , yes 🤦

PR #1353, opened 2020-05-20
Refactor implicit.html 

### LINKED ISSUE #1348: Luigi Mock Identity Provider HTML/CSS refactor
- style according to Fundamental styles the Luigi Mock Identity Provider page (implicit.html)
- refactor HTML/CSS
- add mobile responsivness

UlianaMunich:
Fixes #1348 
- add fundamental styling to implicit.html and logout.html to make icons available (it was broken since FD moved it out)
- refactor and clean the HTML and CSS


website/fiddle/public/auth/idpmock/implicit.html
```diff
@@ -1,77 +1,56 @@
...
-<div class="panel">
-  <h1>Login to Luigi sample app</h1>
-  <form>
-    <input type="text" class="form-input" value="LuigiUsername"><br>
-    <input type="password" class="form-input" value="LuigiPassword"><br>
-    <button id="login-button">Login</button>
-  </form>
-</div>
-<script src="implicit.js">
+  <div class="login-form">
```
marynaKhromova, line 45: Just an idea. If we rework this page anyway and use here fundamental styles, wouldn't it be a good idea to use fundamental components (for example a panel) instead of our custom structure? Also, we could add the labels before inputs. Maybe it should be reworked in Luigi core first. What do you think?

Issue #1352, opened 2020-05-20
How can I style the Luigi core elements?


huberv:
**Description**

From reading the documentation I would say that the css styles are embedded in the Luigi core. This is fine as long as you are using one of the Fiori Fundamentals libaries, but what about using Luigi with e.g. Google's material design? Is there a way to style Luigi in a consistent and future-proof way?

**Reasons**

I guess Luigi shouldn't be tied to the SAP Fiori Fundamentals libraries / styles.

hardl:
Hi,
there are several options at the moment:
1) use fundamental styles theming capabilities, see also https://github.com/SAP/theming-base-content, you can already achieve a lot by customizing the css variables
2) manually overwrite the styles where needed, you can check our documentation app as an example (yes, it is a luigi app ;) )
3) turn off luigi view components completely and implement your own view components for header and navigation

Cheers

huberv:
Hi Philipp,

Thanks for the swift reply! It would be great to have options 1 and 3 in the official documentation.

hardl:
There will be comprehensive examples soon showing how it is done (https://github.com/SAP/luigi/issues/1361)
In general, you just define your custom html components as usual (you are owning index.html) and connect them to luigi via luigi core api

PR #1354, opened 2020-05-20
New FAQ - How can I style the Luigi core elements?

alexandra-simeonova:
**Description**

- Answer the FAQ about how to style Luigi with something different than Fundamental Styles

**Related issue(s)**
#1352


docs/faq.md
```diff
@@ -51,6 +51,13 @@ Development scalability was one of the main goals right from the beginning. Ther
 
 It is crucial that all micro frontends in a solution follow the same design guidelines. Luigi's default UI styling is based on [Fundamentals](https://sap.github.io/fundamental-styles/) but it can be customised. If you don’t want to use Fundamentals, but Bootstrap, Material, or something else instead, you need to re-style the Luigi view components according to your design guidelines or replace them with your own components completely.
 
+### I don't want to use the default Fiori Fundamentals style. How can I style Luigi differently?
+
+There are a few options to do that at the moment:
+- Use the Fundamental Styles theming capabilities which already allow you to achieve a lot by customizing the CSS variables. Find more info [here](https://github.com/SAP/theming-base-content).
```
UlianaMunich, line 57: I'd add CSS/LESS/SASS variables instead of CSS only. All those variables like colors, font sizes, padding, etc. are part of SAP Fiori 3, not Fundamental Styles. It was extracted after version 0.8.1. I'm not sure if we need to. mention it to not confuse the client but wanted to clarify here at least for us.

PR #1357, opened 2020-05-25
Open, close and clear globalsearch field

JohannesDoberer:


JohannesDoberer:
Good point. I decided to fix it in this PR

PR #1376, opened 2020-05-28
Fix left-nav flyout

### LINKED ISSUE #1293: left-nav flyout and navigation component using two different state variables
flyout live-checks local storage "luigi.preferences.navigation.collapsedNavigation" on every click, navigation component stores state internally and just reads it initially.
This can lead to invalid state if local storage is changed.

Steps to reproduce (scenario 1):
* expand left nav
* manually set luigi.preferences.navigation.collapsedNavigation to true
-> result: flyouts appear, even if left nav is expanded
 

Steps to reproduce (scenario 2):
* collapse left nav
* clear localStorage
-> result: flyouts will not appear on click

state for both components must be handled via a single variable, preferrably the internal on already used by nav component, localStorage should just be used to persist the value for setting this state initially after next page refresh

zarkosimic:
Fixes #1293

### DIFF
--- a/core/src/navigation/services/semi-collapsed-navigation.js
+++ b/core/src/navigation/services/semi-collapsed-navigation.js
@@ -18,9 +18,19 @@ class SemiCollapsibleNavigationClass {
       window.innerWidth !== 0 &&
       window.innerWidth < CSS_BREAKPOINTS.desktopMinWidth
     ) {
-      this.setCollapsed(true);
+      if (this.isSemiCollapsed === undefined) {
+        this.isSemiCollapsed = true;
+      } else {
+        this.isSemiCollapsed = this.getCollapsed();
+      }
+    } else {
+      if (this.isSemiCollapsed === undefined) {
+        this.isSemiCollapsed = false;
+      } else {
+        this.isSemiCollapsed = this.getCollapsed();
+      }
     }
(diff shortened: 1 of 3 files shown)


core/src/navigation/services/semi-collapsed-navigation.js
```diff
@@ -18,9 +18,19 @@ class SemiCollapsibleNavigationClass {
       window.innerWidth !== 0 &&
       window.innerWidth < CSS_BREAKPOINTS.desktopMinWidth
     ) {
-      this.setCollapsed(true);
+      if (this.isSemiCollapsed === undefined) {
```
ndricimrr, line 21: I find the if statements a bit cluttered here. If it would make sense for you as well I would suggest something like this: 

```
    let isSemiCollapsedUndefined = this.isSemiCollapsed === undefined
    if (
      this.semiCollapsible &&
      window.innerWidth !== 0 &&
      window.innerWidth < CSS_BREAKPOINTS.desktopMinWidth
    ) {
      this.isSemiCollapsed = isSemiCollapsedUndefined ? true : this.getCollapsed();
    } else {
      this.isSemiCollapsed = isSemiCollapsedUndefined ? false : this.getCollapsed();
    }
```

zarkosimic: Yes, make sense definitely. Thanks.

PR #1383, opened 2020-06-03
Set and get GobalSearch value

JohannesDoberer:


core/src/App.html
```diff
@@ -491,13 +491,32 @@
 
   export const openSearchField = () => {
     isSearchFieldVisible = true;
+    if (document.querySelector('.luigi-search__input')) {
+      document.querySelector('.luigi-search__input').focus();
+    }
   };
```
maxmarkus, line 497: Do you like such approach to avoid duplication of the query selector (simpler unit testing?), which you can reuse in all the functions or is it over-engineered..

```suggestion
const searchInputElem = () => document.querySelector('.luigi-search__input');

export const openSearchField = () => {
    isSearchFieldVisible = true;
    if (searchInputElem()) {
      searchInputElem().focus();
    }
  };
```

core/src/App.html
```diff
@@ -491,13 +491,32 @@
...
+    }
+  };
+
+  export const setGlobalSearchString = searchString => {
+    if (document.querySelector('.luigi-search__input')) {
+      document.querySelector('.luigi-search__input').value = searchString;
+      const triggerSearch = LuigiConfig.getConfigValue(
+        'globalSearch.searchProvider.onInput'
+      );
+      triggerSearch();
```
maxmarkus, line 516: I would suggest to check if triggerSearch is a function, else throw a warning.

core/src/core-api/globalsearch.js
```diff
@@ -64,13 +64,17 @@ class LuigiGlobalSearch {
    * @memberof LuigiGlobalSearch
    * @example Luigi.globalSearch().getSearchString()
    */
-  getSearchString() {}
+  getSearchString() {
+    return Luigi.getGlobalSearchString();
```
maxmarkus, line 68: Could it happen that this function is called to early? and should we do something about it?  .. I remember other functions that only work after Luigi init. But for search it should be obvious that you can only use it when Luigi is there. So could be fine like that.

JohannesDoberer: Yes, I think so too and should be fine. If you want to use Luigi global search feature  it should be obvious that Luigi is there.

PR #1401, opened 2020-06-09
update version of FD for core and E2E App to 0.10.0

### LINKED ISSUE #1380: Update version of FD 0.10.0 in Core and E2E
Breaking changes from FD: 
• Panel renamed to Layout Panel
• remove overflow from popover
• Tile migrated to Fiori 3 design
• remove Status Indicator
• Select component updates
• remove group list  component

For core removed the shadow box for fd-menu__list
Changed ht version in E2E App as well. No more changes required

https://github.com/SAP/fundamental-styles/releases/tag/0.10.0

UlianaMunich:
Fixes #1380


core/fundamentalStyleClasses.js
```diff
@@ -32,7 +32,6 @@ module.exports = [
   //'./node_modules/fundamental-styles/dist/layout-grid.css',
   './node_modules/fundamental-styles/dist/link.css',
   //'./node_modules/fundamental-styles/dist/list.css',
-  //'./node_modules/fundamental-styles/dist/list-group.css',
```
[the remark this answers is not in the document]
marynaKhromova: the list is still missing `avatar.css`and `numeric-content`

core/src/navigation/ContextSwitcherNav.html
```diff
@@ -105,15 +105,18 @@
...
-    border-bottom: 1px solid #eeeeef;
-    border-bottom-left-radius: 0;
-    border-bottom-right-radius: 0;
-  }
-  .fd-menu__list--bottom {
-    border-top: 1px solid #eeeeef;
-    border-top-left-radius: 0;
-    border-top-right-radius: 0;
+  .fd-menu__list {
+    box-shadow: none;
```
marynaKhromova, line 109: If I understand correctly, you could remove this custom style and apply the fundamental class `fd-menu__list--no-shadow`

core/src/navigation/LeftNav.html
```diff
@@ -62,33 +62,46 @@
...
+                tabindex="-1"
+                aria-controls="EX400L2"
+                aria-haspopup="true"
+                aria-expanded="{isExpanded(nodes, expandedCategories)}"
+                on:click|preventDefault="{() => setExpandedState(nodes, !isExpanded(nodes, expandedCategories), this)}"
+              ></a>
+            </div>
             <ul
               class="fd-nested-list fd-nested-list--text-only level-2"
+              id="EX400L2"
```
marynaKhromova, line 104: I have a question regarding `id` and` aria-controls`. For which reason do we have it here? If it's a mechanism to control open/close items, then it's redundant, since we have another mechanism to control it. If it's a part of WCAG, then I would say we need a dynamic `id` and` aria-controls`, otherwise we have the same values ​​on all expanded items.

core/src/navigation/LeftNav.html
```diff
@@ -200,16 +213,16 @@
                 class="fd-nested-list__icon {isSemiCollapsed ? 'sap-icon--rhombus-milestone-2' : ''}"
               ></span>
               {/if}
-              <span
-                class="fd-nested-list__title"
-              >{$getTranslation(node.label)}</span>
+              <span class="fd-nested-list__title">
+                <p>{$getTranslation(node.label)}</p>
```
marynaKhromova, line 217: I thought we decided to not use the tag `<p>` since the solution I found should work without extra custom structure.

core/src/navigation/TabNav.html
```diff
@@ -122,16 +124,28 @@
...
+                  aria-selected="{isSelectedCat(key,selectedNodeForTabNav)}"
+                  on:click|preventDefault="{() => toggleDropdownState(key+index)}"
+                >
+                  <span class="fd-nested-list__title">{$getTranslation(key)}</span>
+                </a>
+                <a
+                  class="fd-nested-list__expand-icon"
+                  href="#"
+                  tabindex="-1"
+                  aria-controls="EX100L2"
```
marynaKhromova, line 143: the same question as for `core/src/navigation/LeftNav.html`. If you need `aria-controls` here, then it should be connected with some `id`. Otherwise, it doesn't make sense to use it here.

JohannesDoberer:
ie11
I found a small issue in ie11 if you have semicollapsed mode the arrow down is not at the right place.

JohannesDoberer:
Good Job! 👍

PR #1428, opened 2020-06-22
Use Svelte Components

ndricimrr:
**Description**

- Added ability to add svelte components in markdown as comments. 

**How to test:**
1. Create a svelte component say: `my-component.svelte` in a new file in `docs/components`.
```
// my-component.svelte
<script>
  let name = 'world';
  let color = Math.floor(Math.random() * 100) < 50  ? 'red' : 'blue';
</script>

<h1 style='color:{color}' >Hello {name.toUpperCase()}!</h1> 
```

2. Add the following comment into a **docs/*.md** file, _svelte:< filename >_ :
```
<!-- svelte:my-component.svelte -->
```
3. Run docu website and check the corresponding changed **.md** file in the documentation page

-------

Alternatively you can simulate importing svelte components by adding import statements in the parent file as follows:
```
<!-- svelte:my-component.svelte -->
```

```
// my-component.svelte
<script>
import Header from './Header.svelte';
</script>
<Header></Header>
````

```
// Header.svelte 
<script>
  let name = 'world';
  let color = Math.floor(Math.random() * 100) < 50  ? 'red' : 'blue';
</script>

Hello inside <h1 style='color:{color}' >imported component {name.toUpperCase()}!</h1> 
```

**Note:** Currently css styling and html are separated when rendering the component. That being said the css styling currently cannot be written on the fly so you can resolve to using inline styling.

### CODE REVIEW

File: website/docs/src/unified-plugins/rehype-add-svelte-component.js [hunk 1]
```diff
@@ -0,0 +1,53 @@
+import visit from 'unist-util-visit';
+import requireFromString from 'require-from-string';
+import unified from 'unified'
+import parse from 'rehype-parse'
+const svelte = require('svelte/compiler');
```
maxmarkus, line 5:
```suggestion
import * as svelte from 'svelte/compiler';
```

ndricimrr:
currently svelte's compiler is not intended for our use case, as the components are not rendered the way we want them to. Also as this task is not high priority it is not worth investing more time into it.

PR #1430, opened 2020-06-22
Add promise cache and cache context-switcher labels

### LINKED ISSUE #1405: contextSwitcherLabelResolver gets called multiple times for the same value
Due to the fact our template is using `contextSwitcherLabelResolver` directly, that function gets called multiple times. If a customer is doing some rest call in the resolver, it appears as three calls for each entry.

Implement a caching mechanism to call the resolver just once but reuse the initial promise.

I did a simple approach for a customer's poc on config side, might help with the solution:

```javascript
const contextSwitcherLabels = new Map();
const contextSwitcherPromises = {}; // object of id arrays, {id[]}
function contextSwitcherLabelResolver(id) {
  if (contextSwitcherLabels.has(id)) {
    return contextSwitcherLabels.get(id);
  }

  var auth = JSON.parse(getAuthData());
  var token = auth.idToken;

  // cache resolver, since it gets called multiple times due to Luigi internals
  let csPromise;
  if (contextSwitcherPromises[id]) {
    // create new entry in promise stack
    csPromise = new Promise((resolve) => {
      contextSwitcherPromises[id].push(resolve);
    });
  } else {
    // create promise with initial rest call
    contextSwitcherPromises[id] = [];
    csPromise = new Promise((resolve) => {
      contextSwitcherPromises[id].push(resolve);
    });

    fetch("https://>graphql-server>/query", {
      method: "POST",
      body: JSON.stringify({query: `{ project(projectId: \"${id}\") { projectId name }  }`}),
      headers: { Authorization: `Bearer ${token}`, "Content-Type": 'application/json' }}
    ).then((res) => {
      const data = res.json();
      // resolve all promises
      contextSwitcherPromises[id].forEach(resolve => resolve(data));
      delete contextSwitcherPromises[id];
    });
  }

  return csPromise.then((data) => {
    if (!data.data) {
      console.warn('ContextSwitcher project query failed for', id);
      contextSwitcherLabels.set(id, id);
      return id;
    }
    contextSwitcherLabels.set(id, data.data.project.name);
    return data.data.project.name
  });
}
```

maxmarkus:
Fixes #1405

- adds a cache map for context switcher fallback labels
- adds a promise cache service which allows to cache any async function to avoid duplicate calls to the backend. is also used for context-switcher fallback labels.

To test with a promise:
`test/e2e-test-application/src/luigi-config/extended/navigation.js:266`
```javascript
    fallbackLabelResolver: (id) => {
      return new Promise((resolve) => {
        setTimeout(() => {
          console.log('fetch mock resolving for', id);
          resolve(id.replace(/\b\w/g, l => l.toUpperCase()))
        }, 300)
      })
    }
```


JohannesDoberer:
I'm getting the same failing test which Jimi describes.

maxmarkus:
This adds to much complexity, caching of promises not coming from same source is hard and even harder to do right.
Opening a new PR for just the label cache.

PR #1451, opened 2020-07-01
Test external micro frontend

### LINKED ISSUE #1438: Introduce a MF running on different domain to the e2e-test-app and test it
Create a new MF which is running on a different domain and integrate it to the e2e-test-application.
It should be possible to start up a second server with the MF when the e2e-test-app will be started.

Write a test if Luigi-client on the external MF can talk to Luigi core app.
E.g. using LinkManager, custom messages, loading Indicator

JohannesDoberer:
Resolves #1438


test/e2e-test-application/externalMf/index.html
```diff
@@ -0,0 +1,44 @@
...
+    <title>External micro frontend</title>
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1, maximum-scale=1, minimum-scale=1"
+    />
+    <script src="./node_modules/@luigi-project/client/luigi-client.js"></script>
+    <link
+      rel="stylesheet"
+      href="./node_modules/@luigi-project/core/luigi.css"
+    />
```
ndricimrr, line 13:
```suggestion
```

I suppose this is a leftover

test/e2e-test-application/e2e/test2/1-angular/external-micro-frontend.spec.js
```diff
@@ -0,0 +1,16 @@
...
+  beforeEach(() => {
+    cy.visitLoggedIn('/');
+  });
+  describe('Normal navigating', () => {
+    it('Check if external micro frontend is available and an alert will be shown.', () => {
+      cy.visit('/projects/pr1/externalmf');
+      cy.wait(1000);
+      cy.get('[data-testid=luigi-alert]').should(
+        'have.class',
+        'fd-message-strip--success'
```
ndricimrr, line 12: The test seemed to fail for me, probably checking for an old type of fundamentals class name, I replaced this with `fd-alert--success` and it found it

test/e2e-test-application/externalMf/index.html
ndricimrr, line 13:
```suggestion
```
this seems like small leftover

test/e2e-test-application/e2e/test2/1-angular/external-micro-frontend.spec.js
maxmarkus: Tests were green for me. Did you run `lerna build` before running the tests?

test/e2e-test-application/externalMf/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "luigi-external-mf",
+  "version": "0.1.0",
+  "description": "",
+  "main": "index.html",
+  "license": "ISC",
```
maxmarkus, line 6: Guess it should also be MIT, since everything is mit.
```suggestion
  "license": "MIT",
```

test/e2e-test-application/e2e/test2/1-angular/external-micro-frontend.spec.js
ndricimrr: I had actually. Seems like I had forgotten to `symlink` 😅 . Works fine for me as well 🟢

UlianaMunich, line 12: I'm always a bit concerned about using the classes from FD, so that it changes all the time quite often. what about using "data-testid" attribute instead?

JohannesDoberer: I know what you mean, but in this case we cannot test against data-testid, because I'm using `cy.get('[data-testid=luigi-alert]')`already. The only thing what we can do is to add a luigi custom class, but I'm not sure if this is a good idea to add a class only for testing.

UlianaMunich: Ok, will pay attention if the class .fd-message-strip--success will change. Not an issue at all.

PR #1453, opened 2020-07-01
Fix logout label

### LINKED ISSUE #1448: Add default logout label
In case `profile.logout` is not defined, we are using _Sign out_ as default label. It used to be there and was breaking change while implementing static profile nav.

maxmarkus:
Fixes #1448


core/src/Authorization.html
```diff
@@ -169,17 +169,18 @@
...
+            GenericHelpers.isFunction(logoutItem.customLogoutFn)
+          ) {
             // TODO: PROFNAVLOGOUT: three smiliar implementations.
             // TODO: whats the difference between this profileLogoutFn, profileNav, auth-layer
-            if (profileLogoutfn && GenericHelpers.isFunction(profileLogoutfn)) {
-              profileLogoutFnDefined = true;
-              AuthLayerSvc.setProfileLogoutFn(profileLogoutfn);
-            }
+            profileLogoutFnDefined = true;
+            AuthLayerSvc.setProfileLogoutFn(profileLogoutfn);
```
JohannesDoberer, line 183: I think `setProfileLogoutFn` should get `logoutItem.customLogoutFn` as parameter otherwise profileLogoutfn is not defined. See removed line 176

website/docs/src/client-js/internal-links.js
```diff
@@ -32,8 +32,12 @@ export class InternalLinksHandler {
           console.debug('navigateInternal URL parse error', elem, elem.getAttribute('href'), error);
```
UlianaMunich, line 32: I know you haven't touched this line but maybe we don't need it there anymore to clean a bit around?

maxmarkus: This is intentional in case the href could not get read, so we get some output if someone says link xy does not work. Its inside the error block of try/catch and should't get called anyways.

PR #1478, opened 2020-07-09
OIDC: Support for Authorization Code flow with PKCE

azriel46d:
**Description**
Currently the plugin is expecting  `access_token` to be part of the `location.hash` . So when the settings are set to authenticate using [PKCE](https://oauth.net/2/pkce/) it causes an infinite redirect-login.
 The [PKCE flow](https://cdn2.auth0.com/docs/media/articles/flows/concepts/auth-sequence-auth-code-pkce.png) changes in that the callback returns a `code` as part of the query string. 

- This change addresses both PKCE and  token based on what the `response_type` is set to. If `token` is set, then an access_token is expected as part of the `hash` , otherwise if `code` is set, it checks for it in the query string.


test/e2e-test-application/src/luigi-config/extended/auth.js
```diff
@@ -35,12 +35,14 @@ class Auth {
...
     scope: 'openid profile email',
     // optional parameters
     // redirect_uri: '',
     // post_logout_redirect_uri: '/logout.html',
     // automaticSilentRenew: true,
+    // response_type: "code", for PKCE
+    // response_mode: "fragment", // change between `query` and `fragment`
     profileStorageInterceptorFn: profile => {
       profile.email = undefined;
       return profile;
```
maxmarkus, line 48:
```suggestion
    scope: 'openid profile email',
    // optional parameters
    // redirect_uri: '',
    // post_logout_redirect_uri: '/logout.html',
    // options settings
    // automaticSilentRenew: true,

    // for PKCE flow
    client_id: 'authorisation-code-pkce-mock-client', // oidc-mockserver client id
    response_type: "code", for PKCE
    response_mode: "fragment", // change between `query` and `fragment`

    // for implicit grant flow
    // client_id: 'implicit-mock-client', // oidc-mockserver client id

    profileStorageInterceptorFn: profile => {
      profile.email = undefined;
      return profile;
```

test/e2e-test-application/src/luigi-config/extended/auth.js
```diff
@@ -35,12 +35,21 @@ class Auth {
...
     scope: 'openid profile email',
     // optional parameters
     // redirect_uri: '',
     // post_logout_redirect_uri: '/logout.html',
+    // options settings
     // automaticSilentRenew: true,
+
+    // for PKCE flow
+    client_id: 'authorisation-code-pkce-mock-client', // oidc-mockserver client id
+    response_type: "code", for PKCE
```
maxmarkus, line 47:
```suggestion
    response_type: "code", // for PKCE
```
Sorry, missed that one.

PR #1483, opened 2020-07-10
test 2 reviewers check for global search

ndricimrr:
**Description**
the globalsearch feature PRs, seem to be skipping github status checks. Also it is still requesting review from old codeowners list. Marking the PR with 'request changes' doesn't block the merging as well. this PR is just for showcasing the situation. linked issues also won't show. could be that we need to modify some configs, or could be a temporary github issue


maxmarkus:
invalid. its a feature branch were it not being enforced.

PR #1484, opened 2020-07-13
Close search result

JohannesDoberer:


test/e2e-test-application/src/luigi-config/extended/globalSearch.js
```diff
@@ -36,8 +36,8 @@ class GlobalSearch {
           pathObject: {
             link: '/settings'
           },
-          label: 'Settings ohne params',
-          description: 'settings',
+          label: 'Settings ohne paramsasadfasdfklasdlfjkasdf',
+          description: 'settingssdfklasjdflsdjfk',
```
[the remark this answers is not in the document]
JohannesDoberer: `only if I click Enter I see a dropdown with search results` -> That is true. This behaviour is implemented in the globalSearch config file. Usually the customer has to implement the behaviour by his own. Maybe some customer want to trigger the search and show the search result after the first typed character, others want to trigger the search and show the search result after pressing enter. 
It is just a example implementation and I changed it on this brach (for reasons which I don't know :) ), but I can change it back.

Regarding your label and description hint.. I absolutely agree.

core/src/navigation/TopNav.html
```diff
@@ -581,6 +582,10 @@
     });
   }
 
+  export function closeSearchResult() {
+    dispatch('closeSearchResult');
+  }
+
```
stanleychh, line 588: Just curious, this `closeSearchResult` function is same as the one at `GlobalSearch.html`. Is there any possibility to extract it out? and import it separately.

JohannesDoberer: Good catch! It's a left over, I remove it :)

PR #1500, opened 2020-07-17
Theming Core API

### LINKED ISSUE #1375: Theming core API
Core API should have a subsection "theming" with the following functionality:
* getAvailableThemes() - returns a list of theme objects defined in luigi config (under theming) (should be retrievable async)
* setCurrentTheme(themeObject) - stores a theme object within luigi app context
* getCurrentTheme() - return the currently set theme from luigi app context, or default (as set in luigi config)
* isThemingAvailable() - true, if theming is set in luigi config

Note: Theme object should be an arbitrary object that can be defined by the developer, since we don't know required information for all theming solutions

maxmarkus:
Resolves #1375


core/src/core-api/theming.js
```diff
@@ -0,0 +1,106 @@
...
+   */
+  async getAvailableThemes() {
+    return await LuigiConfig.getConfigValueAsync('settings.theming.themes');
+  }
+
+  /**
+   * Sets the current theming object
+   * @memberof Theming
+   * @param {Object} a theming object, has the same structure as retrieved from **getAvailableThemes**
+   * @since NEXTRELEASE
```
UlianaMunich, line 33: Since NEXTRELEASE doesn't work for now could you please insert the version of Luigi manually everywhere?

core/src/core-api/index.js
```diff
@@ -13,6 +14,7 @@ export const LuigiNavigation = navigation;
 export const LuigiI18N = i18n;
 export const LuigiCustomMessages = customMessages;
 export const LuigiUX = ux;
+export const LuigiTheming = theming;
```
UlianaMunich, line 17: Theming is a very broad acronym. If this is only related to color (no fonts, sizes, icons) then what about renaming it to LuigiColorTeming?

core/test/core-api/theming.spec.js
```diff
@@ -0,0 +1,65 @@
...
+  const getMockTheming = () => {
+    return {
+      themes: [
+        { id: 1, name: 'light' },
+        { id: 2, name: 'dark' },
+        { id: 3, name: 'blue' },
+        { id: 4, name: 'green' },
+        { id: 5, name: 'red' },
+        { id: 6, name: 'orange' },
+        { id: 7, name: 'yellow' }
```
UlianaMunich, line 19: If colors are just labels and nothing more, maybe we should already provide the meaning of those labels? e.g. "yellow" should apply yellow colors to the core part of the Luigi? Orange and red almost the same, we could remove it. Or let's catch up on this once you are back @maxmarkus

core/src/core-api/theming.js
maxmarkus: No, just because it does not work automatically, we still just manually search-and-replace all NEXTRELEASE with the release. I have no clue at with version it will be released, so it would be wrong to add something there.

core/test/core-api/theming.spec.js
maxmarkus: Its a spec file, so its a test and not an example. Guess its ok for that.

JohannesDoberer:
If there is no theming defined in settings of luigi config and you'll try to get the current theming you will get an error.
`theming.js?eca5:81 Uncaught TypeError: Cannot read property 'defaultTheme' of undefined
    at theming_LuigiTheming.getCurrentTheme (theming.js?eca5:81)
    at <anonymous>:1:17
`
Bildschirmfoto 2020-07-27 um 12 21 39

PR #1507, opened 2020-07-23
Reverse blog entry order

### LINKED ISSUE #1501: Reverse display order for blog entries
newest entries should be on top

maxmarkus:
Adds sort + reverse to [`blogprocessor.js:64`](https://github.com/SAP/luigi/pull/1507/files#diff-cf67bb630f7fb9b969606cfdf9805abdR64-R65)
Removes and ignores generated blog files from website dev + public folders, since we generate them now on netlify.

Resolves #1501


website/landingpage/dev/src/services/blogprocessor.js
```diff
@@ -42,7 +42,7 @@ const getAuthors = (authors) => {
 
 const generateBlogEntry = (blog, content, showButton = false) => {
   // console.log(blog, content)
```
stanleychh, line 44: could we please remove the console even it was commented out?

maxmarkus, line 44:
```suggestion
```

PR #1522, opened 2020-07-29
Docu: make clear that setDirtyStatus is not restricted or tied to forms

### LINKED ISSUE #1328: Docu: make clear that setDirtyState is not restricted or tied to forms
Add some more infos to setDirtyState to make it more clear that it can be used also without forms, just to prevent navigation away from the current view.

alexandra-simeonova:
**Description**

-Add some more infos to setDirtyStatus to make it more clear that it can be used also without forms, just to prevent navigation away from the current view.

**Related issue(s)**
Resolves #1328

### DIFF
--- a/client/src/uxManager.js
+++ b/client/src/uxManager.js
@@ -49,7 +49,7 @@ class UxManager extends LuigiClientBase {
     helpers.sendPostMessageToLuigiCore({ msg: 'luigi.remove-backdrop' });
   }
   /**
-   * This method informs the main application that there are unsaved changes in the current view in the iframe. It can be used to prevent navigation away from the current view. For example, it can be used with form fields which were edited but not submitted, although the functionality is not restricted to forms.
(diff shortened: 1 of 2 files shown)


client/src/uxManager.js
```diff
@@ -49,7 +49,7 @@ class UxManager extends LuigiClientBase {
     helpers.sendPostMessageToLuigiCore({ msg: 'luigi.remove-backdrop' });
   }
   /**
-   * This method informs the main application that there are unsaved changes in the current view in the iframe. For example, that can be a view with form fields which were edited but not submitted.
+   * This method informs the main application that there are unsaved changes in the current view in the iframe. It can be used to prevent navigation away from the current view. For example, it can be used with form fields which were edited but not submitted, although the functionality is not restricted to forms.
```
UlianaMunich, line 52: @alexandra-simeonova I was reading it the first time and I feel like the 2nd occurrence of the phrase "it can be used" can be removed since the context is clear enough. What do you think?

PR #1525, opened 2020-07-30
Theming: Pass to clients

### LINKED ISSUE #1377: Pass theming information to luigi clients
Clients need to know which theme is selected.

This should be done in two ways:
* via context, so it can be consumed in client within init and update listeners, similar to "locale"
* via url parameters, so we need a generic way to manipulate the viewUrl of any navigation node before the iframe is created. This should be easily configurable in luigi theming config section, something like

```
...
theming: {
   ...
   nodeViewURLDecorator: {
      queryStringParameter: {
          keyName: 'sap-theme',
          // value could be optional if no mutation of the value necessary
          value: (themeId) => {
                  return themeId;
              }
      }
   }
}
```

The above example must result in a actual iframe src values like follows:

| original viewUrl | resulting url |
| ----| ----|
|https://xyz.com/mf.html|https://xyz.com/mf.html?sap-theme=someid|
|https://xyz.com/?someParam=true|https://xyz.com/?someParam=true&sap-theme=someid|
|https://xyz.com/#/hash/route|https://xyz.com/?sap-theme=someid#/hash/route|

For MVP, only query string will be supported, later we can add hash and custom decorator function

maxmarkus:
- adds custom query parameter in `settings.theming`
- adds `uxManager().getCurrentTheme()` 

Resolves #1377

How to verify the implementations: 

**Client:**
add to `overview.component.ts`:
```
onInit()
    addInitListener(() => {
      this.theme = uxManager().getCurrentTheme();
    })
```

**Core/queryParam:**
uncomment *nodeViewURLDecorator* in e2e application `settings.js` or check the new e2e tests.


core/src/utilities/helpers/iframe-helpers.js
```diff
@@ -251,7 +252,7 @@ class IframeHelpersClass {
       : luigiDefaultSandboxRules;
 
     const iframe = document.createElement('iframe');
-    iframe.src = viewUrl;
+    iframe.src = ViewUrlDecorator.applyDecorators(viewUrl);
```
JohannesDoberer, line 255: Do we have to call this function if no `nodeViewURLDecorator` in settings.js is defined?

test/e2e-test-application/src/app/project/dynamic/dynamic.component.html
```diff
@@ -94,8 +94,8 @@ <h1 class="fd-section__title">{{ nodeLabel }}</h1>
   </div>
 </section>
 
-<section class="fd-section" *ngIf="nodeParams && hasBack">
-  <div class="fd-layout-panel fd-has-margin-bottom-small">
+<section class="fd-section" *ngIf="nodeParams || hasBack">
```
UlianaMunich, line 97: Isn't *ngIf="nodeParams* duplicated in line 97 and 98?

core/src/services/viewurl-decorator.js
```diff
@@ -0,0 +1,46 @@
...
+    for (let i = 0; i < queryParamDecorators.length; i++) {
+      const decorator = queryParamDecorators[i];
+      if (urlObj.searchParams.has(decorator.key)) {
+        urlObj.searchParams.delete(decorator.key);
+      }
+      const value = decorator.valueFn();
+      urlObj.searchParams.append(decorator.key, value);
+      console.log('appending', decorator.key, value);
+    }
+    console.log('urlObj.href', urlObj.href);
```
UlianaMunich, line 41: are those 2 console.log lines left on purpose or we can get rid of it?

test/e2e-test-application/e2e/tests/0-fiddle/fiddle-navigation.spec.js
```diff
@@ -337,4 +337,88 @@ describe('Fiddle', () => {
...
+        cy.getIframeWindow().then(win => {
+          const defaultTheme = win.LuigiClient.uxManager().getCurrentTheme();
+          expect(defaultTheme).to.equal('light');
+
+          // not yet implemented
+          // win.LuigiClient.uxManager().setCurrentTheme('dark');
+          // expect(defaultTheme).to.equal('dark');
+        });
+      });
+    });
```
UlianaMunich, line 389: looks like there are 2 similar tests, maybe the one was planned to go through false-negative results?

test/e2e-test-application/src/app/project/dynamic/dynamic.component.html
maxmarkus: Nope, since we don't want to show the section at all, if none of the given parameters are defined.

core/src/services/viewurl-decorator.js
maxmarkus: thx, shall be removed.

core/src/utilities/helpers/iframe-helpers.js
maxmarkus: I will add a method `hasDecorators()` to check if there are any decorators defined. Btw, its not tied to `nodeViewURLDecorate` setting, we can use this class from any place for other decorators too (like substitution, etc)

PR #1543, opened 2020-08-10
Improvements on simpledev, live-server, etc

### LINKED ISSUE #1449: Content security policy issues on Luigi
Description from a customer:

> Luigi uses eval in the JS code as well as inline styles. These two solutions are considered as unsafe by CSP, as we all know.
> The problem is that you can allow or not to execute certain script in many ways but you can’t control the content of the script. In our case the luigi.js script is executed properly (via nonce directive) but we still have to globally allow unsafe-eval because of its content.
> The best solution would be to allow unsafe-eval just for this particular script but, as I said before, it’s impossible to be done via CSP.

linked issue: https://github.tools.sap/kyma/backlog/issues/129

maxmarkus:
Improvements after #1449, followup of #1565

### DIFF
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -1,0 +1,8 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: Luigi
+Upstream-Contact: Philipp Pracht <philipp.pracht@sap.com>
+Source: https://github.com/SAP/luigi
+
+Files: *
+Copyright: 2020 SAP SE or an SAP affiliate company and Luigi contributors
+License: Apache-2.0
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,12 +13,10 @@ You are welcome to contribute with your pull requests. These steps explain the c
 2. If you've added code that should be tested, add tests.
 3. If you've changed APIs, update the documentation.
 4. Make sure the tests pass.
-5. Sign the Contributor License Agreement (CLA).
+5. Sign the Developer Certificate of Origin (DCO).
 
-## Contributor License Agreement (CLA)
-You need you to sign the CLA for your pull request to be accepted. You only need to do it once.
-
-Sign your CLA [here](https://cla-assistant.io/SAP/luigi).
+## Developer Certificate of Origin (DCO)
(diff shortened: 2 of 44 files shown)


ndricimrr:
Should we get rid of `babel.config.js` or part of it 🤔  ?

ndricimrr:
Some notes I pulled while looking at this in general. Probably helpful to keep track of why we changed it for future reference. 

### **Removed Parts:** 

1. **babel** 

1. **core-js**

1. **regenerator-runtime**

1. **terser**

1. **live-server**

### **History of removed parts:** 

1.  **babel** , **core-js** and **regenerator-runtime** were first added in the same PR [Svelte 3 migration](https://github.com/SAP/luigi/pull/843) as follows: 

    - [babel](https://github.com/SAP/luigi/pull/843/files#diff-2d09a11eca0d7df6b8d07ac91045ecc9R31 ):  

    - [regenerator-runtime](https://github.com/SAP/luigi/pull/843/files#diff-a0bfe39a73daba90b2545c23cf2c9275R96 ): 

    - [core-js](https://github.com/SAP/luigi/pull/843/files#diff-a0bfe39a73daba90b2545c23cf2c9275R95) : 

2. **Terser** is introduced later in the [Improve js bundle size ](https://github.com/SAP/luigi/pull/996) PR [here](https://github.com/SAP/luigi/pull/996/files#diff-a0bfe39a73daba90b2545c23cf2c9275R31)

3. In same [PR](https://github.com/SAP/luigi/pull/996) some slight modifications to **babel** by adding [source-map option as inline ](https://github.com/SAP/luigi/pull/996/files#diff-a0bfe39a73daba90b2545c23cf2c9275R30)

### Some results so far
There seem to be changes in bundle sizes: 
- JS bundle size before:
image

- JS bundle size after: 
image

Diff in size ≈ -100 KB 

Also, due to removing babel build configs, **luigi.js.map** file which is ≈ 980 KB is also omitted from the resulting build. As it's needed for source mapping (i.e: better debugging) might consider adding source mapping through [webpack's devtool](https://webpack.js.org/configuration/devtool/#production) as an alternative. Could help to consider [this article ](https://css-tricks.com/should-i-use-source-maps-in-production/) to argue with.

maxmarkus:
I am creating a new PR, this one stays and will be merged afterwards with followup improvements.

maxmarkus:
This implementation has some side effects and it was for the purpose of allowing CSP in dev mode.

PR #1561, opened 2020-08-13
Docu: Examples for client lifecycle api

maxmarkus:
I have realized we do not have any examples for LuigiClient lifecycle API, which is the first contact of all users.


client/src/lifecycleManager.js
```diff
@@ -387,8 +419,8 @@ class LifecycleManager extends LuigiClientBase {
    * @param {string} message.id a string containing the message id
    * @param {*} message.MY_DATA_FIELD any other message data field
    * @example
-   * import LuigiClient from '@luigi-project/client';
    * LuigiClient.sendCustomMessage({id: 'environment.created', production: false})
+   * LuigiClient.sendCustomMessage({id: 'environment.created', data: environmentDataObj)
```
ndricimrr, line 423:
```suggestion
   * LuigiClient.sendCustomMessage({id: 'environment.created', data: environmentDataObj})
```

docs/luigi-client-api.md
```diff
@@ -109,23 +153,42 @@ Removes a custom message listener.
...
+#### getEventData
 
 Returns the context object. It is an alias function for getEventData().
 
+##### Examples
+
+```javascript
+const context = LuigiClient.getContext()
+```
+
```
ndricimrr, line 191: The title changed to `getEventData` but the example code is still using `getContext()` . 

In the screenshot getEventData seems to have ended up a duplicate 🤔 :  
Screenshot 2020-08-17 at 12 17 44

docs/luigi-client-api.md
```diff
@@ -182,8 +269,8 @@ Sends a custom message to the Luigi Core application.
 ##### Examples
 
 ```javascript
-import LuigiClient from '@luigi-project/client';
 LuigiClient.sendCustomMessage({id: 'environment.created', production: false})
+LuigiClient.sendCustomMessage({id: 'environment.created', data: environmentDataObj)
```
ndricimrr, line 273:
```suggestion
LuigiClient.sendCustomMessage({id: 'environment.created', data: environmentDataObj})
```

docs/luigi-client-api.md
UlianaMunich: @maxmarkus @ndricimrr I see getEventData() is already described above so getContext should stay as it was, no need to remove it and the example as well. What do you think, guys?

ndricimrr: The example codes seem to be in place. The second title I guess was intended for `getContext` instead of `getEventData`. When searching for `getContext` it should then go to the second title in the screenshot

JohannesDoberer: `getEventData` was first and after confusions we introduced `getContext` which does the same. But we go with `getEventData` as well because of backwards compatibility.

maxmarkus: I have used `* @alias getEventData` and this seems to break it. Guess we should set `getEventData()` as deprecated, to not confuse people with two equal functions.

maxmarkus: I have moved up `getContext` and removed the example for `getEventData`, which is set now as deprecated.
Proposal

ndricimrr (approved):
👍

PR #1564, opened 2020-08-14
Integrate plugin readme in docu

### LINKED ISSUE #1203: Docu: integrate plugin readme files and allow external .md links
Find a way to keep plugin/../README.md files in place but also have them rendered in the documentation. The links should remain working in github, and still work also for in the docu app of course.

And it seems external links that end with .md are treated as internal links by the javascript link replacer. Fix it to ignore external links that end with .md.

ndricimrr:
**Description**

Fixes #1203 

- include plugin readme content as nodes:
  
Screenshot 2020-08-14 at 11 05 24

- ignore external links that end with .md:
   i.e: shouldn't work -> [ test ] ( https://github.com/prettier/prettier-vscode/blob/main/CONTRIBUTING.md )

**Related issue(s)**


ndricimrr:
🤔  @UlianaMunich  , This actually happens to me on the master branch as well (I think it's an old issue and probably worth investigating in another task), but I thought it was a local issue for my machine only 😃  , since I ran it again and it worked fine. The generated node file seemed fine, there were no duplicates there which is where the duplicates would show up. Cannot replicate the issue after the first run tho, since it disappeared. Let me know if it re-occurs to you even after re-running it again.

UlianaMunich:
@ndricimrr the issue happens every 2nd time I run `npm install` and then `npm run start`. I will create a bug report then as a separate ticket.

maxmarkus (approved):
LGTM. Also wasn't able to reproduce the local bug, tried to find a possible root cause but wasn't successful. Production build also looks fine.

UlianaMunich (approved):
LGTM!

Issue #1459, opened 2020-07-03
Investigate CSP unsafe-inline for Luigi Core


ndricimrr:
Take a look at _unsafe inline expressions_ [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src). 

Luigi Core contains inline < script > and < style > already and CSP seems to disagree with such inline tags as it considers them a 'potential risk'. 
Unless user sets `'unsafe-inline'` in the CSP configuration the following error occurs:

 ![Screenshot 2020-07-03 at 11.53.13.png](https://images.zenhubusercontent.com/5ddce54970febb0001f9c812/e1b2e117-8a43-45ea-8a11-8c952b6a5065)

Investigate what kind of approach/solution we could take in this direction, which is especially important for users who do not want to add the `'unsafe-inline'` exception  to their CSP configurations

Issue originated from #1449

ndricimrr:
From Svelte's Discord Channel:
![Screenshot 2020-07-06 at 13.56.38.png](https://images.zenhubusercontent.com/5ddce54970febb0001f9c812/46d50482-e9fa-4fd4-8df8-9077af0d19eb)

From Svelte's Discord Channel:
![Screenshot 2020-07-06 at 14.11.06.png](https://images.zenhubusercontent.com/5ddce54970febb0001f9c812/00ab7ea9-f485-4b4a-9e6b-f6d61694d3f8)

ndricimrr:
unsafe-inline is no more an issue for scripts due to fix in this [PR](https://github.com/SAP/luigi/pull/1565). decided to allow unsafe-inline for styles since we need them in some functionality

PR #1565, opened 2020-08-14
Content security policy clean

### LINKED ISSUE #1449: Content security policy issues on Luigi

### LINKED ISSUE #1458: Investigate CSP unsafe-eval for Luigi Core
Take a look at _unsafe eval expressions_ [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src). 

Our Luigi Core generated bundle currently ends up containing [Function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function) constructors that disagree with CSP's unsafe eval rule. 

Investigate a way to get rid of eval & Function constructor and test it by adding a csp meta to the e2e-test application to check. 

Currently the output on e2e-test application looks like this:

Issue originated from #1449

maxmarkus:
Resolves #1458 

We accepted unsafe-inline for style-src as a potential risk, since we rely on some inline styles for configurable sizes like modals.

```
<meta
    http-equiv="Content-Security-Policy"
    content="default-src 'self' ;
    script-src 'self'  ;
    style-src 'self' 'unsafe-inline' ;
    img-src data:" />
```


ndricimrr:
Seems to work fine for solving the ["unsafe-eval" csp](https://github.com/SAP/luigi/issues/1458) issue. It seems like it doesn't solve the 'unsafe-inline' issue. 
Similar Results for the examples as well: 
Screenshot 2020-08-20 at 16 42 50

That would mean we have to link this PR to the [unsafe-eval issue](https://github.com/SAP/luigi/issues/1458) only and work on the [unsafe-inline](https://github.com/SAP/luigi/issues/1459) one afterwards.

PR #1579, opened 2020-08-20
Luigi sample with NextJs

### LINKED ISSUE #1122: Usage with next.js
**Description**

This may turn into a Feature Request. I'm wondering if anyone has used luigi with next.js. I see instructions for a "regular" React.js app. One immediate issue I see is that next.js has its own router.

**Reasons**

We would like to use Luigi with next.js

**Attachments**

n/a

Thank you!

stanleychh:
Provide a sample code of [Luigi](https://luigi-project.io/) is running on top of [Next JS](https://nextjs.org/)

Fixes #1122


core/examples/luigi-example-next/README.md
```diff
@@ -0,0 +1,31 @@
+# Sample Luigi application written in NextJS
+
+## Overview
+
+This is the NextJS-based sample application which runs with Luigi framework.
```
maxmarkus, line 5:
```suggestion
This is the NextJS-based sample application which runs with Luigi framework.
Due to the nature of NextJS's routing strategy, it is not possible to use `useHashRouting: false` in the Luigi Configuration. If you want to use path based routing, Luigi Core's index.html should be served separately.
```

Maybe something like this? So people know about the fact and could do something about it by themself, if they want.

core/examples/luigi-example-next/.gitignore
```diff
@@ -0,0 +1,34 @@
...
+# debug
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+# local env files
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
```
maxmarkus, line 31:
```suggestion
.env*
```

maxmarkus (changes requested):
Just a small docu change.

UlianaMunich (changes requested):
Have a couple of suggestions:
- Include Fundamental Styles and sap theming same as in other examples in package.jsonn
- re-use the same default page layout we use in other examples (then it will be easier for us to upgrade the version of FD Styles all at once for all examples)

UlianaMunich (changes requested):
All looks good except of one tiny thing is a Home icon on the desktop.
See attached
Screen Shot 2020-08-24 at 17 42 18

stanleychh:
@UlianaMunich , good catch. I made a mistake that Luigi core is not necessary to have fundamental styles, only micro frontend part needs.
Next.js has their own ways to apply CSS. FYI: https://nextjs.org/docs/basic-features/built-in-css-support
Currently, there is no good way to import CSS directly from node_modules, so I changed it to CDN way instead. The good news is [this feature](https://github.com/vercel/next.js/issues/12079) will be support in the near feature. Then I will improve it.

core/examples/luigi-example-next/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "luigi-example-next",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
```
ndricimrr, line 6: I am newly getting an error when first trying to run the app after `npm install`. This didn't occur before . 
`Repro steps: `Re-clone the repo and `npm install` then `npm run dev`

Error log: 
```
$ npm run dev

> luigi-example-next@0.1.0 dev /luigi/core/examples/luigi-example-next
> next dev

ready - started server on http://localhost:3000
Error: Cannot find module 'postcss-preset-env' from '/luigi/core/examples/luigi-example-next/'
```

Let me know if you could reproduce it. 

I did install that missing module afterwards to make it work, but that seemed like a weird fix, also since others didn't get this error probably worth investigating a bit if it occurs.

core/examples/luigi-example-next/pages/home.js
```diff
@@ -0,0 +1,24 @@
...
+export default function Home() {
+  useEffect(() => {
+    require('@luigi-project/client/luigi-client.js');
+  }, []);
+
+  return (
+    <>
+      <Head>
+        <link
+          href="//unpkg.com/fundamental-styles@latest/dist/fundamental-styles.css"
```
ndricimrr, line 13: I am not getting any fundamental styles applied to the micro-frontends. I noticed that you switched from Global CSS (in _app.js) to internally linked to the `<Head>` inside each functional component. None of the options showed any fundamental styles applied for me. Let me know if that's the case for you as well or not 🤔, since others also didn't seem to have this behaviour. 
  
This probably gets to be on a lower priority, since we don't want to be spending too much time on it, and it's a Next.js specific thing so feel free to mark this as resolved anyway 👍 .

core/examples/luigi-example-next/package.json
```diff
@@ -0,0 +1,18 @@
...
+    "build": "next build",
```
ndricimrr, line 7: Production build fails with Luigi on it: 

```
$ npm run build
...
Build error occurred
ReferenceError: Luigi is not defined
at Object.08Lt (/luigi/core/examples/luigi-example-next/build/server/pages/luigi-config.js:104:1)
```

This seems to be related to [routing rules in Next.js](https://nextjs.org/docs/routing/introduction). Since you put `luigi-config.js` inside the `pages` folder Next.js will attempt to build it on its own as a separate route (individually) and Luigi is not defined within that scope. 

I would suggest to move `luigi-config.js` outside of the `pages` folder and update the require path in `LuigiCoreApp.js` . This seems to solve that issue.

ndricimrr (changes requested):
Some improvement suggestions. 
Let me know if any of the mentioned comments is not reproducible on your side .

core/examples/luigi-example-next/package.json
UlianaMunich: I also did npm install and then npm run dev and all worked well for me. just fyi

core/examples/luigi-example-next/package.json
stanleychh: Nice catch. I moved it into `public` folder. It works fine. Thanks

core/examples/luigi-example-next/package.json
stanleychh: Well... I didn't face this issue. Not hear other reviewers bring it up either. @UlianaMunich @maxmarkus , did you guys face this problem?

core/examples/luigi-example-next/pages/home.js
stanleychh: I cannot reproduce this issue, the fundamental styles applied to micro-frontends successfully on my side.

core/examples/luigi-example-next/package.json
ndricimrr: If @maxmarkus  also doesn't get it, I assume then it should be good to go 👍  🟢  since I'm getting weird behaviours from other PR's recently too, probably from changing npm or node versions, so could be something only happening in my machine.

ndricimrr: Seems to be resolved. Something local on my side

ndricimrr (approved):
👍

UlianaMunich (approved):
LGTM!

PR #1592, opened 2020-08-31
Check travis pipeline

maxmarkus:
No tags were available locally due to the shallow cloning of travis.
This PR adds a `git pull --tags` if no local tag was found previously. Restricted this to 500, to not download the full repo. Added another check afterwards to verify a tag was found.


test/e2e-test-application/src/main.ts
```diff
@@ -4,6 +4,9 @@ import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
 import { AppModule } from './app/app.module';
 import { environment } from './environments/environment';
 
+import * as LuigiClient from '@luigi-project/client';
+window.LuigiClient = LuigiClient;
```
ndricimrr, line 8: the tag fix looks well. 
seems to fail on here : 
_`ERROR in src/main.ts(8,8): error TS2339: Property 'LuigiClient' does not exist on type 'Window'.`_

maxmarkus: I have fixed that one. Thanks.

PR #1599, opened 2020-09-02
update npm dependencies

JohannesDoberer:


UlianaMunich (changes requested):
@JohannesDoberer could you please check the packages which are dependant on dot-prop: https://www.npmjs.com/browse/depended/dot-prop?offset=36

I have found lerna only, which requires it. So maybe we could update lerna which will update dot-prop as result?

ndricimrr (approved):
LGTM 👍

PR #1600, opened 2020-09-02
Improve OAuth2 callback.html

### LINKED ISSUE #1266: Improve OAuth2 callback.html
Implement checks that gives developers a hint about wrong or missing auth data in the redirect_uri returned params. 
Could be directly implemented in callback.html or in the auth provider.

Check existence: 
- id_token
- expires_in - if no `expires_in` is available, check if it is possible to parse the id_token which might contain a token expiry timestamp.

Add Fallback:
- use id_token if no access_token is present

We have a variable definition `var tokenLifetimeDays = 7;` which does not make sense since it is an access token with a short lifetime. 
Since [RFC expires_in](https://tools.ietf.org/html/rfc6749) says: _The lifetime in seconds of the access token_ it might be sufficient to do something like `var expiryInMills = new Date().getTime() + expiresIn * 1000`.

stanleychh:
Fixes #1266 

1. use id_token if no access_token is present
2. expires_in - if no expires_in is available, check if it is possible to parse the id_token which might contain a token expiry timestamp.


maxmarkus (approved):
LGTM

JohannesDoberer (approved):
👍 LGTM!

PR #1623, opened 2020-09-11
Feature accordion categories

azriel46d:
Description

In the navigation settings a Boolean to have only one category collapsed at any given time has been added. By default this is considered as false so that it keeps the functionality as is. Once toggled to true, the navigation acts as an accordion.

- Updated LeftNav.html to listen to the variable
- Updated the storeExpandedState helper to have a replace setting
- Updated the documentation.
- Added tests under navigation


maxmarkus (approved):
Good job :-)

docs/navigation-parameters-reference.md
```diff
@@ -55,6 +55,9 @@ You can configure the way Luigi tackles routing in your application in the `rout
   - **wrongPath** (string): the path that the user tried navigating to.
   - **wasAnyPathFitted** (bool): it is true if Luigi managed to fit a valid path which means **wrongPath** was only partially wrong. Otherwise it is false.
 
+### defaults.sideNavAccordionMode
+- **type**: boolean
+- **description**: defines the behaviour of categories whether multiple categories can be collapsed at the same time or similar to an accordion. (Default: false)
```
alexandra-simeonova, line 60:
```suggestion
- **description**: defines the behaviour of categories. When set to `true`, multiple categories can be collapsed at the same time and the navigation is similar to an accordion. 
- **default**: `false` 
```

docs/navigation-parameters-reference.md
```diff
@@ -320,6 +323,11 @@ settings: {
     ```
 - **since**: 1.4.0
 
+### sideNavAccordionMode
+- **type**: boolean
+- **description**: overrides the default behaviour of categories whether multiple categories can be collapsed at the same time or similar to an accordion. (Default: false)
```
alexandra-simeonova, line 328:
```suggestion
- **description**: overrides the default behaviour of categories. When set to `true`, multiple categories can be collapsed at the same time and the navigation is similar to an accordion. 
- **default**: `false` 
```

alexandra-simeonova (commented):
Good work 👍
Just a small language suggestion (I hope I got the correct meaning)

docs/navigation-parameters-reference.md
azriel46d: Hi @alexandra-simeonova , its actually the opposite. When set to true, only one category is open. The default functionality (false or not defined) is that any amount of collapsible categories can be open. :)

azriel46d:
```suggestion
- **description**: overrides the default behaviour of categories whether multiple categories can be collapsed. When set to `true`, only one category is collapsed. The navigation is similar to an accordion; when the user clicks another category the previously collapsed category is closed and the new one is opened.```

File: docs/navigation-parameters-reference.md (hunk 1 above)
Comment (azriel46d) [reply docs/navigation-parameters-reference.md:60]:
```suggestion
- **description**: overrides the default behaviour of categories whether multiple categories can be collapsed. When set to `true`, only one category is collapsed. The navigation is similar to an accordion; when the user clicks another category the previously collapsed category is closed and the new one is opened.
- **default**: `false`
```

JohannesDoberer (approved):
Well done 👍 !

alexandra-simeonova (approved):

PR #1624, opened 2020-09-15
Fix modal content not displaying issue in Safari

### LINKED ISSUE #1609: Content in modal will not be displayed in Safari
height 100% does not work for `iframeModalCtn`

stanleychh:
Fix #1609 

CSS's height 100% doesn't work in Safari this case. 
My solution is applying `viewport height` instead and it works fine, but I am not sure this is the best way.

Please feel free to make a comment if there is any other better solution. Thanks

### DIFF
--- a/core/src/Modal.html
+++ b/core/src/Modal.html
@@ -189,7 +189,7 @@
 <style type="text/scss">
   .iframeModalCtn {
     position: relative;
-    height: 100%;
+    height: 70vh;
     width: 100%;
     overflow: auto;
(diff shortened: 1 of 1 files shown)


JohannesDoberer (approved):
Fine with this solution. Can't break it ;) , even when we want to change to modal sizes somedays.
👍

UlianaMunich (changes requested):
@stanleychh looks good!

UlianaMunich (approved):

PR #1586, opened 2020-08-26
Update fd styles core examples

UlianaMunich:
Fixes #1574 

Changed version of FD Styles in:

- angular
- js
- vue
- svelte
examples

Svelte examples used old version of luigi: 1.0.1 so I updated it to 1.1.0 to align with the rest of examples

PR #1633, opened 2020-09-21
fix issues with brken left side nav in semiCollapsed mode and scroller

UlianaMunich:
Fixes #1586 

before ttesting please set responsiveNavigation = 'semiCollapsible';


ndricimrr (approved):
works well by me 👍  .
didn't break after trying all the different states

JohannesDoberer (approved):

PR #1636, opened 2020-09-23
Update Cypress to 5.0.0 & Check Travis

ndricimrr:
### **Changes proposed in this pull request:**
Updating cypress to **5.0.0** from *3.8.2* requires adapting tests to breaking changes. 

Changes affecting our tests from the [cypress migration guide](https://docs.cypress.io/guides/references/migration-guide.html#Migrating-to-Cypress-5-0) are: 

1.  Update node version used by Travis to **12.18.4** as it's [needed](https://docs.cypress.io/guides/references/migration-guide.html#Node-js-10-support) by Cypress 5.0.0
2. Add [needed](https://docs.cypress.io/guides/references/migration-guide.html#Linux-dependencies) Linux dependencies under _packages_ 
3. Get rid of **cypress-plugin-retries** as it's [not supported](https://docs.cypress.io/guides/references/migration-guide.html#Tests-retries) by cypress 5.0.0
4. Switch to new test retries strategy [built in cypress](https://docs.cypress.io/guides/references/migration-guide.html#Tests-retries) with 5.0.0
4.1 This change requires a change in each of the test files, which produces a lot of prettifier changes as a side effect. Could do a PR that only prettifies spec files and make this one less crowded.
5. Add [npm package install script](https://github.com/SAP/luigi/pull/1636/files#diff-354f30a63fb0907d4ad57269548329e3R31) manually as it's not found during normal run. 

Issues: 
- Tried to [run travis locally](https://stackoverflow.com/a/49019950) on a Linux 'machine' with docker to debug faster but same exact environment doesn't seem reproducible
- Compatibility Tests on Travis [by design](https://github.com/SAP/luigi/blob/master/scripts/testCompatibility.sh#L180) run on the latest tag version, which still uses Cypress 3.8.2, so probably have to cope with compatibility test randomly failing until this PR is merged and the new tag is released.


JohannesDoberer (approved):

stanleychh (approved):
@ndricimrr I was wondering should we go with the latest stable version directly?
Of course, we can try if v4.2 fix random failing tests according to your reference.

ndricimrr:
> @ndricimrr I was wondering should we go with the latest stable version directly?
> Of course, we can try if v4.2 fix random failing tests according to your reference.

Yes, just fearing there might be breaking changes in the latest stable release. But if travis fix is related to this version change as in reference will most definitely try latest stable release as well.

stanleychh:
> > @ndricimrr I was wondering should we go with the latest stable version directly?
> > Of course, we can try if v4.2 fix random failing tests according to your reference.
> 
> @stanleychh
> Yes, just fearing there might be breaking changes in the latest stable release. But if travis fix is related to this version change as in reference will most definitely try latest stable release as well.

Got it. Logically, we should stick with the latest stable version and fix our corresponding broken tests.

test/e2e-test-application/package.json
```diff
@@ -59,17 +59,16 @@
     "babel-preset-minify": "^0.5.0",
     "codelyzer": "^4.4.1",
     "concurrently": "^4.1.0",
-    "cypress": "^3.8.2",
-    "cypress-plugin-retries": "^1.2.0",
```
ndricimrr, line 63: cypress-plugin-retries not applicable anymore with cypress 5.0.0, [reference](https://docs.cypress.io/guides/references/migration-guide.html#Tests-retries)

ndricimrr:
Closing to form a clean [PR](https://github.com/SAP/luigi/pull/1662) since formatter hook added hundreds of changes

PR #1641, opened 2020-09-24
close splitview after navigation

JohannesDoberer:


core/src/App.html
```diff
@@ -1,4 +1,4 @@
-<svelte:window on:resize="{onResize}"/>
+<svelte:window on:resize="{onResize}" />
```
UlianaMunich, line 1: checked other files and we have no space at the end of the tag. Not a thing, but maybe it is better to be consistent.

core/src/services/split-view.js
```diff
@@ -223,6 +223,11 @@ class SplitViewSvcClass {
           IframeHelpers.getIframeContainer().style.paddingBottom = '';
           SplitViewSvc.sendMessageToClients('close.ok');
         });
+    } else {
+      if (comp.get().mfSplitView && comp.get().mfSplitView.settings && comp.get().mfSplitView.settings.onlyExpandBoxVisible) {
+        comp.get().mfSplitView.displayed = false;
+        comp.set({ mfSplitView: comp.get().mfSplitView });
```
UlianaMunich, line 229: what about declaring the variable above the if statement: let splitViewState = comp.get().mfSplitView; ?

UlianaMunich (changes requested):
Just tiny improvements from my side and maybe a new test for this functionality will ensure no one else could break it? Just an idea

core/src/services/split-view.js
JohannesDoberer: The variable `displayed` is already declared. It comes from  `App.html`. If we are in the if-statement we only should there set the displayed state to false otherwise we'll get side effects and a wrong state.

JohannesDoberer:
Add a test scenario for this bug.

core/src/services/split-view.js
stanleychh, line 229: It seems `comp.get().mfSplitView` repeated a lot, I guess @UlianaMunich expects the code below:
```suggestion
      let splitViewState = comp.get().mfSplitView;
      if (splitViewState && splitViewState.settings && splitViewState.settings.onlyExpandBoxVisible) {
        splitViewState.displayed = false;
        comp.set({ mfSplitView: splitViewState });
```
Please double-check if it works.

stanleychh (approved):

core/src/services/split-view.js
UlianaMunich: yeah, this is what I was suggesting above. A small cleanup to reduce the redundancy and increase the readability. Hope it makes sense

UlianaMunich (approved):

core/src/services/split-view.js
JohannesDoberer: Yes, good point @UlianaMunich. It definitely make sense... just didn't get your point :)

PR #1656, opened 2020-10-09
Revert cypress parallel testing option

ndricimrr:
**Description**

- Due to tests failing randomly on the CI when run in parallel, decided to revert to sequential 
- This is currently being done by removing the second testrunner

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -124,10 +124,6 @@ jobs:
     - name: 'Integration Testing 1'
       <<: *integration-testing
 
-    - name: 'Integration Testing 2'
-      <<: *integration-testing
-
-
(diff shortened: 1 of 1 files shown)


stanleychh (approved):

maxmarkus (changes requested):
I would suggest to revert the changes in `e2e.sh` and just remove the second testrunner from `travis.yml:127`, so it runs only one instance and is sequencially. The `--parallel` can stay, so we can switch back to it simpler at a later time by adding the other testrunner again. 

With the current implementation it would still run two times the integration tests, but all tests sequencially.

maxmarkus (approved):
LGTM!

PR #1664, opened 2020-10-15
Support wc in split view 

### LINKED ISSUE #1545: Support for web components in split view container

ndricimrr:
**Description**

- Fixes #1545

### DIFF
--- a/core/src/SplitView.html
+++ b/core/src/SplitView.html
@@ -63,6 +63,9 @@
   let messageHandler;
   let splitViewIframe;
   let splitViewIframeData;
+  let splitViewWC;
+  let splitViewWCData;
+  let wcCreated;
   export let nodepath;
   export let collapsed;
   export let splitViewSettings = {};
@@ -85,7 +88,9 @@
           nodeParams,
           currentNode,
           splitViewIframe,
-          splitViewIframeData
+          splitViewIframeData,
+          splitViewWC,
+          splitViewWCData
         };
       },
       set: obj => {
@@ -109,6 +114,10 @@
               splitViewIframe = obj.splitViewIframe;
             } else if (prop === 'splitViewIframeData') {
               splitViewIframeData = obj.splitViewIframeData;
+            } else if (prop === 'splitViewWC') {
+              splitViewWC = obj.splitViewWC;
+            } else if (prop === 'splitViewWCData') {
+              splitViewWC = obj.splitViewWC;
             }
           });
         }
--- a/core/src/SplitView.html
+++ b/core/src/SplitView.html
@@ -65,7 +65,6 @@
   let splitViewIframeData;
   let splitViewWC;
   let splitViewWCData;
-  let wcCreated;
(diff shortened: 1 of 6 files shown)


JohannesDoberer (commented):
Good Job 👍 .
Just one thing. Open a wc in splitview and after that (w/o closing the splitview) open a mf in splitview it is completely broken. Should we handle this (and protect the user from this missconfiguration) or consider it as missconfiguration?

ndricimrr:
> Good Job 👍 .
> Just one thing. Open a wc in splitview and after that (w/o closing the splitview) open a mf in splitview it is completely broken. Should we handle this (and protect the user from this missconfiguration) or consider it as missconfiguration?

What do you get on your side: I get duplicates of the element. Do you get the same?
I notice now that this is not an issue custom only to webcomponent in split view : 
Trying this: Open iframe SplitView, open another one iframe in splitview , and you get two, in a similar  behaviour.

I would guess if we decided to fix this then, should do it for master branch directly and not just this branch. can discuss in afterscrum

JohannesDoberer (approved):
👍

test/e2e-test-application/src/app/project/project.component.html
```diff
@@ -483,6 +483,27 @@ <h3 class="fd-layout-panel__title">Navigate</h3>
...
+                linkManager().openAsSplitView('/projects/pr2/webcomponent', {
+                  title: 'Webcomponent in a split view',
+                  size: '25'
+                })
+              "
+              >Open webcomponent in splitView</a
+            >
+            <app-code-snippet
+              data="linkManager()
+              .openAsSplitView('/projects/pr2/webcomponent', {
```
maxmarkus, line 500: I can see a space after linkmanager() which is inconvenient to copy. would be cool to see it changed at the upper openAsModal too.
```suggestion
              data="linkManager().openAsSplitView('/projects/pr2/webcomponent', {
```

core/src/SplitView.html
```diff
@@ -109,6 +113,10 @@ <h1 class="fd-splitView__title">{splitViewSettings.title}</h1>
               splitViewIframe = obj.splitViewIframe;
             } else if (prop === 'splitViewIframeData') {
               splitViewIframeData = obj.splitViewIframeData;
+            } else if (prop === 'splitViewWC') {
+              splitViewWC = obj.splitViewWC;
+            } else if (prop === 'splitViewWCData') {
+              splitViewWC = obj.splitViewWC;
```
maxmarkus, line 119: Should it be `splitViewWCData` or is `splitViewWC` correct?

maxmarkus (changes requested):
Looks nice already, just some small things.

core/src/SplitView.html
ndricimrr: yes `splitViewWCData` is correct, nice catch 👍 , I'll update it

maxmarkus (approved):

PR #1665, opened 2020-10-15
1650 bug nav collapsed state

### LINKED ISSUE #1650: persisting collapsed state does not work anymore
steps to reproduce:
* open luigi app
* collapse side nav
* refresh

result:
side nav is not collapsed

expected: behaviour:
side nav should be collapsed

Additional task:
expose collapse/expand functionality in core API

UlianaMunich:
Fixes #1650 

Extends the Core API UX part with a new function that allows collapsing the left side nav. 

Docu and tests are attached


docs/luigi-core-api.md
```diff
@@ -416,6 +416,128 @@ Type: [Object](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Globa
...
+
+```javascript
+Luigi.auth().login();
+```
+
+**Meta**
+
+-   **since**: NEXTRELEASE
+
+> > > > > > > cc8090040fa22aef1a9eeeb45f0f12db9d485c08
```
JohannesDoberer, line 436: I guess you had some merge conflicts. Some docu parts are doubled and also these less- and greater-than signs.

JohannesDoberer (changes requested):

core/src/navigation/services/semi-collapsed-navigation.js
```diff
@@ -7,11 +7,21 @@ class SemiCollapsibleNavigationClass {
...
-    this.semiCollapsible =
-      this.responsiveNavSetting === 'semiCollapsible' ||
-      this.responsiveNavSetting === 'Fiori3'
-        ? true
-        : false;
+    let currentCollapsedState = localStorage.getItem(
+      NavigationHelpers.COL_NAV_KEY
+    );
+    //checking if here was a previous state in localStorage before the reload
+    if ((currentCollapsedState = true)) {
```
JohannesDoberer, line 14: Probably you want to check if `currentCollapsedState === true` but than it will not work anymore.
My suggestion is to check if `currentCollapsedState` is not `null`. If it is **not** null you go with the state from the localstorage, if it is null you do the else part.

UlianaMunich: updated the PR

docs/luigi-core-api.md
UlianaMunich: cleaned

JohannesDoberer (approved):
LGTM 👍 !

ndricimrr (approved):
👍

core/src/navigation/services/semi-collapsed-navigation.js
```diff
@@ -7,11 +7,20 @@ class SemiCollapsibleNavigationClass {
     this.responsiveNavSetting = LuigiConfig.getConfigValue(
       'settings.responsiveNavigation'
     );
-    this.semiCollapsible =
+    let currentCollapsedState = localStorage.getItem(
```
JohannesDoberer, line 10: My suggestion:
```
let currentCollapsedState = JSON.parse(localStorage.getItem(
      NavigationHelpers.COL_NAV_KEY
    ));
```
In `currentCollapsedState` you will get a string instead of a boolean value.

core/src/navigation/services/semi-collapsed-navigation.js
```diff
@@ -7,11 +7,20 @@ class SemiCollapsibleNavigationClass {
...
+    );
+    let getResponsiveNavSettings =
       this.responsiveNavSetting === 'semiCollapsible' ||
-      this.responsiveNavSetting === 'Fiori3'
-        ? true
-        : false;
+      this.responsiveNavSetting === 'Fiori3';
+
+    //checking if here was a previous state in localStorage before the reload
+    if (currentCollapsedState != false && getResponsiveNavSettings) {
```
JohannesDoberer, line 18: I would remove the else part, because after a reload the variable `this.semiCollapsible` will be always undefined and this lead to problems if you resize from desktop to mobile.
Code could look like this:

```
//checking if here was a previous state in localStorage before the reload
    if (currentCollapsedState !== null && getResponsiveNavSettings) {
      this.isSemiCollapsed = this.getCollapsed();
    }
    this.semiCollapsible = getResponsiveNavSettings ? true : false;
```

JohannesDoberer (changes requested):

core/src/navigation/services/semi-collapsed-navigation.js
```diff
@@ -10,16 +10,15 @@ class SemiCollapsibleNavigationClass {
     let currentCollapsedState = localStorage.getItem(
       NavigationHelpers.COL_NAV_KEY
     );
-    let checkTheResponsiveNavSettings =
+    let getResponsiveNavSettings =
```
stanleychh, line 13: It seems using `const` is more suitable. There is no updated afterward.
I would suggest name it with `is-` prefix. e.g. `const isResponsiveNavSemiCollapsibleOrFiori3`  something like this. It is easier to understand at first sight.

UlianaMunich: done

core/src/navigation/services/semi-collapsed-navigation.js
UlianaMunich: done, thanks for checking

core/src/navigation/services/semi-collapsed-navigation.js
UlianaMunich: done

core/src/navigation/services/semi-collapsed-navigation.js
```diff
@@ -7,19 +7,23 @@ class SemiCollapsibleNavigationClass {
     this.responsiveNavSetting = LuigiConfig.getConfigValue(
       'settings.responsiveNavigation'
     );
-    let currentCollapsedState = localStorage.getItem(
-      NavigationHelpers.COL_NAV_KEY
+    let currentCollapsedState = JSON.parse(
```
stanleychh, line 10: This should be `const` as well.
Is this a `boolean` ?    If so, could you please change to `is` prefix naming?

core/src/navigation/services/semi-collapsed-navigation.js
```diff
@@ -7,19 +7,23 @@ class SemiCollapsibleNavigationClass {
...
     );
-    let getResponsiveNavSettings =
+    const isResponsiveNavSemiCollapsibleOrFiori3 =
       this.responsiveNavSetting === 'semiCollapsible' ||
       this.responsiveNavSetting === 'Fiori3';
 
     //checking if here was a previous state in localStorage before the reload
-    if (currentCollapsedState != false && getResponsiveNavSettings) {
+    if (
+      currentCollapsedState != false &&
```
stanleychh, line 19: Besides the naming, Please change to `!==`

core/src/navigation/services/semi-collapsed-navigation.js
```diff
@@ -7,11 +7,24 @@ class SemiCollapsibleNavigationClass {
...
+    //checking if here was a previous state in localStorage before the reload
+    if (
+      currentCollapsedState != false &&
+      isResponsiveNavSemiCollapsibleOrFiori3
+    ) {
+      this.isSemiCollapsed = this.getCollapsed();
+    }
+    this.semiCollapsible = isResponsiveNavSemiCollapsibleOrFiori3
+      ? true
+      : false;
```
stanleychh, line 26:
```suggestion
    this.semiCollapsible = isResponsiveNavSemiCollapsibleOrFiori3;
```
Also, since `semiCollapsible` is same as isResponsiveNavSemiCollapsibleOrFiori3. Please check do we need to keep `semiCollapsible` variable?

core/src/navigation/services/semi-collapsed-navigation.js
UlianaMunich: it is a string, Hannes mentioned it above, when suggesting the changes

core/src/navigation/services/semi-collapsed-navigation.js
UlianaMunich: done

core/src/navigation/services/semi-collapsed-navigation.js
UlianaMunich: thanks for double-checking, We need it for the mobile collapsed state as well, so that on mobile the left side nav always set to 'true' even after the reload.  And setting `this.semiCollapsible = isResponsiveNavSemiCollapsibleOrFiori3;` doesn't give a chance to make it `false`. 

So I'd rather leave it as it is. What do you think?

core/src/navigation/services/semi-collapsed-navigation.js
stanleychh: This should be a boolean after `JSON.parse`

UlianaMunich: right, mixed it up and now renamed to `isCurrentStateCollapsed`

JohannesDoberer (approved):
Good Job!

PR #1666, opened 2020-10-15
Fix example scripts

### LINKED ISSUE #1663: Fix script for luigi examples

JohannesDoberer:
It seems that angular, react and vue are not compatible with webpack 5 and webpack cli 4.
This fix is just a temporary solution as long as will work with their cli's
Fixes #1663


ndricimrr (commented):
All tested and working fine for me 👍 . 
Except for the angular script, I get a weird error about fundamental styles: Module Error

ndricimrr (approved):

UlianaMunich (changes requested):
Works fine for me now. The only thing is that I was following application-setup.md at first and the manual run script is not updated there. It makes sense to update it there as well, I guess

JohannesDoberer:
That is true. Good point 👍

UlianaMunich (approved):
LGTM!

PR #1668, opened 2020-10-21
Prevent duplicate split view when opening a new split view

### LINKED ISSUE #1667: Restrict splitview opening multiple iframes/wc's
Repro steps:
1- Open an iframe in a splitview
2- While the splitview is opened open another iframe in a splitview
You'll see that the iframe's are stacked on top of each other instead of keeping just one.

Alternative Retro Steps: 
1- Open a wc in a splitview
2- While the splitview is opened open another wc in the splitview
You'll see that the web components are stacked on top of each other instead of keeping just one.

maxmarkus:
Fixes #1667


JohannesDoberer:
Open a splitview 
`LuigiClient.linkManager().withParams({test: true}).openAsSplitView('/settings', {title:'Logs', size:25})` and afterwards try to open a splitview with different path `LuigiClient.linkManager().withParams({test: true}).openAsSplitView('/projects/pr2/developers', {title:'Tets', size:25})` does not work.
I see still the `settings` view. Title of splitview changed.

PR #1689, opened 2020-11-04
Safari bug with shellbar dropdowns blue outlie

### LINKED ISSUE #1672: Weird bug with "Select Project" outline in Safari

UlianaMunich:
Screen Shot 2020-11-03 at 10 17 42

Fixes #1672 
Open Safari and then click on any dropdown the blue border lie shouldn't be there.


core/src/navigation/TopNav.html
```diff
@@ -630,6 +630,11 @@
 </script>
 
 <style type="text/scss">
+  //remove default outline of all dropdowns in shellbar
+  *:focus {
```
stanleychh, line 634: Is it necessary to select everything?
Will it work properly If the selector shrink to `fd-shellbar` class only?

UlianaMunich: absolutely. Just tested and worked well. Thanks for the suggestion!

stanleychh (approved):

ndricimrr (approved):
works pretty good for me, nice fix 👍

PR #1697, opened 2020-11-10
Add WC API

### LINKED ISSUE #1546: Web components Client API

ndricimrr:
**Description**
Resolves #1546 

- Currently we expose the whole luigi object to the web component, which is then used to call api functions upon. This can result in misuses on the client web-component developers, i.e.: they might override the object, or might use api calls that would not be client but core related. 
- This PR exposes only parts of the core that makes sense in the client perspective, currently its **uxManager** and **linkManager** but it can be extended later. 
- You can access client api in this form : **this.LuigiClient.uxManager()** instead of _this.luigi.ux()_


hardl (approved):
LGTM

stanleychh (approved):

PR #1754, opened 2020-12-02
LuigiClient API to access user settings

### LINKED ISSUE #1685: Extend LuigiClient API to access the settings
As a MF developer I want to access the user settings via LuigiClient API

stanleychh:
- Provide `LuigiClient.getUserSettings()` API which will return corresponding object for micro-frontend. It is based on user settings' group at luigi config file.

**Related issue(s)**
Fixes #1685


docs/luigi-core-api.md
```diff
@@ -188,7 +190,9 @@ Luigi.readUserSettings();
 
 #### storeUserSettings
 
-Writes the user settings object to the **localStorage**. It is possible to choose a custom storage type by implementing the `settings.userSetting.storeUserSettings` function.
+Reads the user settings object.
```
legteodav, line 193:
```suggestion
Writes the user settings object.
```

docs/luigi-client-api.md
```diff
@@ -293,6 +293,18 @@ LuigiClient.sendCustomMessage({id: 'environment.created', data: environmentDataO
...
+#### getUserSettings
+
+Returns the currently user settings based on selected node.
+
+##### Examples
+
+```javascript
+const userSettings = LuigiClient.getUserSettings()
+```
+
```
UlianaMunich, line 305:
```suggestion
**Meta**

-   **since**: NEXTRELEASE
```

docs/luigi-client-api.md
```diff
@@ -293,6 +293,18 @@ LuigiClient.sendCustomMessage({id: 'environment.created', data: environmentDataO
...
+Returns **[Object](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object)** current user settings
```
UlianaMunich, line 306: I'd suggest moving this line to 299

client/luigi-client.d.ts
```diff
@@ -666,3 +672,10 @@ export type uxManager = () => UxManager;
  */
 export function storageManager(): StorageManager;
 export type storageManager = () => StorageManager;
+/**
+ * Returns the currently user settings.
```
alexandra-simeonova, line 676:
```suggestion
 * Returns the current user settings.
```

client/src/lifecycleManager.js
```diff
@@ -440,5 +440,16 @@ class LifecycleManager extends LuigiClientBase {
     );
     helpers.sendPostMessageToLuigiCore(customMessageInternal);
   }
+
+  /**
+   * Returns the currently user settings based on selected node.
```
alexandra-simeonova, line 445:
```suggestion
   * Returns the current user settings based on the selected node.
```

client/luigi-client.d.ts
```diff
@@ -666,3 +672,10 @@ export type uxManager = () => UxManager;
...
+ * @returns {Object} current user settings
```
alexandra-simeonova, line 677:
```suggestion
 * @returns {Object} current user settings
 * @since NEXTRELEASE
```

client/src/lifecycleManager.js
```diff
@@ -440,5 +440,16 @@ class LifecycleManager extends LuigiClientBase {
...
+   * @returns {Object} current user settings
```
alexandra-simeonova, line 446:
```suggestion
   * @returns {Object} current user settings
   * @since NEXTRELEASE
```

ndricimrr:
Could we have some e2e test for this ?

legteodav:
> Probably we could we have some e2e test for this ?

If you need help, I can write some tests in e2e... let me know

stanleychh:
Thanks @legteodav . Hannes and I are debugging an issue which might be the root cause makes my e2e tests(not push yet) failed. I will contact you if I need any help.

docs/luigi-core-api.md
alexandra-simeonova, line 193:
```suggestion
Writes the user settings object.
```
Unless I'm mistaken the description should be different than ` readUserSettings` above

alexandra-simeonova (approved):

legteodav (approved):

core/src/App.html
```diff
@@ -237,7 +237,17 @@
...
       return;
     }
-    const internal = prepareInternalData(config);
+
+    let internal = prepareInternalData(config);
+    const userSettingsGroupName =
+      config.iframe.luigi.currentNode.userSettingsGroup;
+
+    if (userSettingsGroupName) {
+      const userSettingGroups = await LuigiConfig.readUserSettings();
```
ndricimrr, line 246: In case `readUserSettings` returns null, we would have an error when opening the mf, like: 
**TypeError: Cannot read property 'userAccount' of null**
This would happen in case settings are not yet set either through localStorage or through custom readUserSettings not being set.
Probably we can put a warning here that no settings  were found or sth similar to avoid the error:
Screen Shot 2020-12-16 at 11 51 33

ndricimrr (changes requested):
All looks good 👍 . Just a small thing

core/src/App.html
stanleychh: Thanks @ndricimrr . I fixed this error at https://github.com/SAP/luigi/pull/1754/files#diff-7a7028de38452fdd8eef4c66743213004e94b3127d3e088c63fb65b09d587f7aR256-R258. Please have a look again.

core/src/App.html
```diff
@@ -246,7 +246,17 @@
...
+      const userSettings = userSettingGroups[userSettingsGroupName];
```
ndricimrr, line 256: [To the previous thread](https://github.com/SAP/luigi/pull/1754/files#r544245121): `return localStorageValue && JSON.parse(localStorageValue);`

That would indeed return false/null instead of possible undefined, but you'd still be trying to access `userSettingsGroupName`  of a null value. 
For me it still shows **TypeError: Cannot read property 'userAccount' of null**. I removed the cached when I retested to check the behavior, probably you have it still saved

stanleychh: Thanks, @ndricimrr for pointing this out. I push a fix.

ndricimrr (approved):
Looks good 👍

PR #1763, opened 2020-12-07
Run globally prettier/eslint before every commit (issue-1721)

legteodav:
Here following features:
- Create standard eslint configuration in file .eslintrc (extends the standad)
- Run Prettier/ES link before every git commit (husky will call npm run code-quality)
- Possibility to scan (prettier+eslint) all code using code-quality
- Configuration in package.json do disable prettier and eslint
- Prettier json configuration (every extension type could have different configuration)

Resolves #1721

PR #1765, opened 2020-12-09
Fix drawer undefined bug

ndricimrr:
**Description**

- While testing [EsLinter PR](https://github.com/SAP/luigi/pull/1763) randomly with the **"linkManager.js"** the pre-commit hook failed on: `drawer is undefined.`
Apparently we missed it on the reviews.

 Should probably reconsider priority for the ESLinter since it can avoid a lot of such mistakes/bugs in the future.


legteodav (approved):
Hard review 🥇

JohannesDoberer (approved):

PR #1799, opened 2021-01-05
Make navigateIframe and prepareInternalData functions be async

### LINKED ISSUE #1797: Make navigateIframe and prepareInternalData functions be asynchronous
Issue: MF does not get updated while switching to a different `userSettingsGroup` (listen to context update)

Solution: make both `prepareInternalData` and `navigateIframe` be asynchronous.

stanleychh:
**Description**

- Make `navigateIframe` and `prepareInternalData` functions be asynchronous
- MFs will get corresponding `userSettings` object while switching around different `userSettingGroup`

**Related issue(s)**
Fixes #1797

### DIFF
--- a/core/src/services/iframe.js
+++ b/core/src/services/iframe.js
@@ -197,7 +197,7 @@ class IframeClass {
     return !config.iframe.luigi.initOk;
   }
 
-  navigateIframe(config, component, node) {
+  async navigateIframe(config, component, node) {
     clearTimeout(this.timeoutHandle);
     const componentData = component.get();
     let viewUrl = componentData.viewUrl;
@@ -347,6 +347,8 @@ class IframeClass {
       config.iframe['vg'] = this.canCache(componentData.viewGroup)
         ? componentData.viewGroup
         : undefined;
+      config.iframe.luigi.currentNode = componentData.currentNode;
+      const internalData = await component.prepareInternalData(config);
       const message = {
         msg: 'luigi.navigate',
         viewUrl: viewUrl,
@@ -355,7 +357,7 @@ class IframeClass {
         ),
         nodeParams: JSON.stringify(Object.assign({}, componentData.nodeParams)),
         pathParams: JSON.stringify(Object.assign({}, componentData.pathParams)),
-        internal: JSON.stringify(component.prepareInternalData(config))
+        internal: JSON.stringify(internalData)
       };
 
       const withSync = componentData.isNavigationSyncEnabled;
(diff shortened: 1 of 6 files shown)


core/src/services/iframe.js
```diff
@@ -197,7 +197,7 @@ class IframeClass {
     return !config.iframe.luigi.initOk;
   }
 
-  navigateIframe(config, component, node) {
+  async navigateIframe(config, component, node) {
```
legteodav, line 200: We transformed the function to async.... not whenever we call it, we transform it as a promise....So the callers (at least then do not specify to "await", the will not wait for the end of this function and it could potentially get some errors.... It is called here without await...: 
-  File: iframe.js | method setOkResponseHandler | line 159
-  File iframe| method handleRouteChange | line 390 

seeing quickly the code, I don't think we will have problems, but async function can generate some mess....

core/src/services/iframe.js
```diff
@@ -347,6 +347,8 @@ class IframeClass {
       config.iframe['vg'] = this.canCache(componentData.viewGroup)
         ? componentData.viewGroup
         : undefined;
+      config.iframe.luigi.currentNode = componentData.currentNode;
+      const internalData = await component.prepareInternalData(config);
```
legteodav, line 351: Would it be possible (if it is worth it) to remove the await (and async from function declaration) changing the method invocation:

 const message = {
        msg: 'luigi.navigate',
        viewUrl: viewUrl,
        context: JSON.stringify(
          Object.assign({}, componentData.context, { goBackContext })
        ),
        nodeParams: JSON.stringify(Object.assign({}, componentData.nodeParams)),
        pathParams: JSON.stringify(Object.assign({}, componentData.pathParams)),
        internal: JSON.stringify(internalData)
      };

      const withSync = componentData.isNavigationSyncEnabled;
      if (withSync) {
        // default, send navigation event to client
        IframeHelpers.sendMessageToIframe(config.iframe, message);
        this.setOkResponseHandler(config, component, node);
      } else {
        // `withoutSync()` used. client navigation was skipped, reset after one-time use.
        component.set({ isNavigationSyncEnabled: true });
      }

      // clear goBackContext and reset navigateBack after sending it to the client
      component.set({ goBackContext: undefined, isNavigateBack: false });

We will just lose control on when sending message to the Iframe.... (that I guess it would not be a problem...

stanleychh: @legteodav Thanks for your time taking a look and provide a good solution. I will play this solution around and see how it works.

core/src/services/iframe.js
stanleychh: @legteodav  I've pushed a fix which applied the `await` term on `navigateIframe`. I somehow feel it is cleaner from a syntax point of view. Thanks for point this out.

legteodav (approved):

JohannesDoberer (approved):

PR #1845, opened 2021-01-26
Documentation for luigi client angular support library

### LINKED ISSUE #1811: Documentation for luigi client angular support library
Currently it is documented at a readme file in the pkg folder, should be added to our docu app, possible deserves an own main entry and own category ("framework support libraries" or so).

alexandra-simeonova:
**Description**
Currently it is documented at a readme file in the pkg folder, should be added to our docu app.

- create category with client support library documentation 

**Related issue(s)**
Resolves #1811


docs/framework-support-libraries.md
```diff
@@ -0,0 +1,137 @@
...
+ }
+ ```
+
+We also provide an example of how to extend **LuigiRouteStrategy** in class **LuigiReuseRouteStrategy**.
+In this class, we added the possibility to "reuse" a component and not initialize it every time you load it (it could be useful to keep component state.)
+
+**LuigiReuseRouteStrategy** can be configured in the following way:
+ ```javascript
+{path: 'luigi-client-support-preload',component: Sample1Component,data: { reuse: true }}
+ ```
```
legteodav, line 137:
```suggestion
We also provide an example of how to extend **LuigiRouteStrategy** in class **LuigiReuseRouteStrategy**.
In this class, we added the possibility to "reuse" a component and not initialize it every time you load it (it could be useful to keep component state.)

**LuigiReuseRouteStrategy** can be configured in the following way:
 ```javascript
{path: 'luigi-client-support-preload',component: Sample1Component,data: { reuse: true }}
 ```
```

alexandra-simeonova: If we remove the ``` at the very end, I think the `code block` will not be formatted correctly. (See previous examples in the doc.) All the other text in your suggestion is the same as the original. I'm not sure if you meant to add another suggestion besides removing the ticks 🤔

legteodav: Hi @alexandra-simeonova, 

yes my bad: we actually delete LuigiReuseRouteStrategy from code, so we would just need to remove  from line 131 to line 137. (I fixed the suggestion).

Cheers
Davide

legteodav (approved):

UlianaMunich (approved):

docs/framework-support-libraries.md
```diff
@@ -0,0 +1,129 @@
...
+  ,LuigiAngularSupportModule
+],
+```
+
+### Features
+
+These are the main features provided by the library:
+* [LuigiContextService](#LuigiContextService)
+* [LuigiAutoRoutingService](#LuigiAutoRoutingService)
+
```
hardl, line 46: also Preload component should be mentioned here

hardl (changes requested):

docs/framework-support-libraries.md
```diff
@@ -43,6 +43,7 @@ imports: [
 These are the main features provided by the library:
 * [LuigiContextService](#LuigiContextService)
 * [LuigiAutoRoutingService](#LuigiAutoRoutingService)
+* [Preload component](##LuigiAutoRoutingService)
```
hardl, line 46: no explanation, link not working
Please add something like
"An empty angular component that can be used to build a preload route, see [link to viewgroup...preloadurl]"

hardl (changes requested):

docs/framework-support-libraries.md
alexandra-simeonova: Updated in e01d9f7

hardl (approved):

PR #1859, opened 2021-02-02
Change Profile Avatar layout and logic

### LINKED ISSUE #1858: circular profile avatar
profile avatar should have a circular shape, according to https://sap.github.io/fundamental-styles/?path=/docs/components-shellbar--primary

This might require a html restructuring, using fd-avatar

UlianaMunich:
Guidelines: https://sap.github.io/fundamental-styles/?path=/docs/components-shellbar--primary

New: "The primary shellbar displays a logo, title, and an avatar where the user settings are found. If the avatar does not contain a thumbnail, then the initials of the user will be displayed."

Fixes #1858


JohannesDoberer (changes requested):
user_settings_dialog.spec.js is failling.
Please change on line 7:
`cy.get('.fd-user-menu button').should('exist').click();`
to
`cy.get('.fd-avatar').should('exist').click();`

Besides that.. everything is fine 👍

JohannesDoberer (approved):

legteodav (approved):

PR #1861, opened 2021-02-02
Check if meta data are configured

JohannesDoberer:
check if meta data are configured


core/src/UserSettingsDialog.html
```diff
@@ -117,10 +117,26 @@
...
+          if (!Array.isArray(property.options)) {
+            console.error('There is no options array', property.label ? `for "${property.label}"` : '', "defined in the Luigi userSettings config.");
+          }
+        } else if (property.type === undefined) {
+          console.error('There is no "type" ', property.label ? `for "${property.label}"` : '', "defined in the Luigi userSettings config.");
+        }
+      }
+    }
+  }
+
```
legteodav, line 134:
```suggestion
    function errorHandling(selectedUserSettingGroupData) {
    if (!GenericHelpers.isObject(selectedUserSettingGroupData.settings)) {
      return;
    }
    for (let key in selectedUserSettingGroupData.settings) {
      let property = selectedUserSettingGroupData.settings[key];
      if (property.type === 'enum') {
        if (!Array.isArray(property.options)) {
          console.error('There is no options array', property.label ? `for "${property.label}"` : '', "defined in the Luigi userSettings config.");
        }
      } else if (property.type === undefined) {
        console.error('There is no "type" ', property.label ? `for "${property.label}"` : '', "defined in the Luigi userSettings config.");
      }
    }
  }
```
just a bit better for reading

legteodav (commented):
Just added a very minor code suggestion (just to improve a bit readability)

legteodav (approved):

stanleychh (approved):

stanleychh (approved):

PR #1863, opened 2021-02-03
Support types for confirmation modal

### LINKED ISSUE #1850: support types for confirmation modal
like we have for alerts

see https://sap.github.io/fundamental-styles/?path=/docs/components-message-box--structure, "Semantic types"

so we need to switch from fd-dialog to fd-message-box

stanleychh:
**Description**

- Support types for confirmation modal
- Replace css class from `fd-dialog` to `fd-message-box`
- Semantic types: `confirmation`, `success`, `error`, `warning`, `information`
https://sap.github.io/fundamental-styles/?path=/docs/components-message-box--structure
- confirmation button can be hidden if set `settings.buttonConfirm` to `false`

Fixes #1850


UlianaMunich (changes requested):
A couple of additional suggestions from my side:
- we need to keep the previous version of the confirmation modal to support our non-sap related customers (the one which doesn't have an icon in the dialog header). I mean we still can use the confirmation modal but without the question mark. Means icon might be optional same as buttonConfirm.
- depending on the type of dialog box there might be 2 buttons OR one button in the footer  (error, success, warning, information has only one button)
- I think it makes sense to add all types of the confirmation modal we support into the documentation. Currently, only confirmation is listed
- could you please add some test to test 1 button in the footer, if possible?

UlianaMunich (approved):
Great job!

client/src/uxManager.js
```diff
@@ -72,14 +72,16 @@ class UxManager extends LuigiClientBase {
    * Shows a confirmation modal.
    * @memberof uxManager
    * @param {Object} settings the settings of the confirmation modal. If you don't provide any value for any of the fields, a default value is used
+   * @param {('confirmation'|'success'|'warning'|'error'|'information')} settings.type the content of the modal type. (Optional)
    * @param {string} [settings.header="Confirmation"] the content of the modal header
    * @param {string} [settings.body="Are you sure you want to do this?"] the content of the modal body
-   * @param {string} [settings.buttonConfirm="Yes"] the label for the modal confirm button
+   * @param {string|false} [settings.buttonConfirm="Yes"] the label for the modal confirm button. Set false for not showing confirmation button.
```
alexandra-simeonova, line 78:
```suggestion
   * @param {string|false} [settings.buttonConfirm="Yes"] the label for the modal confirmation button. If set to `false`, the button will not be shown.
```

JohannesDoberer (changes requested):
Looks good 👍 .
One additional thing. We should change also documentation for core api in ux.js and the corresponding core docu as well.

alexandra-simeonova (approved):
Nice work!

JohannesDoberer (approved):
👍  LGTM!!!

PR #1873, opened 2021-02-08
1824 Silent Renew for OIDC is not working (only doc)

legteodav:


JohannesDoberer (approved):

plugins/auth/public/auth-oidc/README.md
```diff
@@ -61,17 +61,39 @@ Luigi.setConfig({
 })
 ```
 
-If you want to use the silent token renewal feature, the `silent-callback.html` needs to be copied to a folder in your Luigi Core installation, which is the return path for the IdP provider, configured through the `redirect_uri` setting. The default location of `redirect_uri` is `/assets/auth-oidc/silent-callback.html`. This file also contains a reference to `/assets/auth-oidc/plugin.js` to make the OIDC client available.
+If you want to use the silent token renewal feature, the `silent-callback.html` needs to be copied to a folder in your Luigi Core installation, 
+which is the return path for the IdP provider, configured through the `redirect_uri` setting. The default location of `redirect_uri` is `/assets/auth-oidc/silent-callback.html`. 
+You must install oidc-client in your project as dev dependency:
```
alexandra-simeonova, line 66:
```suggestion
You must install `oidc-client` in your project as a dev dependency:
```

plugins/auth/public/auth-oidc/README.md
```diff
@@ -61,17 +61,39 @@ Luigi.setConfig({
...
+
+```javascript
+npm i -save-dev copy-webpack-plugin
+```
+
+Now you would need to import plugin files and oidc-client library in your project, using webpack:
```
alexandra-simeonova, line 72:
```suggestion
Then, you need to import the plugin files and `oidc-client` library in your project using webpack:
```

alexandra-simeonova (approved):

Issue #1829, opened 2021-01-21
Collect frequent asked questions from Luigi channel and add it to the FAQ section


JohannesDoberer:

alexandra-simeonova:
PR: https://github.com/SAP/luigi/pull/1928

PR #1928, opened 2021-03-12
Add FAQs

alexandra-simeonova:
**Description**

- Collect frequent asked questions from Luigi channel and add it to the FAQ section

**Related issue(s)**
See also #1829


docs/faq.md
```diff
@@ -87,3 +88,613 @@ TBD
...
+    <!-- IMPORTANT -->
+​
+    <!-- HTML -->
+  </body>
+</html>
+```
+
+### How do I disable the Luigi loading indicator? / My micro frontend page is stuck on a loading screen.
+
+If you include micro frontends in Luigi which don't use Luigi Client, Luigi Core won't know when they are ready because there is no handshake. This can cause the page to remain loading indefinitely. In that case, you should disable the loading indicator using the [loadingIndicator.enabled]https://docs.luigi-project.io/docs/navigation-parameters-reference/?section=loadingindicatorenabled parameter in your configuration file.
```
ndricimrr, line 568: Something wrong in the link here, didn't appear there: 

```suggestion
If you include micro frontends in Luigi which don't use Luigi Client, Luigi Core won't know when they are ready because there is no handshake. This can cause the page to remain loading indefinitely. In that case, you should disable the loading indicator using the [loadingIndicator.enabled](https://docs.luigi-project.io/docs/navigation-parameters-reference/?section=loadingindicatorenabled) parameter in your configuration file.
```

ndricimrr: Other than this LGTM 👍

ndricimrr (approved):
LGTM 👍

PR #1935, opened 2021-03-17
Add warning if both virtualTree and children nodes are defined 

### LINKED ISSUE #1823: display warning on console if children and virtualTree is both defined at a node
If virtualTree is set to true, explicit children are ignored, which  is fine but can be confusing, if you don't know.
Since this state can be considered a misconfiguration, a warning on console seems to be appropriate.

ndricimrr:
Fixes #1823 

**Description**
- Defining children when `virtualTree` enabled does not break the navigation (since when virtualTree is enabled we always ([(re)assign the new virtually built children](https://github.com/SAP/luigi/blob/master/core/src/navigation/services/navigation.js#L293)) it is nonetheless a mis-configuration that can create confusion.

**Solution**
- This small PR adds a warning in case virtualTree and children nodes are defined at the same time, thus mitigating potential confusions.

### DIFF
--- a/core/src/navigation/services/navigation.js
+++ b/core/src/navigation/services/navigation.js
@@ -290,6 +290,11 @@ class NavigationClass {
         _virtualViewUrl
       });
 
+      if (node.children) {
+        console.warn(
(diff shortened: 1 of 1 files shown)


JohannesDoberer (approved):
LGTM!

PR #1958, opened 2021-03-30
Use Byline component for user settings dialog

### LINKED ISSUE #1921: Use byline list component in user settings dialog
Investigate if we can get ride of our customized left nav implementation and use the Fundamental styles `Byline list` component.

Check which Fundamental styles version the byline list component is available.
If it is available we can apply it.

UlianaMunich:
Fixes #1921 

Specs:
FIORI 3:https://wiki.wdf.sap.corp/wiki/display/visualcore/List+%28Fiori+3%29+-+Standard+List+with+Byline
FD Styles: https://sap.github.io/fundamental-styles/?path=/docs/components-list-byline--standard

Keyboard interaction:
https://cxwiki.sap.com/display/luigi/Fiori+Keyboard+Interaction

Screen Shot 2021-03-30 at 13 04 48


core/src/UserSettingsDialog.html
```diff
@@ -370,87 +384,8 @@
...
-    color: var(--sapContent_LabelColor);
-    padding-top: 0.5rem;
-    white-space: nowrap;
-    overflow: hidden;
-    text-overflow: ellipsis;
-    flex: 1 1 auto;
-  }
-
   .iframeUserSettingsCtn {
+    display: none;
```
ndricimrr, line 388: I don't know where this selector _"iframeUserSettingsCtn"_ is later used in detail, but wouldn't a **display:none** make the other properties obsolete. I can see that in DevTools this line is marked with a ~strikethrough~. I'm just curious if there's something behind this that I don't know.

core/src/UserSettingsDialog.html
```diff
@@ -537,6 +472,16 @@
       z-index: 1;
     }
   }
+
+  /*need to be removed after updating FD Styles to 0.16.0, selected state for left side nav*/
```
ndricimrr, line 476:
```suggestion
  /* TODO: need to be removed after updating FD Styles to 0.16.0, selected state for left side nav */
```
just a suggestion. makes it easier to search later on.

ndricimrr:
Other than some small comments, LGTM 👍

core/src/UserSettingsDialog.html
UlianaMunich: thanks, leftovers from removing the scroll on load. Removed

core/src/UserSettingsDialog.html
```diff
@@ -303,6 +308,15 @@
       dispatch('close');
     }
   }
+
+  export function handleKeyUp(event, index) {
+    if (
+      event.keyCode == KEYCODE_ENTER ||
```
stanleychh, line 314:
```suggestion
      event.keyCode === KEYCODE_ENTER ||
```

core/src/UserSettingsDialog.html
```diff
@@ -303,6 +308,15 @@
       dispatch('close');
     }
   }
+
+  export function handleKeyUp(event, index) {
+    if (
+      event.keyCode === KEYCODE_ENTER ||
+      (KEYCODE_SPACE && event.keyCode !== KEYCODE_TAB)
```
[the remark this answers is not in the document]
UlianaMunich: thanks for pointing this out, somehow the brackets were in the wrong place. Fixed that

core/src/UserSettingsDialog.html
```diff
@@ -303,6 +308,15 @@
       dispatch('close');
     }
   }
+
+  export function handleKeyUp(event, index) {
+    if (
+      event.keyCode === (KEYCODE_ENTER || KEYCODE_SPACE) &&
```
ndricimrr, line 314: I think that `event.keyCode === (KEYCODE_ENTER || KEYCODE_SPACE)` will only be true for ENTER and not space. Maybe you meant sth like this instead.  
```suggestion
      (event.keyCode === KEYCODE_ENTER || event.keyCode === KEYCODE_SPACE) &&
```

ndricimrr (approved):

PR #1967, opened 2021-03-30
Limit max-width of Application title only on mobile

UlianaMunich:
Make limitation of the Applicatiion Title width only on mobile, on desktop it should stay as was always (full width)

image (8)


UlianaMunich:
App title screens bigger than mobile screen shouldn't occupy more than 50% of view width.

stanleychh (approved):

PR #1953, opened 2021-03-26
Fix context update bug

ndricimrr:
**Description**
Navigating inside of a micro-frontend (ex: Angular based) using its internal angular routing is facilitated using **withoutSync**. However, even though the navigation is successfully prevented with this function, there seems to be no context update upon internal navigation. 

**Changes**
- Have the Core [send a message to the iframe ](https://github.com/SAP/luigi/pull/1953/files#diff-33caff53e1abed9c033da822ec4bb4c2eee2ea754e6760968883ccf731bb8868R341-R348) when withoutSync is activated. An additional **withoutSync** parameter is added. The rest of the message body is the same.
- [Catch the event message on the Client side](https://github.com/SAP/luigi/pull/1953/files#diff-ea2419053062f053bee8d5e4a11c1df366970c29d50601d12ac2fe1ceebcb149R107-R117) in the lifecycle manager and handle it differently if withoutSync is activated:
    -  Omit navigation if withoutSync is present
    -  Add extra parameters to the context as useful information for the virtualTree structure

**How to test**
1. Open **e2e-test-application** and navigate to **projects/pr2/developers**
2. Click the withoutSync link and check the console for Context Changed print outs. 
3. The printed context should have **pathParams** and **viewUrl** as properties. 
 
**Related issue(s)**
Fixes #1940

PR #1893, opened 2021-02-16
Investigate withoutSync behavior under hashrouting enabled #1787

legteodav:
Fixes #1787

PR #1976, opened 2021-04-08
Disable browser history option

### LINKED ISSUE #1913: configuration option to disable browser history
Add an option to routing section that allows to prevent navigation changes to be added to browser history.

If this flag is set, a luigi navigation action should use history.replaceState instead of pushState (or setting hash)

JohannesDoberer:
Fixes #1913

### DIFF
--- a/core/src/services/routing.js
+++ b/core/src/services/routing.js
@@ -56,22 +56,17 @@ class RoutingClass {
     if (windowPath === GenericHelpers.trimLeadingSlash(route)) {
       return;
     }
+    const hashRouting = LuigiConfig.getConfigValue('routing.useHashRouting');
+    let url = new URL(location.href);
+    hashRouting ? url.hash = route : url.pathname = route;
 
-    if (LuigiConfig.getConfigValue('routing.useHashRouting')) {
-      if (!navSync) {
-        EventListenerHelpers.hashChangeWithoutSync = true;
-      }
-      window.location.hash = route;
-      return;
-    }
-
-    const method = pushState ? 'pushState' : 'replaceState';
+    const method = LuigiConfig.getConfigValue('routing.disableBrowserHistory') ? 'replaceState' : pushState ? 'pushState' : 'replaceState';
     window.history[method](
       {
-        path: route
+        path: hashRouting ? url.hash : decodeURIComponent(url.pathname)
       },
       '',
-      route
+      hashRouting ? url.hash : decodeURIComponent(url.pathname)
     );
 
     // https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent#Browser_compatibility
@@ -147,8 +142,8 @@ class RoutingClass {
     return LuigiConfig.getConfigValue('routing.useHashRouting')
       ? window.location.hash.replace('#', '') // TODO: GenericHelpers.getPathWithoutHash(window.location.hash) fails in ContextSwitcher
       : window.location.search
-      ? GenericHelpers.trimLeadingSlash(window.location.pathname) + window.location.search
-      : GenericHelpers.trimLeadingSlash(window.location.pathname);
+        ? GenericHelpers.trimLeadingSlash(window.location.pathname) + window.location.search
+        : GenericHelpers.trimLeadingSlash(window.location.pathname);
   }
 
   async handleRouteChange(path, component, iframeElement, config, withoutSync) {
@@ -174,7 +169,7 @@ class RoutingClass {
               this.handleRouteChange(path, component, iframeElement, config) &&
               history.replaceState(window.state, '', newUrl);
           },
-          () => {}
+          () => { }
         );
         return;
       }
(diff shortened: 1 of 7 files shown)


core/src/services/routing.js
```diff
@@ -56,22 +56,17 @@ class RoutingClass {
...
-    if (LuigiConfig.getConfigValue('routing.useHashRouting')) {
-      if (!navSync) {
-        EventListenerHelpers.hashChangeWithoutSync = true;
-      }
-      window.location.hash = route;
-      return;
-    }
-
-    const method = pushState ? 'pushState' : 'replaceState';
+    const method = LuigiConfig.getConfigValue('routing.disableBrowserHistory') ? 'replaceState' : pushState ? 'pushState' : 'replaceState';
```
stanleychh, line 63: I feel this line is not easy to understand. Could you please briefly explain the logic?

Does the logic same as the code below?
```
const method = LuigiConfig.getConfigValue('routing.disableBrowserHistory') && pushState ? 'pushState' : 'replaceState';
```

ndricimrr: If I remember right I think it goes like `const method = A && B ? C : D`
step 1 - check if A is true, if it is true return C - **end**
step 2 - A is false, then check B, if B is true return C, else return D -  **end**
but yeah this could be maybe broken up into pieces so it can be easier debugged I guess

ndricimrr: I could see this as more self-explanatory: 
```
const chosenHistoryMethod = pushState ? 'pushState' : 'replaceState';
const method = LuigiConfig.getConfigValue('routing.disableBrowserHistory') ? 'replaceState' : chosenHistoryMethod;
```
This would maybe not work in case **disableBrowserHistory** is not used and `pushState=true` by default, it would return **replaceState** instead of intended **pushState**
```
const method = LuigiConfig.getConfigValue('routing.disableBrowserHistory') && pushState ? 'pushState' : 'replaceState';
```
Maybe this would work as one-liner 🤔 ?
```
const method = LuigiConfig.getConfigValue('routing.disableBrowserHistory') || !pushState ? 'replaceState' : 'pushState'  ;
```

stanleychh: Yeah, not necessary to be one line if two lines of code are clearer.

JohannesDoberer: Go with the two-liner

core/src/utilities/helpers/routing-helpers.js
```diff
@@ -162,16 +162,12 @@ class RoutingHelpersClass {
...
-      });
-    }
-
     EventListenerHelpers.addEventListener('popstate', e => {
-      callback(Routing.getModifiedPathname(), e.detail && e.detail.withoutSync);
+      if (hashRoutingActive) {
+        callback(Routing.getHashPath(location.href), e.detail && e.detail.withoutSync);
+      } else {
+        callback(Routing.getModifiedPathname(), e.detail && e.detail.withoutSync);
+      }
```
stanleychh, line 170:
```suggestion



      const method = hashRoutingActive ? Routing.getHashPath(location.href): Routing.getModifiedPathname();
       callback(method, e.detail && e.detail.withoutSync);
```
This way is cleaner. `method` just an example, please feel free to change the naming.

core/src/services/routing.js
```diff
@@ -56,22 +56,18 @@ class RoutingClass {
     if (windowPath === GenericHelpers.trimLeadingSlash(route)) {
       return;
     }
+    const hashRouting = LuigiConfig.getConfigValue('routing.useHashRouting');
+    let url = new URL(location.href);
+    hashRouting ? url.hash = route : url.pathname = route;
 
-    if (LuigiConfig.getConfigValue('routing.useHashRouting')) {
-      if (!navSync) {
-        EventListenerHelpers.hashChangeWithoutSync = true;
```
ndricimrr, line 62: - I'm not sure if this part needs to be removed. This part was added as part of a bug fix [here](https://github.com/SAP/luigi/pull/1893/files) for the withoutSync issue that we had before. 
- Also, I see that luigi context update when navigating internally with Angular Router, doesn't seem to work as intended. You can follow the ***'How to test'*** part of [this issues](https://github.com/SAP/luigi/pull/1953) description for the steps on reproducing it. In my environment I do not get the context change anymore in the console logs.

JohannesDoberer: We need to sync on both issues, because I think first one is resolved in the new solution and second one works for me as well.

ndricimrr: I double checked and it seemed like I was using an outdated luigi bundle. It seems to work well for me as well. 👍

core/src/utilities/helpers/routing-helpers.js
stanleychh: @JohannesDoberer, please apply this improvement. The rest looks good.

ndricimrr (approved):
Looks good 👍

stanleychh (approved):

PR #1977, opened 2021-04-12
fix issue with styling active tab

UlianaMunich:
Fix a bug where a selected Tab was not visually highlighted on click. 

Styling reference: https://sap.github.io/fundamental-styles/?path=/docs/components-tabs--primary

### DIFF
--- a/core/src/navigation/TabNav.html
+++ b/core/src/navigation/TabNav.html
@@ -16,7 +16,7 @@
       key.indexOf(virtualGroupPrefix) === 0} {#each nodes as node, index2} {#if
       !node.hideFromNav} {#if node.label}
       <span
-        class="fd-tabs__item"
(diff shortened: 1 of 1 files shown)


core/src/navigation/TabNav.html
```diff
@@ -16,7 +16,7 @@
       key.indexOf(virtualGroupPrefix) === 0} {#each nodes as node, index2} {#if
       !node.hideFromNav} {#if node.label}
       <span
-        class="fd-tabs__item"
+        class="fd-tabs__item { (node === selectedNodeForTabNav) === true ? 'is-selected' : ''}"
```
stanleychh, line 19:
```suggestion
        class="fd-tabs__item { node === selectedNodeForTabNav ? 'is-selected' : ''}"
```
Would this way work properly?

UlianaMunich: Thanks, tested, and works fine.

stanleychh (approved):

PR #1986, opened 2021-04-19
Fix navigation bug with angular router

ndricimrr:
**Description**

- Fixes described issue on angular router navigation in 1.12.1

**Related issue(s)**


client/src/lifecycleManager.js
```diff
@@ -104,17 +104,11 @@ class LifecycleManager extends LuigiClientBase {
 
     helpers.addEventListener('luigi.navigate', e => {
       setContext(e.data);
-      if (!this.currentContext.internal.isNavigateBack && !this.currentContext.withoutSync) {
+      const withoutSync = this.currentContext.withoutSync === true;
+      if (!this.currentContext.internal.isNavigateBack && !withoutSync) {
```
stanleychh, line 108: I think we should avoid double negative boolean logic like this. e.g `!withoutSync`
Are we able to slightly improve it?

ndricimrr: true, refactored it 👍

ndricimrr:
Seems to be an issue with Angular. 
[Related Reference]( https://stackoverflow.com/a/41678403) -  _"When only route params or query params are updated but the route is the same, the component won't be destroyed and recreated"_.

PR #1987, opened 2021-04-19
Confirmation modal not visible

### LINKED ISSUE #1982: Modal cover up confirmation modal
Steps to reproduce:
Open a modal and within this modal open a confirmation modal.
Result:
Confirmation modal is not visible until you close the modal.
Expected Result:
Confirmation modal is visible and modal is behind the backdrop of confirmation modal.

Suggestion:
z-index

Side note: also test the solution with drawer.

JohannesDoberer:
Fixes #1982


test/e2e-test-application/src/app/project/drawer/drawer.component.ts
```diff
@@ -10,7 +10,11 @@ import { linkManager, uxManager } from '@luigi-project/client';
...
   public linkManager = linkManager;
   public uxManager = uxManager;
-  public constructor() {}
+  public constructor() { }
 
-  public ngOnInit() {}
+  public ngOnInit() { }
+
+  openConfirmationModal() {
+    uxManager().showConfirmationModal({ body: 'Just a confirmation modal' }).then(() => { console.log('test'); })
```
UlianaMunich, line 18: Just a tiny suggestion, maybe we could have some more explicit console.log text instead of "test". E.g. "the confirmation modal is shown"

core/src/ConfirmationModal.html
```diff
@@ -95,6 +95,7 @@ <h2 class="fd-title fd-title--h5">{settings.header}</h2>
   @import 'styles/variables';
 
   .fd-message-box {
+    z-index:1001;
```
stanleychh, line 98: We should manage all z-index in a central file. It's fine to do it in a separate ticket.
https://www.smashingmagazine.com/2021/02/css-z-index-large-projects/

UlianaMunich: Let's discuss it after daily and there is another z-index: 1100; in our code for alert overlay (.luigi-alert--overlay)

UlianaMunich (approved):
LGTM! All 4 scenarios

core/src/ConfirmationModal.html
JohannesDoberer: 👍  jup let's discuss in the daily.. but I like the idea to manage the z-index in a central file

PR #2003, opened 2021-04-30
Bump versions of Luigi client, core and auth to 1.12.1 in fiddle

UlianaMunich:
Fix alerts regarding versions in Fiddle: https://saas.whitesourcesoftware.com/Wss/WSS.html#!alertsReport


ndricimrr (approved):
👍

PR #2030, opened 2021-05-14
Fix setDocumentTitle bug

### LINKED ISSUE #2001: `Luigi.ux().setDocumentTitle` not updating document.title
**Description**

When document title is initially an empty string then setting title programicaly have no effect.

**Expected result**

Document title should be changed.

**Actual result**

Document title is not updated.

**Steps to reproduce**

Luigi Config:
```typescript
Luigi.setConfig({
  settings: {
    header: {
      title: '',
      altText: 'Brand name',
      logo: '/assets/images/logo.svg',
      favicon: '/assets/images/favicon.svg',
    }
  }
});
```

After application is loaded run `Luigi.ux().setDocumentTitle('xyz');` in console.

**Troubleshooting**

Found that title is not updating when header title is empty string.

stanleychh:
`Fixes #2001 `

It happens when no title was set at beginning only. `setDocumentTitle` function works fine when title is not empty.

### DIFF
--- a/core/src/navigation/services/header.js
+++ b/core/src/navigation/services/header.js
@@ -16,14 +16,15 @@ export const processHeaderSettings = component => {
           (component.get().appSwitcherItems && component.get().appSwitcherItems.length > 0)
       });
       return LuigiConfig.getConfigValueAsync('settings.header').then(header => {
+        if (header) {
+          component.set({ defaultTitle: header.title || '' });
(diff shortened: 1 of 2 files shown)


JohannesDoberer (approved):
Works as expected 👍 .
One thing: If no header in config is defined and you call setDocumentTitle and getDocumentTitle, you can see (on browser console) that document title is set but nothing will change on document title in browser. Maybe we can improve it and print a warning if no header is set.

PR #2049, opened 2021-05-25
Extend options of enum type in user settings

### LINKED ISSUE #2041: Extend enum type in user settings
Extend enum type in user settings that key values are possible for options.

Example:
```
enum:
          {
            type: 'enum',
            label: 'Label',
            options: [{en: 'English'}, {de: 'German'}, {es:'Spanish'}, {pl: 'Poland'}],
            description: 'Description'
          },
```

JohannesDoberer:

- options array of type enum in user settings can also have objects as entries

Fixes #2041


ndricimrr:
@JohannesDoberer 
 Could you run in the root folder : 
`npm run full-code-quality-prettier` 
Might need to `npm install` if not done previously. 
and just commit the result?

There are some non-prettified stuff in the PR, or I think it is prettified differently from what is expected, and many changes are showing

core/src/UserSettingsEditor.html
```diff
@@ -152,8 +152,8 @@
                   <input
                     type="text"
                     class="fd-input fd-input-group__input"
-                    placeholder="{$getTranslation(storedUserSettingData[userSettingGroup[0]][key])}"
                     data-testid="lui-us-input{i}"
+                    placeholder="{GenericHelpers.isObject(storedUserSettingData[userSettingGroup[0]][key])?$getTranslation(Object.values(storedUserSettingData[userSettingGroup[0]][key])[0]):$getTranslation(storedUserSettingData[userSettingGroup[0]][key])}"
```
ndricimrr, line 156: This line seems quite long.
I'm thinking of two options here: 
- Do you think it would make sense to extract this into a **GenericHelpers. getUserSettingsFromhmhmhm...........**
- Maybe, that's too much but, what about breaking this line up into the condition and the two options, it would maybe be bit better than the whole line. Sth similar or if you have a better split in mind:
```
// I didn't test this
{GenericHelpers.isObject(storedUserSettingData[userSettingGroup[0]][key]) ? 
        $getTranslation(Object.values(storedUserSettingData[userSettingGroup[0]][key])[0])
        :$getTranslation(storedUserSettingData[userSettingGroup[0]][key])} 
```

I checked and **Prettier** is not able to cut long lines all the time when the line is too long, automatically. Maybe because it is just a string. So, I think Prettier won't complain if you split this to 2-3 lines.

ndricimrr (commented):
The functionality seems to be working fine for me 👍 . Added one code suggestion if you think it makes sense. 

Also, I see that the get functions are returning a string instead of the object for both client and core api functions, when u chose an object as language:
Screen Shot 2021-05-26 at 19 24 29

If you check the   
https://docs.luigi-project.io/docs/luigi-core-api/?section=getcurrentlocale
> getCurrentLocale
> Gets the current locale.
> Returns **string** current locale

So, maybe we also need to modify these two functions to return either Object or string in this PR or in a followup ticket?

core/src/UserSettingsEditor.html
```diff
@@ -78,6 +78,14 @@
 
     storedUserSettingData[userSettingGroup[0]][key] = option;
   }
+
+  function getPlaceholderValue(optionsEntry) {
+    if (GenericHelpers.isObject(optionsEntry)) {
+      return $getTranslation(Object.values(optionsEntry)[0]);
+    } else {
+      return $getTranslation(optionsEntry);
+    }
```
stanleychh, line 87:
```suggestion
    const value = GenericHelpers.isObject(optionsEntry) ? Object.values(optionsEntry)[0] : optionsEntry;
    
     return $getTranslation(value);
```

core/src/UserSettingsEditor.html
```diff
@@ -78,6 +78,12 @@
 
     storedUserSettingData[userSettingGroup[0]][key] = option;
   }
+
+  function getPlaceholderValue(optionsEntry) {
+    const value = GenericHelpers.isObject(optionsEntry) ? Object.values(optionsEntry)[0] : optionsEntry;
+    
+     return $getTranslation(value);
+  }
```
ndricimrr, line 86: This is just a prettier auto formatting. It would be nice to add. Code looks good to me, works well 👍   
```suggestion

  function getPlaceholderValue(optionsEntry) {
    const value = GenericHelpers.isObject(optionsEntry)
      ? Object.values(optionsEntry)[0]
      : optionsEntry;

    return $getTranslation(value);
  }
```

ndricimrr (commented):
Tested again and it works well. Code looks good

ndricimrr (approved):

ndricimrr (commented):
Looks good to me. 

Please run `npm run full-code-quality-prettier` on root folder as there are many non-prettified changes.

alexandra-simeonova (approved):

PR #2095, opened 2021-06-16
#1981 | Add role button and tabindex 0 to Logo

### LINKED ISSUE #1981: Can't focus logo
**Description**

Even if logo is clickable I can't focus it using keyboard.

**Expected result**

Logo should be an button, anchor or have `role="button" tabindex="0"`

**Actual result**

Can't focus logo

**Steps to reproduce**

Press tab button and try to focus logo

**Troubleshooting**

rafalgamon:
**Description**

- #1981 | Add role button and tabindex 0 to Logo

**Related issue(s)**
`Fixes #1981 `


UlianaMunich (changes requested):
Please move new attributes instead of `` into `<a>`. And ensure the `role="button"` doesn't break the old styling of Logo and App Title. Maybe adding `tabindex="0"` would be already enough?

UlianaMunich (approved):

rafalgamon:
On Firefox, type "about:config" in the URL bar. There is no accessibility.tabfocus preference on the mac os, so you'll have to make one. Right click in the window, create a new "integer" pref, and set it to 7
But this is changing inside browser config not in our config/app.
When you set this tabfocus to 7 it works same as on Chrome ane Edge.
On Safari focus works totally different as on Chrome, Firefox and Edge.  
Of course changing tabfocus setting is tested on Mac OS and as I know this issue is only on Mac OS  but someone who has Windows should check it to be totally sure. Code which actual on this branch is correct and ready to test on Windows.

JohannesDoberer:
Safari:
```
Testing on a Mac: System Preferences -> Keyboard -> ShortCuts (tab) -> Full Keyboard Access -> All Controls

For Tabbing to work on Safari: Preferences -> Advanced -> Press tab to highlight each item on a page (check this)
```
If everything is enabled it works like a charm.

FF:
It works for me also without setting the `accessibility.tabfocus` property on both, osx and win.
Only the border is not visible but if you have focus on logo and press enter the page will be reloaded.

PR #2113, opened 2021-06-24
2097 | Current 'rem' unit for all App is set in global browser settings

rafalgamon:
- Changed global app font size settings. Removed hardcode 16px from App.html styles


UlianaMunich (approved):
LGTM!

PR #2129, opened 2021-07-01
2087: Disable left nav animation when switching from/to mfe with hideSideNav

### LINKED ISSUE #2087: disable left nav animation when switching from/to mfe with hideSideNav
slide-in/-out animation should only be activated for manual expand/collapse trigger.

Probably the easiest solution:
Add the following selectors to css

`

		.lui-semiCollapsible #app.no-side-nav .fd-app__sidebar,
		.lui-semiCollapsible #app.no-side-nav .iframeContainer, 
		.lui-semiCollapsible #app.no-side-nav .spinnerContainer {
			-webkit-transition: none;
			-moz-transition: none;
			-ms-transition: none;
			-o-transition: none;
			transition: none;
		}

`

wdoberschuetz:
**Description**
Do not animate hiding of the leftSideNav when it's been done by user clicking/programmatically.

- #2087

**Related issue(s)**
Resolves #2087


core/src/App.html
```diff
@@ -1,5 +1,8 @@
 <svelte:window on:resize="{onResize}" />
-<div id="app" class="{hideNav? 'no-nav' : ''} {hideSideNav? 'no-side-nav':''}">
+<div
+  id="app"
```
JohannesDoberer, line 3: In general we try to avoid to use `id`'s on html elements, because we offer a framework and don't want to "block" ids  for customers project. If the customer use the same id he will get unforeseen side effects. Especially`app` is most likely an id which will be used by someone else.

rafalgamon (approved):

core/src/App.html
wdoberschuetz: You probably noticed that I've maintained the status quo. 
Apart from me, ```app``` was already used by some other CSS. :)
I was assuming that this id is used for some booting but I can only see one reference to ```app``` and it is in the vue.js example - would it use the ```id="app"``` in ```App.html``` to boot?
Shall I refactor or would this be another ticket?

Issue #2101, opened 2021-06-18
visibleForFeatureToggles should result in node not being present


hardl:
Currently, if a root node is hidden behind a feature toggle, it is not visible in the navigation but it is still accessible via its route. This should be prevented.
For non-root nodes everything works fine.

PR #2133, opened 2021-07-05
#2101 | visibleForFeatureToggles Should block main path when we chang…

rafalgamon:
- visibleForFeatureToggles Should block main path when we change config of main root


stanleychh:
@rafalgamon , I closed it and created new one because this PR contains few changes under /dist/ folder. Those look unnecessary.

Issue #2116, opened 2021-06-29
'nav href' for all navigation controls


hardl:
Setting the property "addNavHrefs" to true allows the user to right-click->open-in-new-tab on navigation entries in the left side nav.
This should also be applied to entries in 
- [x] global nav
- [x] top nav 
- [x] profile dropdown
- [x] application switcher
- [x] product switcher
- [x] context switcher

PR #2154, opened 2021-07-14
2118: Add addNavHref support for GlobalNav

wdoberschuetz:
**Description**

#2118 

**Related issue(s)**
#2116


stanleychh:
Already discussed issues with @wdoberschuetz.
Besides a failed Cypress test.
When right-click on a `GlobalNav` node and try to `Open Link in New Tab/Window`, it did not work. Browser tab shows `about:blank#blocked`. (at simpledev env.)

core/src/navigation/GlobalNav.html
```diff
@@ -10,6 +10,26 @@
           class="fd-nested-list__item {node === selectedNode ? 'is-selected' : ''}"
           data-testid="{getTestId(node)}"
         >
+          {#if LuigiConfig.getConfigBooleanValue('navigation.addNavHrefs')}
+          <a href="{getRouteLink(node)}" title="{$getTranslation(node.label)}">
+            <div class="lui-fd-nested-list__content" on:click="{() => handleClick(node)}">
```
stanleychh, line 15: When user clicks on a global menu. Which one handle the redirection?  The onClick event of div or the hyperlink of a tag?

test/e2e-test-application/e2e/tests/1-angular/navigation.spec.js
```diff
@@ -1002,8 +1002,11 @@ describe('Navigation', () => {
           cy.get('[data-testid="settings_settings"]')
             .should('exist')
             .not('.is-selected')
-            .click()
-            .should('have.class', 'is-selected');
+            .click();
+
+          // select global nav node again to wait for the rerendering of the element
+          cy.get('[data-testid="settings_settings"]').should('have.class', 'is-selected');
```
stanleychh, line 1008: 👍

core/src/navigation/GlobalNav.html
UlianaMunich: Just adding on top to check whether the `on:click ` needs to be extended with `on:click|preventDefault` to avoid full page reload on click. Remember this issue when was working on a similar task.

UlianaMunich: From the LogoTitle.html it might be possible to move `on:click` under the `<a></a>` without breaking anything. @wdoberschuetz , please check. Thanks

wdoberschuetz: @stanleychh The ```on:click``` handles the redirection before ```href``` can do it.

@UlianaMunich Thanks, I've used your suggestions :).

test/e2e-test-application/src/luigi-config/extended/settings.js
```diff
@@ -57,11 +57,12 @@ class Settings {
...
-  //    profileMenuFiori3: true,
-  //    globalNav: true,
-  //   webcomponents:true
-  // };
+  experimental = {
+    //    profileMenuFiori3: true,
+    globalNav: true
+    //   webcomponents:true
+  };
+  globalSideNavigation = true;
```
UlianaMunich, line 65:
```suggestion
  //    experimental = {
    //    profileMenuFiori3: true,
    //    globalNav: true
    //   webcomponents:true
  };
//    globalSideNavigation = true;
```

core/src/navigation/GlobalNav.html
stanleychh: @wdoberschuetz , but do we need to have one on `div` and the other on `<a>`? I guess even without `on:click` event , only keep `href`. It is still working fine. 
@UlianaMunich , do you have any thought about this ?

wdoberschuetz: I made it this way, to be able to deduplicate the part in the ```div``` which is the same here and lower in the file as well.

stanleychh (approved):

stanleychh (approved):
Checked it and worked properly.

stanleychh:
@wdoberschuetz , Please merge it directly after Travis CI is happy.

PR #2168, opened 2021-07-26
2120: Add navHref possibility for profile dropdown

wdoberschuetz:
- Add navHref support for the profile dropdown

**Related issue(s)**
#2120 AddNavHref support for profile dropdown


UlianaMunich (changes requested):
Please update href attribute for every custom entity of the profile dropdown such as: Login/Lgout HTML element, user settings. 

In addition to it, please switch Profile Menu to use "Fiori3" profileType and Authorization.html as result needs to be updated as well.

core/src/Authorization.html
```diff
@@ -76,6 +76,7 @@
         class="fd-list__link"
         aria-label="User Settings"
         data-testid="settings-link"
+
```
UlianaMunich, line 79:
```suggestion
        
```

UlianaMunich (approved):

PR #2172, opened 2021-07-29
Add possibility to enable product switch to be opened in new Tab 

### LINKED ISSUE #2122: addNavHref support for product switcher

UlianaMunich:
Fixes #2122


core/src/navigation/ProductSwitcher.html
```diff
@@ -7,8 +7,26 @@
   <div class="fd-product-switch">
     <div class="fd-popover fd-popover--right">
       <div class="fd-popover__control" on:click|stopPropagation="{() => {}}">
-        <!-- default: sap-icon--grid -->
-
+        {#if LuigiConfig.getConfigBooleanValue('navigation.addNavHrefs')}
+        <a
+          href="/"
```
ndricimrr, line 12: Just wondering if this is necessary or redundant, I would not usually right click on this one to get home. But maybe that's some kind of standard ❔

ndricimrr (commented):
All looks fine to me. Working properly. Just some small comments. Otherwise approve 👍

core/src/navigation/ProductSwitcher.html
ndricimrr: On a side note, should we think about having this enabled on mobile version as well, or is that not relevant 🤔

core/src/navigation/ProductSwitcher.html
```diff
@@ -40,6 +60,36 @@
               on:click="{() => onActionClick(productSwitcherItem)}"
               data-testid="{getTestId(productSwitcherItem)}"
             >
+              {#if LuigiConfig.getConfigBooleanValue('navigation.addNavHrefs')}
+              <a
+                href="{addNavHrefForAnchor ? getRouteLink(productSwitcherItem) : undefined}"
```
ndricimrr, line 65:
```suggestion
                href="{getRouteLink(productSwitcherItem)}"
```

**addNavHrefForAnchor** seems like a good idea, but might be redundant. 

I can see that **getRouteLink** also does the check too internally through **getNodeHref**. Seems like we added it long time ago:
getRouteLink -> RoutingHelpers.getNodeHref -> 

```
  getNodeHref(node, pathParams) {
    if (LuigiConfig.getConfigBooleanValue('navigation.addNavHrefs')) {
    ....
```

so maybe we could avoid this new variable here altogether.

ndricimrr (approved):
Tested again. Looks good 👍

PR #2177, opened 2021-08-03
Enable addNavHrefs to be applied to the App Dropdown in TopNav

### LINKED ISSUE #2121: addNavHref support for application switcher

UlianaMunich:
Fixes #2121

### DIFF
--- a/core/src/navigation/LogoTitle.html
+++ b/core/src/navigation/LogoTitle.html
@@ -116,7 +116,7 @@ LuigiConfig.getConfigBooleanValue('navigation.addNavHrefs')}
             role="button"
             class="fd-menu__link"
             on:click|preventDefault="{() => goTo(item.link)}"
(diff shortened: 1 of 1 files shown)


core/src/navigation/LogoTitle.html
```diff
@@ -84,7 +116,7 @@
             role="button"
             class="fd-menu__link"
             on:click|preventDefault="{() => goTo(item.link)}"
-            href="{getRouteLink(item, pathParams)}"
+            href="{addNavHrefForAnchor ? getRouteLink(item, pathParams) : undefined}"
```
ndricimrr, line 119: I commented on a very similar case [here](https://github.com/SAP/luigi/pull/2172/files#r683521273) that **addNavHrefForAnchor** could be redundant. Since I'm seeing it here as well, maybe that's due to some reason I am not aware of?

ndricimrr (approved):
Looks good 👍

PR #2183, opened 2021-08-05
A configurable tooltipText parameter has been added

rafalgamon:
**Description**
- A configurable tooltipText parameter has been added


core/src/navigation/LeftNav.html
```diff
@@ -48,7 +48,7 @@
           <li class="fd-nested-list__item">
             <a
               href="{getRouteLink(node)}"
-              title="{$getTranslation(node.label)}"
+              title="{node.tooltipText ? $getTranslation(node.label) : ''}"
```
ndricimrr, line 51: What is the general idea behind the new `tooltipText` we are adding , or how is it different from the `label`. 

Unless something else was planned, it would be nice to maybe have this added in our documentation section: [navigation-parameters section](https://docs.luigi-project.io/docs/navigation-parameters-reference?section=navigation-parameters).

docs/navigation-parameters-reference.md
```diff
@@ -96,6 +96,12 @@ The navigation parameters allow you to configure **global** navigation settings
 - **description**: renders all views in new frames. This setting overrides the same-domain frame reuse.
 - **default**: the parameter **defaults.isolateView** is `false` by default, and you can overwrite it using the **isolateView** value on a single node level.
 
+### tooltipText
+- **type**: string
+- **description**: Add tooltip to link/button of navigation.
```
ndricimrr, line 101: This looks much better with documentation now 👍 . 

I would propose a slightly different sentence as this looks quite shallow, although I'm still struggling to understand the context overall. 

Feel free to reply to my initial comment above, or let's discuss this at a daily. Could be something I am getting wrong.

core/src/navigation/LeftNav.html
```diff
@@ -90,7 +90,7 @@
             <div class="fd-nested-list__content has-child">
               <a
                 href="javascript:void(null)"
-                title="{$getTranslation(key)}"
+                title="{nodes.tooltipText ? $getTranslation(nodes.tooltipText) : ''}"
```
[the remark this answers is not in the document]
rafalgamon: I check it and there (in this one example) has to be `nodes. tooltipText`  but in every other case we have to use `node.tooltipText`. 
And if talk about tests I fixed it.

docs/navigation-parameters-reference.md
rafalgamon: I propose  here `**description**: Add tooltip to navigation element.  If param is set tooltip will show if not tooltip will stay hide. `

docs/navigation-parameters-reference.md
```diff
@@ -96,6 +96,12 @@ The navigation parameters allow you to configure **global** navigation settings
 - **description**: renders all views in new frames. This setting overrides the same-domain frame reuse.
 - **default**: the parameter **defaults.isolateView** is `false` by default, and you can overwrite it using the **isolateView** value on a single node level.
 
+### tooltipText
+- **type**: string
+- **description**: Add tooltip to all of the links and buttons of the left side navigation and top navigation.
+- **default**: the parameter **defaults.tooltipText** is `null` by default, and you can overwrite it using the **tooltipText** value on a single node level.
```
alexandra-simeonova, line 102:
``````suggestion
- **description**: allows you to add a tooltip to all of the links and buttons of the left side navigation and top navigation. You can set the parameter with the text that will appear on the tooltip. 
- **example**:
```javascript
tooltipText: 'Useful links'
```
``````
Maybe it's clearer if we add an example on how to use it?

File: docs/navigation-parameters-reference.md [hunk 5]
```diff
@@ -96,6 +96,12 @@ The navigation parameters allow you to configure **global** navigation settings
...
+
```
alexandra-simeonova, line 103: I'm not sure if `defaults.tooltipText` is a separate parameter, for example like [isolateView](https://docs.luigi-project.io/docs/navigation-advanced/?section=isolateview) and [deafults.isolateView](https://docs.luigi-project.io/docs/navigation-advanced/?section=defaultsisolateview). In that case, I think it would be better to write a separate description for it 🙂 But I'm not sure what that description should be/what it does, @ndricimrr what do you think? 
For example, 
```
### defaults.tooltipText
- **type**: string
- **description**: <!-- ??? sets the default tooltip text for every navigation node ???  --> 
- **default**: the parameter **defaults.tooltipText** is `null` by default, and you can overwrite it using the **tooltipText** value on a single node level.
```

rafalgamon: Ok but now we have `Add tooltip to all of the links and buttons of the left side navigation and top navigation.`  so you suggest  `sets the default tooltip text for every navigation node`? @aleksandr-suvorov

alexandra-simeonova: Right, the question is if `tooltipText` adds it to *all* the navigation nodes or just one? 
As far as I understand, `defaults.tooltipText` adds a tooltip to all the nodes, and `tooltipText` to individual nodes.
Thanks, I changed my suggestion above to reflect this.

alexandra-simeonova (approved):

docs/navigation-parameters-reference.md
rafalgamon: But `defaults.tooltipText` will not have sense. The idea was that the client could add any tooltip himself, e.g. when the length of the label exceeds the width of the button / navigation.

docs/navigation-parameters-reference.md
```diff
@@ -96,6 +96,20 @@ The navigation parameters allow you to configure **global** navigation settings
...
+- **example**:
+```javascript
+tooltipText: 'Useful links'
+```
+
+### defaults.tooltipText
+- **type**: string
+- **description**: allows you to add a tooltip to **all** the links and buttons of the left side navigation and top navigation.
+- **default**: the parameter **defaults.tooltipText** is `null` by default, and you can overwrite it using the **tooltipText** value on a single node level.
+
```
ndricimrr, line 111:
```suggestion
```
(cc: @alexandra-simeonova) If I'm not mistaken @rafalgamon I  believe users cannot define navigation.**defaults.tooltipText**. So this field **defaults.tooltipText** is not valid.  I know that @rafalgamon might have intended to write about its default value. By default, it is inferred that if you don't set this property, you just get no tooltip 👍

ndricimrr (approved):
Looks good 👍

ndricimrr (changes requested):
I notice that there are some **faulty** auto-generated documentation changes here as well. 
We wouldn't want those in `master` as it would break docu website. 

It should be enough to discard all changes to these 2 files:
`docs/luigi-client-api.md`
`docs/luigi-core-api.md`

[This](https://github.com/SAP/luigi/pull/2223) will already prohibit them from being generated again.

PR #2190, opened 2021-08-09
Add newTab option to linkManager

### LINKED ISSUE #2188: Add linkManager option to navigate to new tab
**Description**
Add a new option to linkManager to enable navigating to a new tab. 
eg.:
`linkManager.openInNewTab().navigate('/some-where') `
should open '/some-where' in a new tab. 

option should be disabled by default. 

**Reasons**
Need for the option rises from the [PR](https://github.com/SAP/luigi/issues/2184) about extending intent based navigation.

ndricimrr:
**Description**

- add new client option to allow opening paths in new window through Luigi Client linkManager option.
```
   LuigiClient.linkManager()
  .newTab()
  .navigate('/projects/xy/foobar');
```

**Related issue(s)**

Resolves #2188

### DIFF
--- a/core/src/App.html
+++ b/core/src/App.html
@@ -1244,7 +1244,7 @@
 
       if ('luigi.navigation.open' === e.data.msg) {
         isNavigateBack = false;
-        if (e.data.params.openInNewTab) {
+        if (e.data.params.newTab) {
           let path = buildPath(e.data.params);
           path = GenericHelpers.addLeadingSlash(path);
           openViewInNewTab(path);
--- a/core/src/App.html
+++ b/core/src/App.html
@@ -918,8 +918,7 @@
   resetMicrofrontendModalData();
 
   const openViewInModal = async (nodepath, settings) => {
-    const { nodeObject } = await Navigation.extractDataFromPath(nodepath);
-    if (await Navigation.shouldPreventNavigation(nodeObject)) {
+    if (await NavigationHelpers.shouldPreventNavigationForPath(nodepath)) {
       return;
     }
     mfModal.displayed = true;
@@ -981,8 +980,7 @@
   resetMicrofrontendDrawerData();
 
   const openViewInDrawer = async (nodepath, settings) => {
-    const { nodeObject } = await Navigation.extractDataFromPath(nodepath);
-    if (await Navigation.shouldPreventNavigation(nodeObject)) {
+    if (await NavigationHelpers.shouldPreventNavigationForPath(nodepath)) {
       return;
     }
     mfDrawer.displayed = true;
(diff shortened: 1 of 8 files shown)


docs/luigi-client-api.md
```diff
@@ -386,28 +387,45 @@ LuigiClient.linkManager().withoutSync().fromClosestContext().navigate('settings'
...
+
+##### Examples
+
+```javascript
+LuigiClient.linkManager().newTab().navigate('/projects/xy/foobar');
+```
+
+**Meta**
+
+*   **since**: NEXT_RELEASE
```
ndricimrr, line 404: * Added documentation that might need review.

core/src/App.html
```diff
@@ -1045,12 +1045,25 @@
     internalUserSettingsObject.displayed = false;
   };
 
+  // Open View in New Tab
+
+  const openViewInNewTab = async nodepath => {
+    const { nodeObject } = await Navigation.extractDataFromPath(nodepath);
+    if (await Navigation.shouldPreventNavigation(nodeObject)) {
+      return;
+    }
```
stanleychh, line 1054: It seems this part is repeated as function `openViewInDrawer` and `openViewInModal`. Are we able to extract to a single function?

stanleychh (commented):
Overall looks good, well done 👍. 
Left a minor comment only.

core/src/App.html
ndricimrr: I notice that 🤔 . I am trying to refactor it though, but it is so far seeming like we don't end up saving a lot of lines 😄 . If you compare **old usage** vs. **new usage** below seems like there's only one single line difference:
```
    // old usage
    const { nodeObject } = await Navigation.extractDataFromPath(nodepath);
    if (await Navigation.shouldPreventNavigation(nodeObject)) {
      return;
    }
```

```
// new function
shouldPreventNavigation(nodepath) {
    const { nodeObject } = await Navigation.extractDataFromPath(nodepath);
    if (await Navigation.shouldPreventNavigation(nodeObject)) {
      return true;
    }
    return false;
}

// new usage
if (shouldPreventNavigation(nodepath)) {
      return;
}
```
Not sure if it is worth the change. Maybe you have a better suggestion than my snippet above, that would be nice. Otherwise I would leave it as is. What do you think ?

stanleychh: Cool. We still definitely should go for the new approach. Saving lines is the bonus, not the first priority in this case.
The point is to put this logic in a single place. The benefit will be seen whenever we need to update the business logic.

ndricimrr: Added changes 👍

core/src/utilities/helpers/navigation-helpers.js
```diff
@@ -305,6 +305,18 @@ class NavigationHelpersClass {
     delete strippedNode.navHeader;
     return strippedNode;
   }
+
+  /**
+   * Checks if for the given node path navigation should be prevented or not
+   * @param {string} nodepath path to check
```
stanleychh, line 311:
```suggestion
   * @param {string} nodepath path to check
   * @returns {boolean} navigation should be prevented or not
```

stanleychh (approved):
Looks good 👍 
Please resolve file conflicts.

alexandra-simeonova (approved):

PR #2210, opened 2021-08-11
Fix accessibility issues

### LINKED ISSUE #2191: Accessibility Issues
**Description**

Three bugs of accessibility issue. 
DXP ticket: https://zenhub.tools.sap.corp/app/workspaces/developer-portal-5f8da7225f55a70faa8323d5/issues/dxp/jukebox/690 

**Issues**

1. Global side navigation can be reached via "Tab"-key. However, after pressing "Enter"-key on a global side navigation item, it will refresh the whole browser.  e.g. https://sap.dev.dxp.k8s.ondemand.com/?ft=home

2. User profile menu cannot be opened via pressing "Enter"-key
3. User profile menu opened via mouse → tab to "settings" entry → not possible to open settings via
"Enter"-key 

**Expected result**

1. Update the micro-frontend only. (Same behavior via mouse clicking)
2. Able to open profile dropdown via pressing "enter"-key
3. Able to open "settings" dialog via pressing "enter"-key

stanleychh:
**Description**
Fixed accessibility issues.

**Issues**
- Not able to navigate to "My Settings" and "Sign Out" via `Tab` key, also not able to trigger event via `Enter` key
- The whole browser reloaded when press `Enter` key at a selected global navigation item.

**Solution**
Use semantic elements directly. The browser will take care of accessibility.
e.g.  If the element is clickable, use `button` instead of `div`/`span`/`a` with `role='button'`

**Accessibility Knowledge**
[WAI-ARIA Authoring Practices 1.1](https://www.w3.org/TR/wai-aria-practices-1.1/)
[button_role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role)

**Related issue(s)**
Fixes #2191


UlianaMunich (approved):
LGTM!

PR #2221, opened 2021-08-18
LuigiClient linkManager wrong navigation scope in modal

### LINKED ISSUE #2211: LuigiClient linkManager wrong navigation scope in modal
LuigiClient linkManager is using wrong navigation scope for relative path navigation when called from a a MFE within a modal.

It seems that the route from the MFE that is currently open in the normal content area is used instead.

stanleychh:
**Description**
When navigate to a relative path from an opened micro-frontend in a modal. 
The navigated path should go from the URL of the modal.

- Check the scope of current active iframe. Relative navigation should go from there.

**Related issue(s)**
Fixes #2211

### DIFF
--- a/core/src/App.html
+++ b/core/src/App.html
@@ -165,7 +165,7 @@
   export let getTranslation;
 
   let showLoadingIndicator = false;
-
+  let currentIframe;
   let mfSplitView = {
     displayed: false
   };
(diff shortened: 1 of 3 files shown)


JohannesDoberer (approved):
LGTM!

PR #2231, opened 2021-08-26
e2e statistic test coverage

rafalgamon:
**Description**
- e2e statistic test coverage


stanleychh:
Generally looks good. Good job.
Could you please provide the documentation about how to run it? Where could we get the coverage report? 
That will be awesome. Thanks

rafalgamon:
All the logic that generates the report is ready.
The problem is that the stats are only generated for the e2e folder and not for the core. From what I was able to find on the internet, this solution can be in the files that are includes and excludes in the nyc configuration. But I try it and still I don't have any working solution.

rafalgamon:
I have added a new configuration to 'nyc'. At the moment it generates a report for 'core' but unfortunately it's still 0/0% everywhere. Maybe someone else will see it.

JohannesDoberer:
General docu https://docs.cypress.io/guides/tooling/code-coverage#Introduction
 -> https://docs.cypress.io/guides/tooling/code-coverage#Instrumenting-code
 -> https://github.com/istanbuljs/nyc/blob/master/docs/instrument.md
Instrumenting code is only possible inside `/test/e2e-test-application` and not outside of the root of e2e...
It looks like we cannot test the coverage of luigi core.

wdoberschuetz:
I wanted to check if something's changed and I found this: https://stackoverflow.com/questions/53399098/enabling-nyc-istanbul-code-coverage-for-files-outside-the-package-directory

It's not new but have any of you looked at ```cwd``` before? I had a go at it and couldn't make it work but it seems to be what we need (?).

PR #2239, opened 2021-08-30
Update routing.js

### LINKED ISSUE #2238: wrong hideSideNav inheritance
Bug description:
setting `hideSideNav` to true at a nested node does not have any effect, if any of the ancestor nodes has it set explicitly to false.

Expected behaviour:
setting `hideSideNav` to true at a nested node should hide the side nav, if this node is activated, regardless of its ancestor node configuration.

hardl:
fixes #2238


wdoberschuetz (approved):
Works as expected in the description of the bug this PR resolves.

When a child node (nested node) has ```hideSideNav``` to ```true``` when its parent is set to ```false```, then the parent's setting is ignored as expected.

Issue #2160, opened 2021-07-20
Allow to get rid of tooltips in the menu which have the same text as menu item


Sawthis:
**Description**

We can't get rid of tooltips in the menu which have the same text as menu item. They are added here by Luigi [here](https://github.com/SAP/luigi/blob/master/core/src/navigation/LeftNav.html#L51), and the have always the same text as the immediately visible label.

**Reasons**

We want to unify tooltips in our app and remove these tooltips which are redundant in the menu. We have different style of tooltips in our app.

**Attachments**

hardl:
Implementation solution: we add a new node property 'tooltipText'.
If present, it will be used as title attribute for the node entry.
If empty string or explicitly set to false, no title attribute will be rendered.
It will also be added to our "node defaults" mechanism (see e.g. https://docs.luigi-project.io/docs/navigation-advanced/?section=defaultsisolateview) in order to change the default behaviour for your application.

PR #2242, opened 2021-08-31
Add tooltipText param to TopNav and LeftNav

rafalgamon:
**Description**

- Add tooltipText param to TopNav and LeftNav


alexandra-simeonova (approved):

core/src/navigation/LeftNav.html
```diff
@@ -56,7 +63,11 @@
           <li class="fd-nested-list__item">
             <a
               href="{getRouteLink(node)}"
-              title="{$getTranslation(node.label)}"
+              title="{
+                (node.tooltipText && node.tooltipText.length) ?
+                $getTranslation(node.tooltipText) :
+                (node.tooltipText === false) ? '' : $getTranslation(node.label)
```
ndricimrr, line 69: `(node.tooltipText === false) ? '' : $getTranslation(node.label)`
Worked fine for me, but I think that the original [Issue](https://github.com/sap/luigi/issues/2160#issuecomment-885539914) required **empty string** or **false** for disabling the tooltip. Here it seems like you are not checking for empty string, maybe should make it an **or** statement instead ? 

> If empty string or explicitly set to false, no title attribute will be rendered. [comment](https://github.com/sap/luigi/issues/2160#issuecomment-885539914)

ndricimrr (changes requested):
Worked fine 👍 . 
- However, noticed that it doesn't disable tooltips if tooltipText set to empty string. 

- Other thing is `defaults.tooltipText` is also missing. If I got it right from [Philipps comment](https://github.com/sap/luigi/issues/2160#issuecomment-885539914) , one should be able to define. 
```
      navigation: { 
            defaults: {
                  tooltipText: false
            },
            ...
       }
```
Not sure on details of the defaults mechanism (boolean vs string), but can discuss this in daily.

core/src/navigation/LeftNav.html
ndricimrr: yep, something like that 👍

core/src/navigation/LeftNav.html
```diff
@@ -201,7 +224,11 @@
           <!-- Category nodes -->
           <li
             class="fd-nested-list__group-header lui-category"
-            title="{$getTranslation(key)}"
+            title="{
+              (nodes.tooltipText && nodes.tooltipText.length) ?
+              getTooltipText(nodes) :
```
ndricimrr, line 229: I find a flaw in this logic. 
```
if ( nodes.tooltipText && nodes.tooltipText.length ) {
    //nodes.tooltipText is definitely not false or empty here
    // the following 'getTooltipText' might never return 'defaultsTooltipTextConfig' as 'node.tooltipTex' will initially always be true, never reaching second one.
     getTooltipText(nodes) => return node.tooltipText || defaultsTooltipTextConfig; =
} else {
// (nodes.tooltipText === false || nodes.tooltipText === '') ? '' : $getTranslation(key)
}
```
Due to that I also don't get the **defaults.tooltipText** affecting the tooltips at all if I add it in the luigi configurations, while testing locally. I believe this might be the reason 🤔 . Does it work for you ?

This would be made much easier if **defaults.tooltipText** was documented, so I could also know what is the intended result.

test/e2e-test-application/e2e/tests/1-angular/navigation.spec.js
```diff
@@ -570,6 +571,19 @@ describe('Navigation', () => {
...
+      cy.get('.fd-shellbar')
+        .contains('Projects')
+        .click();
+      cy.get('.fd-app__sidebar .fd-nested-list__item')
+        .contains('Project One')
+        .click();
+      cy.get('.fd-nested-list__link')
+        .contains('Miscellaneous2')
+        .parent()
+        .should('have.attr', 'title', 'Miscellaneous2');
```
ndricimrr, line 585: Looking at this test now, I realize this might not capture, cases where **node.tooltipText** and **defaults.tooltipText** are defined. Also the title `Left navigation title attr not exist` is a bit off as the test runs if it does exist.

You would need to reset the Luigi config and then try with **node.tooltipText**  being defined for a node, and **defaults.tooltipText** as well. 

Take a look at this [example](https://github.com/SAP/luigi/blob/7500837126227fc5ea03f3d86c653ab43c62d2b9/test/e2e-test-application/e2e/tests/1-angular/navigation.spec.js#L822), on how you can update luigi configuration while in tests.

ndricimrr (changes requested):
Had another look, added some suggestions for improvements below. Currently, `defaults.tooltipText` was not working for me locally.

core/src/navigation/LeftNav.html
rafalgamon: You have right. I changed logic.

`function resolveTooltipText(node, translation) {
    if (node.tooltipText && (node.tooltipText === undefined || node.tooltipText === null)) {
      return defaultsTooltipTextConfig;
    }

    if (node.tooltipText && node.tooltipText.length) {
      return node.tooltipText;
    }

    if (node.tooltipText === false || node.tooltipText === '') {
      return '';
    }

    return translation;
  }`

test/e2e-test-application/e2e/tests/1-angular/navigation.spec.js
rafalgamon: I changed tests (add new tests which check defaults).

docs/navigation-parameters-reference.md
```diff
@@ -89,13 +89,22 @@ The navigation parameters allow you to configure **global** navigation settings
 
 ### nodeChangeHook
 - **type**: function
-- **description**: allows you to invoke and execute a specific function on the global level when a request to navigate to the node occurs. The function receives two node objects as input parameters: the previous node and current node, as described in the configuration. 
+- **description**: allows you to invoke and execute a specific function on the global level when a request to navigate to the node occurs. The function receives two node objects as input parameters: the previous node and current node, as described in the configuration.
 
 ### defaults.isolateView
 - **type**: boolean
 - **description**: renders all views in new frames. This setting overrides the same-domain frame reuse.
 - **default**: the parameter **defaults.isolateView** is `false` by default, and you can overwrite it using the **isolateView** value on a single node level.
```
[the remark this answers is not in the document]
rafalgamon: I add `defaults.tooltipText` to docu

core/src/navigation/LeftNav.html
```diff
@@ -10,7 +10,14 @@
...
     <ul class="fd-nested-list">
       <li class="fd-nested-list__item">
-        <a class="fd-nested-list__link" title="{$getTranslation(navHeader.label)}">
+        <a
+          class="fd-nested-list__link"
+          title="{
+                (navHeader.tooltipText && navHeader.tooltipText.length) ?
+                getTooltipText(navHeader) :
+                (navHeader.tooltipText === false || navHeader.tooltipText === '') ? '' : $getTranslation(navHeader.label)
+                }"
```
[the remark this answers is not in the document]
rafalgamon: I changed logic

core/src/navigation/LeftNav.html
```diff
@@ -545,6 +551,22 @@
...
+      }
+    }
+
+    if (node.tooltipText === false || node.tooltipText === '') {
+      return '';
+    }
+
+    return translation;
+  }
+
```
ndricimrr, line 569: This looks great already 👍 . Thanks for applying suggestions.
- I can see the methods became larger then I expected. Do you think it would make sense to move them to **core/src/utilities/helpers/navigation-helpers.js**. Got a similar request on my [PR](https://github.com/SAP/luigi/pull/2267#discussion_r706195896)  from @stanleychh , sounds like a similiar use case here.

rafalgamon: @ndricimrr Hmm good point. Actual we use it twice but maybe we will implement it in another place so we can move it to navigation-helpers.js.

docs/navigation-parameters-reference.md
```diff
@@ -89,13 +89,28 @@ The navigation parameters allow you to configure **global** navigation settings
...
+- **description**: allows you to add a tooltip to any of the links and buttons of the left side navigation and top navigation. You can set the parameter with the text that will appear on the tooltip.
+- **default**: it is `null` by default, and you can overwrite it using the **tooltipText** value on a single node level.
+- **example**:
+```javascript
+tooltipText: 'Useful links'
+```
+
+### defaults.tooltipText
+- **type**: string
+- **description**: defaults properties apply to nodes where they are not set explicitly, this is just a dumb mechanism.
```
alexandra-simeonova, line 110:
```suggestion
- **description**: applies the [tooltipText](navigation-parameters-reference.md#tooltiptext) property to all navigation nodes where it is not set explicitly. 
```

Thank you for adding more docu. But we probably shouldn't call it a "dumb mechanism" 😄 .

docs/navigation-parameters-reference.md
```diff
@@ -89,13 +89,28 @@ The navigation parameters allow you to configure **global** navigation settings
...
+- **default**: the parameter **defaults.tooltipText** is taking from navigation config, and you can overwrite it using the **tooltipText** value on a single node level.
```
alexandra-simeonova, line 111:
```suggestion
- **default**: the parameter **defaults.tooltipText** is `false` by default, and you can overwrite it using the **tooltipText** value on a single node level.
```

I guess according to [this comment](https://github.com/SAP/luigi/pull/2242#pullrequestreview-744075750) false is the default, i.e. no tooltip will be shown unless explicitly specified.

alexandra-simeonova (changes requested):
I added a few comments to the docu above ^

alexandra-simeonova (approved):

ndricimrr:
Global **defaults.tooltipText** and **tooltipText** on node level, both worked well. However, I couldn't see how to make it work on the Top navigation. 

Some small addition to the defaults.tooltipText could be done, to clear the confusion on how it works together with the node level one.

docs/navigation-parameters-reference.md
```diff
@@ -89,13 +89,28 @@ The navigation parameters allow you to configure **global** navigation settings
...
+- **description**: allows you to add a tooltip to any of the links and buttons of the left side navigation and top navigation. You can set the parameter with the text that will appear on the tooltip.
+- **default**: it is `null` by default, and you can overwrite it using the **tooltipText** value on a single node level.
+- **example**:
+```javascript
+tooltipText: 'Useful links'
+```
+
+### defaults.tooltipText
+- **type**: string
+- **description**: applies the [tooltipText](navigation-parameters-reference.md#tooltiptext) property to all navigation nodes where it is not set explicitly. If **tooltipText** is added to some specific node but it set value as null or undefined, user should see value frome **defaults.tooltipText**.
```
ndricimrr, line 110:
```suggestion
- **description**: applies the [tooltipText](navigation-parameters-reference.md#tooltiptext) property to all navigation nodes where it is not set explicitly. If **tooltipText** is added to some specific node and its is set as as `null` or `undefined`, the value of  **defaults.tooltipText** is used instead.
```

ndricimrr (approved):
LGTM 

only some minor docu suggestions.

PR #2244, opened 2021-08-31
fixed wrong context

### LINKED ISSUE #2211: LuigiClient linkManager wrong navigation scope in modal

hardl:
fixes #2211

### DIFF
--- a/core/src/App.html
+++ b/core/src/App.html
@@ -301,8 +301,8 @@
     }
   };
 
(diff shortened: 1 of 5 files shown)


stanleychh:
Solution applied to #2252. Thanks

PR #2252, opened 2021-09-01
Fix wrong context from modal

### LINKED ISSUE #2229: Fix wrong navigation scope from modal
- Fix issue #2211 : LuigiClient linkManager wrong navigation scope in modal  (PR #2221)

Little refactoring on `buildPath` and `getSubPath` functions at App.html
- Make it stateless. Reduce the usage of global variable _(params, navigationPath, pathParams, currentNode)_. 
- Make those arguments as optional pass to function.

stanleychh:
**Description**

- Refactor `buildPath` and `getSubPath` at App.html
- Fix issue #2211 : LuigiClient linkManager wrong navigation scope in modal

**Related issue(s)**
Fixes #2229


ndricimrr (commented):
So far so good. But there are currently some tests failing after merge from master. Would be good to have a look at that, if it can be fixed.

stanleychh:
@ndricimrr Thanks for reviewing. Delay the DOM rendering fixed the broken test.

ndricimrr (approved):
Re-tested. Not able to find and issues. LGTM 👍

core/src/App.html
```diff
@@ -769,7 +792,7 @@
   let alerts = [];
 
   const getAlertWithId = (alertQueue, id) => {
-    if (!alertQueue || !alertQueue.length) return;
+    if (!alertQueue || !alertQueue.length > 0) return;
```
hardl, line 795: Wrong check... in this context it always leads to the correct result, but be aware that 
!x > 0 is not the same as !(x > 0) for negative x!

core/src/App.html
```diff
@@ -1088,7 +1111,7 @@
 
     EventListenerHelpers.addEventListener('popstate', async e => {
       const alertQueue = alerts;
-      if (!alertQueue || !alertQueue.length) return;
+      if (!alertQueue || !alertQueue.length > 0) return;
```
hardl, line 1114: Wrong check... in this context it always leads to the correct result, but be aware that 
!x > 0 is not the same as !(x > 0) for negative x!

core/src/App.html
stanleychh: Good point! Thanks for reviewing it.

PR #2267, opened 2021-09-08
Expand path exists to intent check

### LINKED ISSUE #2265: Extend pathExists function to include intent check
**Description**
Currently `linkManager().pathExists(...)` doesn't work with intent based paths. 
Add intent option check so that pathExists can also check for intents.

ndricimrr:
**Description**

- Add support for intents: 
i.e.:

```
 Luigi.navigation().pathExists('#?intent=Sales-settings?param1=abc&param2=bcd')
 .then(
   (pathExists) => { 
     console.log('core test', pathExists);
    }
 );
```

and 

```
LuigiClient.linkManager().pathExists('#?intent=Sales-settings?param1=abc&param2=bcd').
then((exists) => {
   console.log("client test", exists)
});
```

**Related issue(s)**

Fixes #2265


client/src/linkManager.js
```diff
@@ -73,7 +73,7 @@ export class linkManager extends LuigiClientBase {
       return;
     }
 
-    const hasIntent = path.toLowerCase().includes('?intent=');
+    const hasIntent = path.toLowerCase().includes('#?intent=');
```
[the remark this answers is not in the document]
ndricimrr: The original call is supposed to be: 
`LuigiClient.linkManager().navigate('#?Intent=Sales-order?id=13') // intent navigation. `
from [here](https://docs.luigi-project.io/docs/luigi-client-api/?section=examples-23) and [Expert scenarios>Use Intent-Based Navigation in Luigi Client](https://docs.luigi-project.io/docs/advanced-scenarios).
So `#?Intent` is the static part. Apparently I'd forgotten to add the hashtag before, so added it here.

client/src/linkManager.js
```diff
@@ -297,11 +297,13 @@ export class linkManager extends LuigiClientBase {
       }.bind(this)
     );
 
+    const hasIntent = path.toLowerCase().includes('#?intent=');
     const pathExistsMsg = {
       msg: 'luigi.navigation.pathExists',
       data: Object.assign(this.options, {
         id: currentId,
         link: path,
+        intent: hasIntent,
```
stanleychh, line 306: Line [300](https://github.com/SAP/luigi/pull/2267/files#diff-ee577ceabc2718709233de3e6b792c3684f0f7cd122e76a9a80471b2eedb188fR300) and line [76](https://github.com/SAP/luigi/blob/06c58b0083e2f0b9cb3b345f67ebe869e262c08e/client/src/linkManager.js#L76) are doing the same.  
Could we move it out and reuse it? (Maybe `helpers`?)
So we can inline function as suggestion. Would this be better? 
```suggestion
        intent: hasIntent(path),
```

ndricimrr: I added the helper function.

stanleychh: Well done. Thanks

core/src/utilities/helpers/routing-helpers.js
```diff
@@ -404,6 +404,7 @@ class RoutingHelpersClass {
           item => item.semanticObject === intentObject.semanticObject && item.action === intentObject.action
         );
         if (!realPath) {
+          console.warn('No mappings found for the specified intent.');
```
ndricimrr, line 407:
```suggestion
```

PR #2277, opened 2021-09-14
Fix critical github security alert under scripts folder

UlianaMunich:
Fixes critical alert for the package "remark-html" under scripts folder


stanleychh (approved):
Tested and worked properly.

PR #2278, opened 2021-09-14
rephrase the way a client can hide or show the loading indicator

### LINKED ISSUE #2269: wrong information about default behaviour for hideAppLoadingIndicator
In our documentation about app loading indicator https://docs.luigi-project.io/docs/luigi-ux-features/?section=app-loading-indicator the paragraph saying
'By default, the loading indicator is removed after Luigi.setConfig({}) has been executed.'
is wrong. It is not removed by default but only , if settings.appLoadingIndicator.hideAutomatically is set to true (see https://github.com/SAP/luigi/blob/9cd4db627b5956a2b1dcd3abd604c9cb5ed6590c/core/src/services/lifecycle-hooks.js#L5).

UlianaMunich:
Fixes #2269

### DIFF
--- a/docs/luigi-ux-features.md
+++ b/docs/luigi-ux-features.md
@@ -76,9 +76,9 @@ To show a loading indicator before Luigi Core or your first micro frontend is re
 </div>
 ```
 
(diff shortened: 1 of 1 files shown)


alexandra-simeonova (approved):
LGTM 👍

PR #2282, opened 2021-09-20
Update versions of npm packages set-vallue, trim-newlines under angul…

UlianaMunich:


stanleychh:
I faced this error while building the angular example. Could someone please also play around?
It seems like require npm [node-sass](node-sass) 
error

stanleychh:
It seems I had an issue of `node-sass` at local. After re-checked, all good.

stanleychh (approved):

PR #2284, opened 2021-09-22
Fix wrong modal creation

JohannesDoberer:
Workaround
fix wrong navigation scope from modal, PR #2252

DXP ticket: https://github.tools.sap/dxp/jukebox/issues/744

### DIFF
--- a/core/src/App.html
+++ b/core/src/App.html
@@ -298,9 +298,15 @@
       const nodePath = Routing.getNodePath(currentNode, urlParamsRaw);
 
       $: preservedViews.push({
-        path: config.iframe.luigi && config.iframe.luigi.pathParams
-          ? GenericHelpers.replaceVars(nodePath, config.iframe.luigi.pathParams, ':', false)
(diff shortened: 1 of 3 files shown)


stanleychh (approved):
Thanks for providing this solution 👍

core/src/App.html
```diff
@@ -1293,7 +1301,16 @@
...
-          let path = buildPath(e.data.params, srcNode, srcPathParams);
+          let path;
+          const specialIframeMessageSource = IframeHelpers.getSpecialIframeMessageSource(
+            e,
+            specialIframeProps
+          );
+          if (specialIframeMessageSource.length > 0) {
+            path = buildPath(e.data.params, srcNode, srcPathParams);
+          } else {
+            path = buildPath(e.data.params, undefined, undefined);
```
stanleychh, line 1312: I prefer this way, which is cleaner. It should work as well.
```suggestion
            path = buildPath(e.data.params);
```

PR #2294, opened 2021-09-30
Fixed resizing main iframe container if a drawer is opened

rafalgamon:
**Description**

- Fixed resizing main iframe container if a drawer is opened


JohannesDoberer (changes requested):
Works as expected 👍 , but please move _resizeMainIframe_ to drawer settings object.
Core api : https://github.com/SAP/luigi/blob/master/core/src/core-api/navigation.js#L98
Client api: https://github.com/SAP/luigi/blob/master/client/src/linkManager.js#L169

JohannesDoberer:
Please name it _overlap: true/false_ instead of _resizeMainIFrame_ because the main content view can also be an webcomponent.

One additional issue:
- open a drawer which resizes the content view (everything is fine)
- navigate somewhere else
Result: Drawer stays (which is good) but it goes in an overlapping mode.
Expected:  Overlapping mode should stay even you navigate somewhere else

PR #2300, opened 2021-10-01
Removed waitForKeyExistency fn in async helpers

JohannesDoberer:


wdoberschuetz (approved):
I presume you've checked whether it was being used anywhere or not, right? :)

PR #2304, opened 2021-10-04
Feature - customAlertHandler

### LINKED ISSUE #2186: Alert: Provide a option to disable Alert.
Please provide an option to disable alert .

We have a MFE container and various MFEs,
We wanted to disable the alert as we have to handle all the alerts as per our theme.

JohannesDoberer:
Fixes #2186

### DIFF
--- a/docs/general-settings.md
+++ b/docs/general-settings.md
@@ -158,6 +158,8 @@ theming : {
      }
   }
 ```
+* **customAlertHandler** is a function which will be called if it is defined. With this function the Luigi alerts will be disabled and you can implement own alerts. This function gets `settings` and `openFromClient` as parameter.
+
 
 ### Third-party cookies support check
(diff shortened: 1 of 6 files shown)


docs/general-settings.md
```diff
@@ -158,6 +158,8 @@ theming : {
      }
   }
 ```
+* **customAlertHandler** is a function which will be called if it is defined. With this function the Luigi alerts will be disabled and you can implement own alerts. This function gets `settings` and `openFromClient` as parameter.
```
alexandra-simeonova, line 161:
```suggestion
* **customAlertHandler** is a function which will be called if it is defined. With this function, Luigi alerts will be disabled and you can implement your own alerts. This function gets `settings` and `openFromClient` as parameters.
```

PR #2359, opened 2021-11-03
added expanded class

hardl:
set an expanded style class at the li element of left nav category item


JohannesDoberer (approved):
LGTM ;)!

PR #2380, opened 2021-11-15
clear children and titleResolver cache

hardl:

PR #2383, opened 2021-11-17
Introduce clearNavigationCache() to clear children and titleResolver cache

### LINKED ISSUE #2379: Expose functionality in core api to clear navigation related caches
There are 2 caches currently, nodeData cache for children resolution and titleResolver cache. Both should be cleared when function is called.

stanleychh:
**Description**

Provide a core API to clean nodeData cache for children resolution and titleResolver cache.
`Luigi.clearNavigationCache()`

**Related issue(s)**
Fixes #2379

Draft PR: https://github.com/SAP/luigi/pull/2380


core/src/core-api/config.js
```diff
@@ -279,6 +280,32 @@ class LuigiConfig {
...
+          if (node.children) {
+            clearTitleResolverCache(node.children);
+          }
+        });
+      }
+    };
+
+    clearTitleResolverCache(this.getConfig().navigation.nodes);
+  }
 }
```
wdoberschuetz, line 309: (q) We clear the cache of the nested navigation nodes - shouldn't we return a warning of some kind of any of the nodes do not meet the criteria (i.e. titleResolver doesn't exist in the node) within the ```if```s? :)

wdoberschuetz (approved):

docs/luigi-core-api.md
```diff
@@ -228,6 +228,20 @@ Luigi.reset();
 
 -   **since**: 1.14.0
 
+#### clearNavigationCache
+
+Delete cache. Clear titleResolver cache of nested navigation nodes.
```
hardl, line 233: also (and more important) the node cache is cleared, i.e. asynchronously fetched children will be fetched again.
But i would phrase it more generic, since there might  be additional caches in the future that might also be cleared by this function... so sth like:
"Clear navigation node related caches."

hardl (changes requested):

docs/luigi-core-api.md
stanleychh: Updated the description. Thanks for your suggestion.

hardl (approved):

PR #2385, opened 2021-11-17
Searchfield configurable

JohannesDoberer:
Follow-up:
https://github.com/SAP/luigi/issues/2415


docs/navigation-parameters-reference.md
```diff
@@ -843,6 +843,15 @@ The global search is an input field available in the top navigation bar. The sea
...
 
+### searchFieldCentered
+- **type**: boolean
+- **description**: The search input field will be rendered in the center of the shellbar.
+- **since**: NEXTRELEASE
+
+### globalSearchCenteredCancelButton
+- **type**: string
+- **description**: defines the label of the cancel button. It will be displayed if you want to hide the search field on a smaller viewport.
+- **since**: NEXTRELEASE
```
alexandra-simeonova, line 854: Ordering the parameters in alphabetical order according to the new convention
```suggestion
### globalSearchCenteredCancelButton
- **type**: string
- **description**: defines the label of the cancel button. It will be displayed if you want to hide the search field on a smaller viewport.
- **since**: NEXTRELEASE

### searchFieldCentered
- **type**: boolean
- **description**: The search input field will be rendered in the center of the shellbar.
- **since**: NEXTRELEASE

```

JohannesDoberer: Good point Alex 👍  :). Totally forgot

alexandra-simeonova (approved):

docs/navigation-parameters-reference.md
```diff
@@ -843,6 +843,16 @@ The global search is an input field available in the top navigation bar. The sea
...
+
+### searchFieldCentered
+- **type**: boolean
+- **description**: The search input field will be rendered in the center of the shellbar.
+- **since**: NEXTRELEASE
+
```
UlianaMunich, line 855: I'd suggest to switch the order. First mention the `searchFieldCentered` and then `globalSearchCenteredCancelButton`. Is `globalSearchCenteredCancelButton` mandatory or optional? What would be shown if it not defined or empty?

UlianaMunich (changes requested):
- In addition to the comment below, there is a tiny issue on mobile. When a "Cancel" button is shown there is no need for y "x" next to the magnifier in the search field. 

- when a User activates the search via starting typing it has to have a focus state on entire search input not on the buttons on the right.

docs/navigation-parameters-reference.md
JohannesDoberer: Since we have to rule for alphabetical order I would leave it as it is.
I would add the optional flag and add a hint that the cancel btn is only availible if searchfield is centered

JohannesDoberer: > when a User activates the search via starting typing it has to have a focus state on entire search input not on the buttons on the right.
This works for me.

core/src/navigation/GlobalSearchCentered.html
```diff
@@ -0,0 +1,499 @@
...
+    <button
+      class="fd-button fd-shellbar__button"
+      aria-haspopup="true"
+      aria-expanded="{!isSearchFieldVisible}"
+      on:click="{toggleSearch}"
+      data-testid="luigi-search-cancel-btn"
+    >
+      {$getTranslation(cancelBtn)}
+    </button>
+  </div>
```
UlianaMunich, line 112: <button
      class="fd-button fd-shellbar__button"
      aria-haspopup="true"
      aria-expanded="{!isSearchFieldVisible}"
      on:click|stopPropagation="{()=> toggleSearch()}"
```

Do you think this would work as well?

Comment (UlianaMunich) [inline core/src/navigation/GlobalSearchCentered.html:112]:
```suggestion
     <button
      class="fd-button fd-shellbar__button"
      aria-haspopup="true"
      aria-expanded="{!isSearchFieldVisible}"
      on:click|stopPropagation="{()=> toggleSearch()}"
      data-testid="luigi-search-cancel-btn"
      >
      {$getTranslation(cancelBtn)}
    </button>
```

UlianaMunich: Would that work? Just wanted to remove empty `div` is possible

core/src/navigation/GlobalSearch.html
```diff
@@ -181,12 +212,22 @@
         }
       } else if (GenericHelpers.isFunction(search.searchProvider.onInput)) {
         search.searchProvider.onInput();
+        if (search.searchFieldCentered) {
+          document.querySelector('.luigi-search__input').value
+            ? (clearSearchFieldBtn = true)
+            : (clearSearchFieldBtn = false);
```
stanleychh, line 218: I did not get this logic. Did you want to clear the search input value?
Could you please briefly explain what should be done here?

JohannesDoberer: Showing the clear button if input field has some value

core/src/navigation/TopNav.html
```diff
@@ -22,8 +22,25 @@
...
+      bind:inputElem
+      bind:luigiCustomSearchRenderer__slot
+      on:closeSearchResult
+    />
+  </div>
+  {/if}
   <div class="fd-shellbar__group fd-shellbar__group--actions">
-    {#if !authorizationEnabled || isLoggedIn} {#if isGlobalSearchAvailable}
+    {#if !authorizationEnabled || isLoggedIn} {#if globalSearchConfig} {#if
+    !isGlobalSearchCentered}
```
stanleychh, line 43:
```suggestion
    {#if !authorizationEnabled || isLoggedIn} {#if
    !isGlobalSearchCentered}
```
It seems `isGlobalSearchCentered` already included `{#if globalSearchConfig}` condition. Do we need to check it twice?

core/src/utilities/helpers/global-search-helpers.js
```diff
@@ -0,0 +1,15 @@
...
+class GlobalSearchHelperClass {
+  constructor() {}
+
+  handleVisibilityGlobalSearch() {
+    const globalSearchCtn = document.querySelector('.lui-global-search');
+    if (globalSearchCtn.offsetWidth <= 384) {
+      globalSearchCtn.classList.add('lui-global-search-toggle');
+    } else {
+      globalSearchCtn.classList.remove('lui-global-search-toggle');
+    }
```
stanleychh, line 11:
```suggestion
    const globalSearchCtn = document.querySelector('.lui-global-search');
    const condition = globalSearchCtn.offsetWidth <= 384;
    globalSearchCtn.classList.toggle('lui-global-search-toggle', condition);
```
`toggle()` seems perfect to be here. FYI: https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/toggle

Please test whether this solution works properly.

stanleychh (approved):

PR #2391, opened 2021-11-19
Style language dropdown under User Settings Dialog

### LINKED ISSUE #2346: Improve Language dropdown in User Settings
[DXP Issue](https://github.tools.sap/dxp/backlog/issues/391)

Issues

 - Wrong content density of Input component (cozy, but frame is compact)
 - The should be no gap between Input and opened Drop-Down list
 - Drop-Down toggle button should remain in active state when list is open
 - Selected list item is not rendered with correct interactive state
-  Language labels should be matching the target language: "Deutsch (de), 简体中文 (zh), English (en), Français (fr), ..." to allow the user to more easily identify their language. Ideally, use the two column variant of the select list with the technical language identifier as a secondary info (following ISO 639-1).

- [Fix a tiny bug](https://github.tools.sap/dxp/backlog/issues/333)

UlianaMunich:
Fixes #2346


core/fundamentalStyleClasses.js
```diff
@@ -21,7 +21,6 @@ module.exports = [
   './node_modules/fundamental-styles/dist/form-item.css',
   './node_modules/fundamental-styles/dist/form-label.css',
   //'./node_modules/fundamental-styles/dist/form-message.css',
-  //'./node_modules/fundamental-styles/dist/form-select.css',
```
ndricimrr, line 24: Why do we remove this 🤔 . Just curious.

ndricimrr:
Could you rename the PR title @UlianaMunich . Seems like it got the commit name instead. It'd just be easier when doing the upcoming release

UlianaMunich:
> Could you rename the PR title @UlianaMunich . Seems like it got the commit name instead. It's just for making it easier when doing the upcoming release

Sure!

ndricimrr (approved):
Looks good 👍

PR #2393, opened 2021-11-23
Strip query params in linkmanager navigate

### LINKED ISSUE #2387: strip query params when added to route in client linkmanager navigate
a micro frontend should not be able to control search query parameters unless it has permissions to do so (https://docs.luigi-project.io/docs/navigation-parameters-reference/?section=clientpermissionsurlparameters).

But currently it seems to be possibel to call:
linkManager().navigate('/some/path?someParam=someValue') and the param is added to the main apps url.
This must be prevented!

-> Encode values of withParams()
linkManager().withParams({bar: "test&t=foo"}).navigate("/search/components")will navigate to /search/components?~bar=test&t=foo.
-> Encode values of addCoreSearchParams()

Expected: In both cases Luigi core encodes the parameters correctly

JohannesDoberer:
Fixes #2387


core/src/App.html
```diff
@@ -1348,10 +1348,11 @@
 
       if ('luigi.navigation.open' === e.data.msg) {
         isNavigateBack = false;
-
         const srcNode = isSpecialIframe ? iframe.luigi.currentNode : undefined;
         const srcPathParams = isSpecialIframe ? iframe.luigi.pathParams : undefined;
-
+        if (e.source !== window && !e.data.params.intent) {
```
hardl, line 1353: i guess it wouldn't harm to also check if e.data.params.link is truthy here

hardl (changes requested):

core/src/App.html
```diff
@@ -1348,10 +1348,11 @@
 
       if ('luigi.navigation.open' === e.data.msg) {
         isNavigateBack = false;
-
         const srcNode = isSpecialIframe ? iframe.luigi.currentNode : undefined;
         const srcPathParams = isSpecialIframe ? iframe.luigi.pathParams : undefined;
-
+        if (e.source !== window && !e.data.params.intent && e.data.params.link) {
+          e.data.params.link = e.data.params.link.split('?')[0];
+        }
```
stanleychh, line 1355:
```suggestion
        const params = e.data.params;
        const { intent, link, newTab, modal, splitView, drawer, withoutSync } = params;
        if (e.source !== window && !intent && link) {
          link = link.split('?')[0];
        }
```

We repeated a lot of `e.data.params` here, as well as inside `luigi.navigation.open`.

stanleychh: We could declare `e.data.params` on the top and make the other conditions(newTab, drawer, splitView, etc...) shortly.

stanleychh:
Let's move this forward. 
The [luigi.navigation.open](https://github.com/SAP/luigi/blob/8ba179df2f5f81eb9cef812648b019504be6f85f/core/src/App.html#L1349) part needs further refactoring to reduce repetition in codebase. I will do that in a separate ticket then.

stanleychh (approved):

Issue #2402, opened 2021-11-26
LuigiContextService Observable runs outside of Angular Zone


SomeKay:
**Description**

When using the `contextObservable` from https://github.com/SAP/luigi/blob/master/client-frameworks-support/client-support-angular/projects/client-support-angular/src/lib/service/luigi-context.service.impl.ts#L33, we are getting issues with the emitted values being run outside of Angular's zone.

That leads to Angular's Change Detection not working properly and our UI not refreshing.

**Expected result**

Angular's Zone should be respected.

**Steps to reproduce**

The easiest way to test it would be to subscribe to `contextObservable` and look at `NgZone.isInAngularZone()`.
We have discovered it through using NGRXs Runtime Checks: https://ngrx.io/guide/store/configuration/runtime-checks#strictactionserializability which throw errors when our Effects start working with LuigiContextService.

**Troubleshooting**

Using `getContextAsync` and wrapping it with a `from` doesn't throw the code outside of the Angular Zone.

PR #2403, opened 2021-11-26
Fix Luigi Context Observable running outside of Angular Zone

SomeKay:
**Description**
Fixes the problems arising from the Context Observable running outside of the Angular Zone.

**Related issue(s)**
Resolves https://github.com/SAP/luigi/issues/2402


ndricimrr (approved):
Looks good. Tested and works as expected 👍

PR #2424, opened 2021-12-14
TASK-2405 | Fixed navigation flyout styles CSS variables

rafalgamon:
**Description**

- Fixed navigation flyout styles CSS variables


UlianaMunich:
See attached the difference in sap_fiori_3_hdc theme, it is different than all our popovers.
Screenshot 2021-12-15 at 14 05 28

Here is the Spec of popover from Fiori 3. Hope this will help to reuse the right CSS variables which will be the right ones for all 4 themes we support.

other than that, the arrow is somehow white but should be the same color as the popover.

UlianaMunich (approved):
LGTM!

PR #2450, opened 2021-12-20
Increase test coverage

JohannesDoberer:


core/src/core-api/theming.js
```diff
@@ -89,7 +89,7 @@ class LuigiTheming {
    * @example
    * Luigi.theming().isThemingAvailable()
    */
-  isThemingAvailable() {
+  isThemingAvailable() /* istanbul ignore next */ {
```
ndricimrr, line 92: Added tests look good 👍 .

What is the incentive around the many **/*istanbul ignore next */**'s, or do we want to ignore them for now.

ndricimrr (approved):
👍

Issue #2260, opened 2021-09-06
Refactoring - convert .html file to .svelte file (/core/src/navigation/)


stanleychh:
Issue: According to this [PR](https://github.com/SAP/luigi/pull/2236/files#diff-2b5d04d984372d11919588635e3dd35851ee7aba20af0065721e057a43d78c68), it seems missing the prettier setup for Svelte syntax. 

For improving better syntax readability, convert /core/src/navigation/*.html files to .svelte files.

TO DO: introduce [prettier-plugin-svelte](https://www.npmjs.com/package/prettier-plugin-svelte) to project. Or any other package/plugin to make it work.

VS Code settings: https://marketplace.visualstudio.com/items?itemName=svelte.svelte-vscode

ndricimrr:
I believe we've already had a small discussion in the past with our svelte files being in `.html` and not `.svelte`. I don't remember why it was chosen to be html in the very beginning, so that could be a good topic for tomorrows after scrum. 
 
Regarding prettier, I think the most simple and easy solution is to first convert html files to `.svelte` and then add a prettier config for svelte files as you suggested. However, need to discuss it first if that can have any implications on functionality.

PR #2466, opened 2021-12-28
Proper styling for App Switcher with one entity

### LINKED ISSUE #2408: UI Bug when displaying one-element dropdown
**Description**

When using `appSwitcher` and the `items` attribute in the returned object includes only one element there is a UI bug related to Luigi's navigation.  It tries to display a one-element list, so there is just an arrow with a dot. 
It appears on every browser.

image

**Expected result**
When the `items` array includes only one element there shouldn't be an option to dropdown 

**Actual result**

The one-element list is displayed with a UI bug that shows an arrow and a dot.

**Steps to reproduce**

1. Use the code below or create similar `appSwitcher` in the configuration that returns one element array with a proper link.
```js
appSwitcher: {
  items: [{
    title: 'Cluster Overview',
    link: '/clusters',
  }]
}
```
2. Click on the Dropdown you've created.

UlianaMunich:
Fixes #2408


stanleychh (approved):
The logic looks good.
We need to convert those `.html` files to `.svelte` files and applied `prettier-plugin-svelte`  sooner rather than later (#2260). Otherwise, the readability of the nested if else conditions are bad.

PR #2472, opened 2022-01-10
Node params are not deleted

### LINKED ISSUE #2471: Function addNodeParams does not delete properties
Function addNodeParams does not delete properties with value undefined

JohannesDoberer:
Node params are not deleted if value is undefined
fixes #2471


stanleychh (approved):
Looks good to me.

PR #2475, opened 2022-01-12
Fix - getCoreSearchParams should not return undefined

### LINKED ISSUE #2473: getCoreSearchParams returns sometimes undefined
getCoreSearchParams returns sometimes undefined.
This happens when no searchParams are sent if "luigi.navigate" in context of withoutSyc is called.
Problem is most likely there:
https://github.com/SAP/luigi/blob/9d8f46f035350db9ae717b846b46c3aaf572103a/core/src/services/routing.js#L383
(Webcomponend compound view)

JohannesDoberer:
Fixes #2473 
How to reproduce

Example config for simpledev:
```
{
              pathSegment: 'home2', 
              label: 'h2', 
              hideFromNav: false, 
              children: [
                {
                
                  pathSegment:'test',
                  label:'test',
                  loadingIndicator: {
                      enabled: false
                  },
                  viewUrl: 'microfrontend.html',
                  clientPermissions:{
                    urlParameters:{
                      q:{
                        write: true,
                        read: true
                      }
                    }
                  }
                },
                {
                  pathSegment:'test2',
                  label:'test2',
                  loadingIndicator: {
                      enabled: false
                  },
                  viewUrl: 'microfrontend2.html'
                }
              ]
          }
```
- navigate **without sync** from **test** to **test2**
`LuigiClient.linkManager().withoutSync().navigate('/home2/test2')`
- `LuigiClient.getCoreSearchParams()` should return an empty object instead of **undefined**


ndricimrr (approved):
Works as expected 👍

PR #2481, opened 2022-01-14
option to use default alert with customAlertHandler

hardl:


docs/general-settings.md
```diff
@@ -158,7 +158,7 @@ theming : {
      }
   }
 ```
-* **customAlertHandler** is a function which will be called if it is defined. With this function, Luigi alerts will be disabled and you can implement your own alerts. This function gets `settings` and `openFromClient` as parameters. In addition, we recommend to return a [promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise): 
+* **customAlertHandler** is a function which will be called if it is defined. With this function, Luigi alerts will be disabled and you can implement your own alerts. This function gets `settings` and `openFromClient` as parameters. It must either return a [promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) that gets resolved, when the alert is dismissed, or false, if default luigi alert should be shown:
```
alexandra-simeonova, line 161:
```suggestion
* **customAlertHandler** is a function which will be called if it is defined. With this function, Luigi alerts will be disabled and you can implement your own alerts. This function gets `settings` and `openFromClient` as parameters. It must either return a [promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) which gets resolved when the alert is dismissed, or `false` if the default Luigi alert should be shown. This example shows how the function can be used: 
```

JohannesDoberer (approved):
👍

alexandra-simeonova (approved):

PR #2483, opened 2022-01-17
Escaping helpers improvement and config fix of e2e app

### LINKED ISSUE #2477: Wrong config for E2E App, /projects/pr2/on-node-activation/navigated
Check why there a modal which has an undefined content and whether it is needed there, since it blocks the entire page and only a full reload helps.

JohannesDoberer:
Fixes #2477


ndricimrr (approved):
Looks good!

PR #2488, opened 2022-01-19
Add function which allow to get the footer container

### LINKED ISSUE #2321: Allow HTML in sideNavFooterText OR make it a function
**Description**

Currently it's only possible to set `sideNavFooterText` to plain text. I'd like to add more elements here, for example links with `a` element.

We could either allow setting HTML here, or make `sideNavFooterText` accept either string or function - in latter case, the function would be called with each navigation refresh.

**Reasons**

As a workaround, I can use `document.querySelector('.fd-side-nav')` and insert a child below the left navigation, but it will disappear after any navigation change.

UlianaMunich:
Fixes #2321 

Test it in the console of the browser:

```
var span = document.createElement("span");
span.innerHTML='simpleTest';
Luigi.elements().getNavFooterContainer().append(span);
```

Result: a new <span>simpeTest</span> is added after a footer Text (Luigi version x.x.x)


stanleychh:
@UlianaMunich, the `LeftNav.html` already converted to `LeftNav.svelte`.  Please let me know if you need help on solving conflict. Thanks.

alexandra-simeonova (approved):

test/e2e-test-application/e2e/tests/0-fiddle/fiddle-navigation.spec.js
```diff
@@ -905,4 +905,37 @@ describe('Fiddle', () => {
       cy.get('.lui-nav-title .fd-nested-list__title').should('contain', 'test');
     });
   });
+
+  describe('Custom text in the footer', () => {
+    context('Desktop', () => {
```
JohannesDoberer, line 910: There is actually no need to define a 'desktop context' if there is no other context distinction.

JohannesDoberer (approved):
LGTM

PR #2490, opened 2022-01-20
Add node params works when hashrouting enabled

### LINKED ISSUE #2484: Node params does not work if hash routing is enabled
If you add node params when hash routing is enabled you get an url like this
LuigiClient.addNodeParams({test:'tets'},false) =>
`https://fiddle.luigi-project.io/?%7Etest=tets#/home/overview`

JohannesDoberer:
Fixes #2484 
How to test:
- Enable hash routing
- add node params to url using Luigi core api and Luigi client

Expected:
Node params should be added to hash and not before hash (which happens with luigi client => https://fiddle.luigi-project.io/?%7Etest=test#/home/overview)

Follow-up issues
- encode is not consistend
- deleting params does not work see #2491


core/src/core-api/routing.js
```diff
@@ -86,7 +68,11 @@ class LuigiRouting {
 
     const paramPrefix = RoutingHelpers.getContentViewParamPrefix();
     const url = new URL(location);
-    this._modifySearchParam(params, url.searchParams, paramPrefix);
+    if (LuigiConfig.getConfigValue('routing.useHashRouting')) {
+      url.hash = RoutingHelpers.addParamsOnHashRouting(params, url.hash, paramPrefix);
+    }else{
```
stanleychh, line 73: It seems like ESLint was not applied.

stanleychh, line 73:
```suggestion
    } else {
```

core/src/utilities/helpers/routing-helpers.js
```diff
@@ -597,6 +597,30 @@ class RoutingHelpersClass {
...
   }
+
+  // Adds and remove properties from searchParams
+  modifySearchParam(params, searchParams, paramPrefix) {
+    for (const [key, value] of Object.entries(params)) {
+      const paramKey = paramPrefix ? `${paramPrefix}${key}` : key;
+
+      searchParams.set(paramKey, value);
+      if (value === undefined) {
+        searchParams.delete(paramKey);//TODO change to key
```
stanleychh, line 608: Do we need to change back to Key?  https://github.com/SAP/luigi/pull/2472/files#diff-60b3f2d3aa92a9cc5e51872f739096e2023d2a5412d5b6c0d566c73458186463L76

core/src/utilities/helpers/routing-helpers.js
```diff
@@ -597,6 +597,30 @@ class RoutingHelpersClass {
     };
     component.showAlert(alertSettings, false);
   }
+
+  // Adds and remove properties from searchParams
+  modifySearchParam(params, searchParams, paramPrefix) {
```
stanleychh, line 602:
```suggestion
  modifySearchParams(params, searchParams, paramPrefix) {
```
Please change to `modifySearchParams` which is more consistent to other params related functions. Thanks

core/src/utilities/helpers/routing-helpers.js
JohannesDoberer: No, not needed. Was a comment for me to avoid merge conflicts

stanleychh (approved):

PR #2512, opened 2022-02-03
Set CSS-variable for badge color

### LINKED ISSUE #2511: change css variable for badge text color
should be 
`var(--sapContent_BadgeTextColor);`
but currently is 
`var(--sapShell_TextColor,#fff);` (probably coming from fundamental styles)

Proposed Fix:
set `color: var(--sapContent_BadgeTextColor);` here: https://github.com/SAP/luigi/blob/97626408de51a678017f0cf0b872141f8a367470/core/src/navigation/BadgeCounter.svelte#L19

UlianaMunich:
Fixes #2511


hardl (approved):
LGTM

PR #2566, opened 2022-02-16
Keyboard accesibility for user settings dialog

### LINKED ISSUE #2453: A11Y: User Setting dialog keyboard interactions
[DXP Issue-origin](https://github.tools.sap/dxp/backlog/issues/391)

Accessibility

The navigation menu list is not compliant yet with [Fiori Keyboard specifications](https://wiki.wdf.sap.corp/wiki/display/Acc/Fiori+Keyboard+Interaction?preview=/1842806926/1867396960/Fiori%20Keyboard%20Interaction%20(untagged).pdf) (p. 219-222)
- When the focus is on a list item, it should be able to use HOME to go to the first item and use END for the last item. In Fundamental NGX this is working fine.

The **select list** is not compliant with keyboard specifications. Fundamental NGX Select is working fine.
-  When the focus is set on the select (Language), it should be able to use ALT + DOWN to open the dropdown. Also use ALT + UP to close the dropdown when it's open.
-  After choosing a value from the select list, the focus should remain on the select.
-  When the select dropdown list is open, pressing ESC should only close the dropdown, not the entire dialog.
- 
UI:
- Add the possibility to show User Initials if available and no Account Profile picture is in pace
- Clicking on the Label "Language" should set the focus on the dropdown

UI references from UI5:https://sapui5.hana.ondemand.com/#/entity/sap.m.Select/sample/sap.m.sample.Select

_FOLLO UP TASK for select language dropdown:_
-  When opening the dropdown list, the focus should be already set on the selected list item.
-  When the dropdown is open, it should be able to navigate through the list with DOWN and UP. Not with TAB, because this should take focus to the "Save" button. (Follow up task #2565 )

UlianaMunich:
Fixes #2453


core/src/UserSettingsDialog.svelte
```diff
@@ -305,19 +307,174 @@
...
+                            aria-label={userSettingsGroupProperty[1].altText
+                              ? userSettingsGroupProperty[1].altText
+                              : ''}
+                            style="background-image:url('{userSettingsGroupProperty[1]
+                              .icon}'); background-size:cover;"
+                          />
+                        {/if}
+                      {:else if userSettingsGroupProperty[1].initials}
+                        <span
+                          class="fd-avatar fd-avatar--s fd-avatar--circle fd-avatar--thumbnail fd-user-menu__avatar"
```
JohannesDoberer, line 392: `margin-right: 0.75rem;` is missing
Bildschirmfoto 2022-02-21 um 19 31 16

JohannesDoberer (changes requested):
It looks good. Few two small things.
1. Space between initials and title is too small.
2. When you set `style: 'list'` on type enum you will get a dropdown as well. 
(In context of the e2e test app it would be "time format", [see](https://github.com/SAP/luigi/pull/2566/files#diff-2a5a05a31efee4d1fec64af3f3a27f14f21621418376736d1e815fd06c507106R33) and change it to `list`)
For this "kind of" dropdown the keyboard accessibility does not work. 
3. Revert user settings configuration changes (order + custom user settings micro frontend)

test/e2e-test-application/src/luigi-config/extended/userSettings.js
```diff
@@ -38,33 +66,6 @@ class UserSettings {
         }
       }
     },
-    language: {
```
JohannesDoberer, line 41: Please change it back. First item in the list should be `User Account` and second `Language` otherwise we will get a few failing tests + more failing backwards compatibility tests.

test/e2e-test-application/src/luigi-config/extended/userSettings.js
```diff
@@ -95,12 +96,6 @@ class UserSettings {
           options: ['red', 'green']
         }
       }
-    },
```
JohannesDoberer, line 98: This is needed. We have tests for displaying custom user settings micro frontend in general + on those nodes it's not necessary to have `settings` defined.

test/e2e-test-application/src/luigi-config/extended/userSettings.js
```diff
@@ -95,12 +96,6 @@ class UserSettings {
...
-    custom: {
-      label: 'Custom',
```
JohannesDoberer, line 100: The custom user settings node is needed. We have tests for displaying custom user settings micro frontend in general + on those nodes it's not necessary to have `settings` defined.

ndricimrr (approved):
Looks good. All listed guideline specs working well 👍

JohannesDoberer (approved):

docs/user-settings.md
```diff
@@ -99,7 +100,12 @@ These parameters used in the example above allow you to configure the items in t
 #### icon
 
 - **type** string (optional)
-- **description** name of the icon, without the `sap-icon--` prefix or path to an image.
+- **description** name of the icon, without the `sap-icon--` prefix or path to an image. It is also possible to use a picture instead of an icon via passing the right path to it.
```
alexandra-simeonova, line 103:
```suggestion
- **description** name of the icon, without the `sap-icon--` prefix. It is also possible to use an image instead of an icon by specifying its path.  
```

alexandra-simeonova (approved):

PR #2594, opened 2022-02-25
confirmation modal from special iframe mfes

hardl:
With this PR confirmation modal result is now passed back correctly to modal micro frontends.


wdoberschuetz (approved):
LGTM

PR #2598, opened 2022-02-28
keepURL in pagenotfoundhandler

hardl:
**Description**

With this PR, it will be possible to change main url behaviour in case of pagenotfound error, i.e. one can now define to keep the erroneous url in the browsers address bar (e.g. to correct a typo). An example config would look like this:
```
...
 routing: {
              useHashRouting: true,
              pageNotFoundHandler: (path, anyMatch) => {
                return {
                  redirectTo: '/error',
                  keepURL: true
                };
              }
          },
...
```


hardl:
Created as draft, needs test adjustments and docu.

docs/navigation-parameters-reference.md
```diff
@@ -50,7 +50,7 @@ You can configure the way Luigi tackles routing in your application in the `rout
 
 ### pageNotFoundHandler
 - **type**: any
-- **description**: defines custom behavior when a `404` error occurs.  Luigi handles it by default. Leave its body empty if you have an external `404` handling. You can return an Object with **redirectTo** parameter if you want Luigi to redirect to a specific navigation path after execution.
+- **description**: defines custom behavior when a `404` error occurs.  Luigi handles it by default. Leave its body empty if you have an external `404` handling. You can return an Object with **redirectTo** and  **keepURL** as parameters. You can use the **redirectTo** parameter if you want Luigi to redirect to a specific navigation path after execution. Setting the **keepURL** parameter to true will keep the erroneous url onto the browsers address bar.
```
alexandra-simeonova, line 53:
```suggestion
- **description**: defines custom behavior when a `404` error occurs.  Luigi handles it by default. Leave its body empty if you have an external `404` handling. You can return an Object with **redirectTo** and  **keepURL** as parameters. You can use the **redirectTo** parameter if you want Luigi to redirect to a specific navigation path after execution. Setting the **keepURL** parameter to `true` will keep the erroneous URL onto the browser's address bar. 
```

alexandra-simeonova (approved):

stanleychh (approved):

PR #2599, opened 2022-03-01
Url anchor support for micro frontends

stanleychh:
**Description**
Introduce two functions at LuigiClient
1. getAnchor()

Path is `http://some.url.de#myAnchor` 
`LuigiClient.getAnchor()` return `myAnchor`

3. setAnchor(String)

Path is `http://some.url.de` 
`LuigiClient.setAnchor('myAnchor)`
The path becomes to `http://some.url.de#myAnchor` 

**Related issue(s)**
Fixes ##2433


ndricimrr (approved):
Looks good 👍

PR #2605, opened 2022-03-07
Fix faulty pathExists race condition

### LINKED ISSUE #2592: linkManager().pathExists(…) returns wrong results if called multiple times at the same time
**Description**

When calling `linkManager().pathExists(…)` multiple times at the same time, it might return wrong results. 

The problem is that the messageId for the request to the Core is just `Date.now()`, when called twice in the same millisecond, this will assign the same id and the responses will be mapped to the wrong promises. https://github.com/SAP/luigi/blob/19e1bc9d2bbeb3692ee306300884e8809d8da0ec/client/src/linkManager.js#L292

**Expected result**
Returns the correct response for each query.

**Actual result**

Returns either the correct result or the same value for multiple requests, even if the answers should differ.

**Steps to reproduce**

Assuming path "a" does exist, and "b" does not.
``` 
const promiseA = linkManager().pathExists("a");
const promiseB = linkManager().pathExists("b");
console.log(await promiseA);
console.log(await promiseB);
```

It might log `true false` (expected) or `true true` or `false false`.

**Troubleshooting**

ndricimrr:
**Description**

- Replaces **Date.now()** with **getRandomId()** to avoid consecutively run promises having same id, which would then lead into incorrect promise results.

**Related issue(s)**
Fixes #2592


JohannesDoberer (approved):
LGTM 👍

PR #2626, opened 2022-03-21
hide empty categories

hardl:
hide categories in left nav if they don't have visible children


wdoberschuetz (approved):
LGTM. Tested as well locally.

PR #2627, opened 2022-03-22
Fixed search params encoded twice issue

### LINKED ISSUE #2604: Search Params encoded twice in the URL
**Description**

When calling `addCoreSearchParams` from a MFE, the value in the URL is encoded twice. 
Reason: Encoded explicitly in https://github.com/SAP/luigi/blob/97626408de51a678017f0cf0b872141f8a367470/core/src/utilities/helpers/routing-helpers.js#L516 and implicitly with the combination of https://github.com/SAP/luigi/blob/97626408de51a678017f0cf0b872141f8a367470/core/src/utilities/helpers/routing-helpers.js#L604 and https://github.com/SAP/luigi/blob/b6737095d837be28cfbab2f5d40ea7d05e01d7e1/core/src/core-api/routing.js#L81

This is problematic as `getCoreSearchParams` only decodes once.

Tested with "@luigi-project/core": "1.20.2-dev.202202251428" and checked in the code at 63196f9f42e6e8ee2323a06fdb8fb3e723550d1e

**Expected result**

Encoded only once. 

**Actual result**
Encoded twice.

**Steps to reproduce**
`addCoreSearchParams({q: "abc:def"}, true)` leads to the URL `?q=abc%253Adef`. Calling `getCoreSearchParams()` then returns `{q: "abc%3Adef"}` 

**Troubleshooting**

stanleychh:
**Description**
When calling `addCoreSearchParams` from a micro-frontend, the value in the URL is encoded twice.

**Related issue(s)**
Fixes #2604

### DIFF
--- a/core/src/utilities/helpers/routing-helpers.js
+++ b/core/src/utilities/helpers/routing-helpers.js
@@ -511,11 +511,7 @@ class RoutingHelpersClass {
     if (!GenericHelpers.isObject(localSearchParams)) {
       return;
     }
-    Object.keys(localSearchParams).forEach(key => {
-      if (localSearchParams[key] !== undefined) {
-        localSearchParams[key] = encodeURIComponent(localSearchParams[key]);
-      }
-    });
+
     if (currentNode && currentNode.clientPermissions && currentNode.clientPermissions.urlParameters) {
       const filteredObj = {};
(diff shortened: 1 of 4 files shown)


wdoberschuetz (approved):
LGTM.

PR #2629, opened 2022-03-24
Fix click support for open in new tab

### LINKED ISSUE #2623: shortcut click support for opening node in new tab
We have the option to show hrefs for navigation node elements (https://docs.luigi-project.io/docs/navigation-parameters-reference/?section=addnavhrefs). With that, one can use right click and select "open in new tab". But what doesn't work is the OS-specific CTRL+click (windows) / CMD+click (mac), probably because the click handler always does a preventDefault.

Solution:
click event handler should check for ctrl&meta key and if pressed, not call preventDefault

ndricimrr:
**Description**
Applying Ctrl + Click on Windows, or CMD + Click on MacOS on a navigation item, should open the tab in new window as expected, **when** addNavHref is set to true in the navigation configurations.
Same applies for Shift + Click. 

Need to check that the navigation items of the following, do open in a new tab properly through the above commands :
- Context Switcher
- Top Navigation
- Product Switcher
- Logo
- App Switcher 
- Left Navigation
- Global Navigation

- Add preventDefault() only if CMD, CTRL or Shift keyboard buttons are also pressed upon the click. 

**Related issue(s)**

Fixes #2623


core/test/utilities/helpers/navigation-helpers.spec.js
```diff
@@ -714,4 +714,49 @@ describe('Navigation-helpers', () => {
       assert.equal(NavigationHelpers.renderIconClassName(''), '');
     });
   });
-});
\ No newline at end of file
+
+  describe('handleNavAnchorClick', () => {
+    var event;
```
wdoberschuetz, line 719: (n2h) Could we go for ```let``` or ```const```? :)

wdoberschuetz (approved):

wdoberschuetz (approved):

JohannesDoberer (approved):

Issue #2326, opened 2021-10-16
allowrules should be separated by semicolon


hardl:
unlike in sandbox attribute, Feature Policies in allow attribute are separated by semicolon.
Currently luigi just concatenates the values in `allowRules` though, adding whitespace.

Expected behaviour:

A config like
```
allowRules: ['a','b','c']
```
should result in the following allow attribute:
```
<iframe ... allow="a; b; c;">
```

Note: For compatibility reasons (user could have already workarounds adding semicolon to the value), also the following config should have the same result:
```
allowRules: ['a;','b;','c;']
```

PR #2642, opened 2022-04-04
Add functionallity for allow attribute to be seperated by semicolons

viktorsperling:
**Description**

- Feature Policies in `allow` attribute are separated by semicolon according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Feature_Policy.
Also, for compatibility reasons, this PR covers workarounds implemented by the user adding semicolons.

**Related issue(s)**
Fixes - [#2326](https://github.com/SAP/luigi/issues/2326)

**Results of local testing**

- 598 passing
- 6 pending
- 0 failed

**Coverage in summary**
- Statements   : 80.23% ( 2001/2494 )
- Branches     : 72.65% ( 1304/1795 )
- Functions    : 78.26% ( 450/575 )
- Lines        : 80.83% ( 1927/2384 )


core/src/utilities/helpers/iframe-helpers.js
```diff
@@ -216,6 +216,13 @@ class IframeHelpersClass {
     const iframe = document.createElement('iframe');
     iframe.src = ViewUrlDecorator.hasDecorators() ? ViewUrlDecorator.applyDecorators(viewUrl) : viewUrl;
     if (allowRules) {
+      allowRules.forEach((rule, index) => {
+        if (rule.indexOf(';') != -1) {
+          allowRules[index] = rule;
+        } else {
+          allowRules[index] = rule + ';';
+        }
```
ndricimrr, line 224: Works well 👍 
Have a suggestion here, which could make this a bit easier and save some lines:
```suggestion
        allowRules[index] = rule + (rule.includes(';') ? '' : ';');
```

ndricimrr:
There are some merge conflicts related to you latest commit wrt to package-lock.json's.
 I'd recommend to discard the changed package-lock.json files altogether as resolving conflicts here is not really worth the time. As some versions are dynamic we can't maintain the pacakge-lock.jsons for every minimal version change, we usually do this for releasing so you can ignore them here.

ndricimrr (approved):
Looks good 👍

PR #2666, opened 2022-04-11
Error handling on productswitcher columns calculation

### LINKED ISSUE #2650: Investigate daily dashboard console error
**Description**
- There seems to be an error showing on our Dashboard:
Screenshot 2022-04-07 at 13 56 05

Right when loading the page

**Expected result**
no errors should show, or uncaught exception be handled. Investigate root of the issue and issue fix as needed.

JohannesDoberer:
Fixes #2650

### DIFF
--- a/core/src/utilities/helpers/navigation-helpers.js
+++ b/core/src/utilities/helpers/navigation-helpers.js
@@ -19,8 +19,9 @@ class NavigationHelpersClass {
 
   getProductSwitcherColumnsNumber() {
     const productSwitcherConfig = this.getProductSwitcherConfig();
+    if (!productSwitcherConfig.items) return;
     let productSwitcherColumns = productSwitcherConfig.columns;
-    let productSwitcherItemsAmount = productSwitcherConfig.items().length;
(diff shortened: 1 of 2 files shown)


core/src/utilities/helpers/navigation-helpers.js
```diff
@@ -19,8 +19,14 @@ class NavigationHelpersClass {
 
   getProductSwitcherColumnsNumber() {
     const productSwitcherConfig = this.getProductSwitcherConfig();
+    if (!productSwitcherConfig.items) return;
```
wdoberschuetz, line 22: (n2h) Could we get a test case where items ```undefined``` or ```null```? :)

core/src/utilities/helpers/navigation-helpers.js
```diff
@@ -19,8 +19,14 @@ class NavigationHelpersClass {
...
     let productSwitcherColumns = productSwitcherConfig.columns;
```
wdoberschuetz, line 23: (n2h) This seems to be never reassigned so it could be a const.

wdoberschuetz (approved):

PR #2674, opened 2022-04-19
Custom item renderer for app switcher

### LINKED ISSUE #2539: Add custom renderer for app switcher.
**Description**
Context switcher has the possibility to use a custom renderer. Please add the same functionality of custom renderer for app switcher as well. 

**Reasons**
Fundamental-styles ([link](https://fundamental-styles.netlify.app/?path=/docs/components-menu--desktop-and-tablet#list-with-icon)) allows adding icons for the dropdown items. We use it to distinguish generic action from the rest of the items on the list. We would like to have the same for the app switcher.

**Attachments**
We use custom renderer in the context switcher:
Screen Shot 2022-02-11 at 11 40 58

We would like to have that same for the app switcher:
Screen Shot 2022-02-11 at 11 35 08

JohannesDoberer:
Fixes #2539


docs/navigation-parameters-reference.md
```diff
@@ -865,6 +865,45 @@ The app switcher is a drop-down list available in the top navigation bar. It all
 - **type**: boolean
 - **description**: includes the link to the root of the Luigi application in the drop-down using the **title** specified in the **settings/header** section of the configuration as a label.
 
+### customAppSwitcherItemRenderer
+- **type**: Function
+- **description**: This function allows you to customize the single list element rendered in the default app switcher popover.
+- **attributes**:
+  - **item** single application element
+  - **slot** `ul` element as slot. You can append your custom `li` entries to this `ul` element.
```
alexandra-simeonova, line 873: One more small thing: since the "rule" now is for parameters to be in alphabetical order, and this one begins with `c`, it should be around line 855 instead of 868. But I'm not able to make this edit in a GitHub suggestion  😄

```suggestion
### customAppSwitcherItemRenderer
- **type**: function
- **description**: This function allows you to customize the single list element rendered in the default app switcher popover.
- **attributes**:
  - **item** single application element.
  - **slot** `ul` element as slot. You can append your custom `li` entries to this `ul` element.
```

JohannesDoberer: Thx Alex. Good catch!

alexandra-simeonova (approved):

core/src/navigation/LogoTitle.svelte
```diff
@@ -51,7 +54,9 @@
               pathData = obj.pathData;
             } else if (prop === 'appSwitcherItems') {
               appSwitcherItems = obj.appSwitcherItems;
-            } else if (prop === 'pathParams') {
+            } else if (prop === 'customAppSwitcherItemRenderer') {
+              customAppSwitcherItemRenderer = obj.customAppSwitcherItemRenderer;
+            }else if (prop === 'pathParams') {
```
stanleychh, line 59:
```suggestion
            } else if (prop === 'pathParams') {
```
It seems like linting did not work properly.

core/src/utilities/helpers/generic-helpers.js
```diff
@@ -140,6 +140,9 @@ class GenericHelpersClass {
    * @returns {string} string without leading slash
    */
   trimLeadingSlash(str) {
+    if (typeof str !== 'string') {
+      return '';
+    }
```
stanleychh, line 145: I would suggest having a generic function like
`const isString = (value) => typeof value === 'string' || value instanceof String;`

Then we could reuse it here, also for `trimTrailingSlash` function

stanleychh (approved):

docs/navigation-parameters-reference.md
```diff
@@ -861,18 +861,21 @@ The app switcher is a dropdown list available in the top navigation bar. It allo
...
   ```javascript
     appSwitcher: {
       items:[...],
-      customAppSwitcherItemRenderer: (item, slot) => {
-        let li = document.createElement('li');
-        li.setAttribute('class', 'fd-menu__item');
+      customAppSwitcherItemRenderer: (item, slot, appSwitcherApiObj) => {
         let a = document.createElement('a');
-        a.setAttribute('class', 'fd-menu__link');
+        a.setAttribute('class', 'fd-menu__link test');
```
stanleychh, line 873: Is `test` class a leftover?

JohannesDoberer: yes it's a left over thanks

stanleychh (approved):

PR #2686, opened 2022-04-27
Fix getNodeParams decoding issue

### LINKED ISSUE #2673: LuigiClient.linkManager().navigate() doesn't send query parameters
**Description**

Using `LuigiClient.linkManager().navigate("/some/path?param=test)` with path that contains query parameters isn't working

Is this expected behavior or not? If this is the expected it would be great to have a note in the documentation that is not supported.

I have seen that in the Navigate with Intent it accepts "hash" so I assumed it also accepted query params `linkManager().navigate('/#?intent=Sales-settings?project=pr2&user=john')`

**Expected result**

navigate to new page with URL: `<host>/some/path?param=test`

**Actual result**

Query param is not send

 `<host>/some/path`

**Steps to reproduce**

call navigate with query param

**Troubleshooting**

I am thinking of using `addCoreSearchParams`

ndricimrr:
**Description**
Adding node params from the client results in extra sanitisation when special characters are encountered.

- de-sanitise back the parameter string upon returning it

**Related issue(s)**
Fixes #2673


client/src/helpers.js
```diff
@@ -125,6 +125,22 @@ class Helpers {
...
   }
+
+  deSanitizeParamsMap(paramsMap) {
+    return Object.entries(paramsMap).reduce((sanitizedMap, paramPair) => {
+      sanitizedMap[this.deSanitizeParam(paramPair[0])] = this.deSanitizeParam(paramPair[1]);
+      return sanitizedMap;
+    }, {});
+  }
+
+  deSanitizeParam(param = '') {
```
wdoberschuetz, line 136: I think we could extend this list with an ampersand ('&') and backtick ('`') https://stackoverflow.com/a/48226843

wdoberschuetz (approved):

client/src/helpers.js
ndricimrr: We are de-sanitising(reverting) the sanitisation that we do in another place [here](https://github.com/SAP/luigi/blob/master/core/src/utilities/helpers/escaping-helpers.js#L54) manually. I only need to revert what that sanitiser does to the string, to get the original input, sanitisation for sanitisations sake is not the intention here.

client/src/lifecycleManager.js
```diff
@@ -430,7 +430,7 @@ class LifecycleManager extends LuigiClientBase {
    * const nodeParams = LuigiClient.getNodeParams()
    */
   getNodeParams() {
```
ndricimrr, line 432: add param

Issue #2565, opened 2022-02-16
Scalable keyboard accessibility for Language Dropdown in User Setttings Dialog


UlianaMunich:
[DXP Issue-origin](https://github.tools.sap/dxp/backlog/issues/391)

- When the dropdown is open, it should be able to navigate through the list with DOWN and UP. Not with TAB, because this should take focus to the "Save" button.
- After choosing a value from the select list, the focus should remain on the select.
- When opening the dropdown list, the focus should be already set on the selected list item.

UI references from UI5:https://sapui5.hana.ondemand.com/#/entity/sap.m.Select/sample/sap.m.sample.Select

Keep in mind other accessibility functionality from this ticket #2453

PR #2691, opened 2022-04-29
2565: Scalable keyboard accessibility for Dropdown in User Settings Dialog

wdoberschuetz:
- handling of the keyboard up and down arrows, as well as confirmation with enter as per description in the ticket

**Related issue(s)**
#2565


UlianaMunich:
Please make sure the selector of the dropdown can be applied to any dropdown independently from the language, index, aria-label, placeholder, etc. Most likely the unique class or id added to the HTML might help. Keeping in mind there might be different content inside of the dropdown we call "language dropdown" and there could be multiple dropdowns on the same view.

UlianaMunich:
Couple of things to adjust from my side.

- There is an example that DXP is referring to and using a lot when it comes to components' behavior. [here](https://sap.github.io/fundamental-ngx/#/core/select). Please use it to mimic the behavior in our code. 

- The dropdown can't be opened now on the "Enter" key-press, same as it does now on the "Space" key-press. Please check it. It did working the past

- Regarding the active state of the selected item in the dropdown. When the dropdown is open, the focus has to be on a selected item. Once the arrow up or down is pressed, the focus needs to show which item is in focus without adding the "is-sected" class. The "is-selected" class is only applied if the "Enter" key was pressed. 

- if you might have a possibility to add classes it would be amazing.
- 
Thank you!

core/src/UserSettingsEditor.svelte
```diff
@@ -117,38 +120,111 @@
...
+          list.item(chosenElementIndex).classList.add('is-selected');
+          list.item(chosenElementIndex).classList.add('is-focus');
+          return;
+        }
+        if (chosenElementIndex < schemaItem.options.length - 1) {
+          list.item(chosenElementIndex).classList.remove('is-selected');
+          list.item(chosenElementIndex).classList.remove('is-focus');
+          chosenElementIndex += 1;
+          list.item(chosenElementIndex).classList.add('is-selected');
+          list.item(chosenElementIndex).classList.add('is-focus');
```
UlianaMunich, line 153: these lines repeated quite often, could you optimize it and remove console.log?

UlianaMunich (changes requested):
Here is an example of the arrow behavior from the openUI5. When a user clicks on the arrow the focus should show the navigation, while the selected item stays the same. Once a user clicks on Enter the new item is selected. 
Screenshot 2022-05-23 at 21 16 31

UlianaMunich (approved):
LGTM! Great job!

PR #2692, opened 2022-05-02
Fix empty nodeParams on browser back navigation

### LINKED ISSUE #2689: When navigating back using the browser back button, the `luigi.navigate` message does not contain the correct nodeParams
**Description**

When setting nodeParams, e.g: ` addNodeParams({ q:"test"}, true);`, then updating them, e.g.: ` addNodeParams({ q:"test2"}, true);` and then using the browser back button, the URL is updated correctly to `/...?~q=test`, however the `luigi.navigate` message sends the nodeParams as an empty object. 
Reason: The nodeParams are computed from the path, however the path used is only the path, without the query parameters: 

Computation of the nodeParams: https://github.com/SAP/luigi/blob/69ffb4f3f8ed3d034730c8e7a909005aa3f208f6/core/src/services/routing.js#L273-L274

Computation of the path: 
https://github.com/SAP/luigi/blob/cf91e9c7521db30af88e92d7668882ec358e1adb/core/src/utilities/helpers/routing-helpers.js#L184-L187 together with https://github.com/SAP/luigi/blob/69ffb4f3f8ed3d034730c8e7a909005aa3f208f6/core/src/services/routing.js#L131

**Expected result**

The nodeParams are set correctly after using the back button.

**Actual result**

The nodeParams are empty in the message from the core after using the back button and therefore `getNodeParams` will not return the correct result.

**Steps to reproduce**

See above.

**Troubleshooting**

ndricimrr:
**Description**

- Add parameters to pathname when hash routing disabled.

**Related issue(s)**
Fixes #2689

### DIFF
--- a/core/src/services/routing.js
+++ b/core/src/services/routing.js
@@ -128,7 +128,8 @@ class RoutingClass {
       const intentPath = RoutingHelpers.getIntentPath(hash);
       return intentPath ? intentPath : '/';
     }
-    const path = (window.history.state && window.history.state.path) || window.location.pathname;
+    const params = window.location.search ? window.location.search : '';
+    const path = (window.history.state && window.history.state.path) || window.location.pathname + params;
     return path
       .split('/')
       .slice(1)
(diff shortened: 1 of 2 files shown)


wdoberschuetz (approved):
LGTM. Checked locally as well.

PR #2703, opened 2022-05-12
Prevent setting an invalid feature toggle

### LINKED ISSUE #2700: prevent setting a feature toggle via param
It should be possible to prevent a feature toggle being activated if it is "disabled" as feature toggle param in the url, e.g. like `...?ft=!myFeature`. In that case, Luigi.featureToggles().setFeatureToggle('myFeature') should have no effect.

stanleychh:
**Description**

Prevent a feature toggle being activated if it is "disabled" as feature toggle. 
e.g. like `...?ft=foo,!bar`. For this case, only `foo` feature toggle will be set. `!bar` is an invalid usage.

Same behavior on`Luigi.featureToggles().setFeatureToggle('!bar')`. This invalid usage won't be set.

**Related issue(s)**
Fixes #2700


ndricimrr (commented):
If I understand the original issue problem correctly, I should be able to do the following:

1. Navigate to **http://localhost:4200/projects/pr1/settings?ft=berlin,munich,!paris**

2. Running **Luigi.featureToggles().getActiveFeatureToggleList()** should return :
 **['berlin', 'munich']**

3. I would then run **Luigi.featureToggles().setFeatureToggle('paris')** , and **'paris'** toggle should **not** be activated.

4. Now if I run   Luigi.featureToggles().getActiveFeatureToggleList() again then I should be getting:
 **['berlin', 'munich']**
since **'paris'** is disabled from step 1 above. However what I do get instead is:
 **['berlin', 'munich', 'paris'].**

If I understand the ticket correctly **'paris'** shouldn't have been set, since in the url it had an exclamation :exclamation: mark ? 🤔 

Could you clarify if thats intended or not.

stanleychh:
@ndricimrr Thanks for reviewing. 
Good catch, I implemented the logic which is an opposite of the point 3.

@hardl, could you please take a look above scenario and give us inputs? Thanks

hardl:
@ndricimrr is right here
if !paris is in the ft query param list, paris should not be "activateable".
As stated in the ticket description, in that case Luigi.featureToggles().setFeatureToggle('paris') should have no effect, i.e. after calling it, of course paris should not be in the list of active feature toggles

stanleychh:
@hardl, thanks for clarifying this. I will implement this accordingly.

stanleychh:
@ndricimrr, it's ready for review again. Please have a look. Thanks

core/src/core-api/featuretoggles.js
```diff
@@ -54,7 +46,37 @@ class LuigiFeatureToggles {
...
+   * @example Luigi.featureToggles().getFeatureToggleList();
+   */
+  getFeatureToggleList() {
     return [...get(this.featureToggleList)];
   }
+
+  isValid(featureToggleName) {
+    if (isString(featureToggleName)) return true;
+
+    console.warn('Feature toggle name is not valid or not a type of string');
```
ndricimrr, line 65:
```suggestion
    console.warn(`Feature toggle name is not valid or not of type 'string'`);
```

core/src/core-api/featuretoggles.js
```diff
@@ -54,7 +46,37 @@ class LuigiFeatureToggles {
...
+   * Get a list of active feature toggles
+   * @memberof FeatureToggles
+   * @return {Array} of feature toggles
+   * @example Luigi.featureToggles().getFeatureToggleList();
+   */
+  getFeatureToggleList() {
     return [...get(this.featureToggleList)];
   }
+
+  isValid(featureToggleName) {
```
ndricimrr, line 62: Definitely not a must, but I think it'd be nice if we had JSDoc style comments for all the newly **added** methods, like the one above this.

core/src/core-api/featuretoggles.js
```diff
@@ -54,7 +46,37 @@ class LuigiFeatureToggles {
...
+    if (isString(featureToggleName)) return true;
+
+    console.warn('Feature toggle name is not valid or not a type of string');
+    return false;
+  }
+
+  isDuplicateOrDisabled(featureToggleName) {
+    if (get(this.featureToggleList).includes(`${featureToggleName}`)) {
```
ndricimrr, line 70:
```suggestion
    if (get(this.featureToggleList).includes(featureToggleName)) {
```
We could remove the back-ticks here I suppose, in case it is not necessary.

core/src/core-api/featuretoggles.js
```diff
@@ -54,7 +46,37 @@ class LuigiFeatureToggles {
...
+      console.warn('Feature toggle name already exists');
+      return true;
+    }
+
+    if (get(this.featureToggleList).includes(`!${featureToggleName}`)) {
+      console.warn('Disabled feature toggle can not be activateable');
```
ndricimrr, line 76:
```suggestion
      console.warn('Disabled feature toggle can not be activated');
```
'activated' might fit better in this context.
 
(I was wondering if this is even needed. Do we even need to show the user this warning that they cannot activate a feature toggle. I would assume if I am not allowed to access something, it's because I shouldn't. Therefore I would not be necessarily entitled to so much information. What are your thoughts on this ? )

core/src/core-api/featuretoggles.js
```diff
@@ -54,7 +46,37 @@ class LuigiFeatureToggles {
...
+   * @memberof FeatureToggles
+   * @return {Array} of feature toggles
+   * @example Luigi.featureToggles().getFeatureToggleList();
+   */
+  getFeatureToggleList() {
     return [...get(this.featureToggleList)];
   }
+
+  isValid(featureToggleName) {
+    if (isString(featureToggleName)) return true;
```
ndricimrr, line 63: **isString(featureToggleName)** is quite neat indeed, 
& I know that `featureToggleName && typeof featureToggleName === 'string'` looks a bit ugly, but it might be better than importing another package for one function. I guess here we can use the ugly version, as I presume importing a library might not be a luxury we can easily afford. Being that it can easily add to the bundle size in the long run might be better to avoid it as much as we can.

core/test/services/feature-toggle.spec.js
```diff
@@ -19,6 +23,12 @@ describe('Feature Toggle', () => {
       'test'
     ]);
```
ndricimrr, line 25: **isDuplicateOrDisabled** and **isValid** seem to be newly added functions. It would be nice to have some unit functions for them as well.

ndricimrr (commented):
Tested and functionality looks good 👍 

Had some comments that might be worth taking a look

core/src/core-api/featuretoggles.js
```diff
@@ -54,7 +46,37 @@ class LuigiFeatureToggles {
...
+  }
+
+  /**
+   * Get a list of active feature toggles
+   * @memberof FeatureToggles
+   * @return {Array} of feature toggles
+   * @example Luigi.featureToggles().getFeatureToggleList();
+   */
+  getFeatureToggleList() {
     return [...get(this.featureToggleList)];
```
ndricimrr, line 59: 🤔 I am not sure if we need this function anymore, but for the time being seems like its not used anywhere. Unless this was specifically requested for something planned in the future, I would remove altogether.

core/src/core-api/featuretoggles.js
stanleychh: Yeah, I thought this as well.
I will move it to `generic-helpers.js` and that can be reused in other places.

core/src/core-api/featuretoggles.js
```diff
@@ -54,7 +46,37 @@ class LuigiFeatureToggles {
    * @example Luigi.featureToggles().getActiveFeatureToggleList();
    */
   getActiveFeatureToggleList() {
+    return [...get(this.featureToggleList)].filter(ft => !ft.startsWith('!'));
```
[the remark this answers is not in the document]
stanleychh: I got your point. However, we need to have this feature to prevent activating a feature toggle which was disabled before. 

The steps:
1. Navigate to `http://localhost:4200/projects/pr1/settings?ft=berlin,munich,!paris`

2 Running `Luigi.featureToggles().getActiveFeatureToggleList()` should return :
['berlin', 'munich']

3 Run `Luigi.featureToggles().setFeatureToggle('paris')` , and `paris` toggle should not be activated.

The way to prevent `paris` be activated is keeping the `disabled` feature toggle(`!paris`) in the list. And prevent it be set.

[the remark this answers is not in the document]
stanleychh: After synced with Philipp, I adjusted the logic. As below
1. `...?ft=!paris` is valid and will be kept in the FT list for preventing `paris` to be active in the future. 
2. Luigi.featureToggles().setFeatureToggle('!paris') will be ignored. Because it starts with 
`!` which is not a valid call.

core/src/utilities/helpers/generic-helpers.js
```diff
@@ -21,6 +21,10 @@ class GenericHelpersClass {
     return anyParam && this.isFunction(anyParam.then);
   }
 
+  isString(value) {
```
ndricimrr, line 24: Could you add some JSDoc here as well. Also would be great to have a unit test for this one as well. It would be simple but also helps with the test coverage threshold later.

docs/luigi-core-api.md
```diff
@@ -1321,6 +1322,46 @@ Returns **[Array](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Gl
 
 -   **since**: 1.4.0
 
+#### isValid
```
ndricimrr, line 1325: I believe **isValid** and is **isDuplicatedOrDisabled** are not intended to be publicly available API calls as they are internal class helper functions, unless it was specifically required . 
They are most likely automatically added here because they are in the core-api/ folder. I think you can add the following comment similar to [here](https://github.com/SAP/luigi/blob/master/core/src/core-api/config.js#L177) and then run **npm run docu** again to remove the docu about these functions:

`   * @private`

core/src/core-api/featuretoggles.js
```diff
@@ -54,7 +49,43 @@ class LuigiFeatureToggles {
...
+
+    console.warn(`Feature toggle name is not valid or not of type 'string'`);
+    return false;
+  }
+
+  /**
+   * Check if feature toggle is duplicated or already disabled
+   * @memberof FeatureToggles
+   * @param {string} featureToggleName
+   * @since 1.22.0
```
ndricimrr, line 74:
```suggestion
```
private helper function. seems redundant

core/src/core-api/featuretoggles.js
```diff
@@ -54,7 +49,43 @@ class LuigiFeatureToggles {
...
   getActiveFeatureToggleList() {
-    return [...get(this.featureToggleList)];
+    return [...get(this.featureToggleList)].filter(ft => !ft.startsWith('!'));
+  }
+
+  /**
+   * Check if it is a valid feature toggle
+   * @memberof FeatureToggles
+   * @param {string} featureToggleName
+   * @since 1.22.0
```
ndricimrr, line 59:
```suggestion
```

docs/luigi-core-api.md
stanleychh: Good catch! Thanks for pointing this out.

ndricimrr (approved):
Looks good 👍

core/src/core-api/featuretoggles.js
```diff
@@ -10,19 +11,17 @@ class LuigiFeatureToggles {
   /**
    * Add a feature toggle to an active feature toggles list
    * @memberof FeatureToggles
+   * @param {string} featureToggleName the name of the feature toogle
```
alexandra-simeonova, line 14:
```suggestion
   * @param {string} featureToggleName the name of the feature toggle
```

core/src/core-api/featuretoggles.js
```diff
@@ -10,19 +11,17 @@ class LuigiFeatureToggles {
...
+   * @param {boolean} fromUrlQuery whether the feature toogle is sent from url query. Default is false.
```
alexandra-simeonova, line 15:
```suggestion
   * @param {boolean} fromUrlQuery specifies whether the feature toggle is sent from URL query. The default is `false`.
```

docs/luigi-core-api.md
```diff
@@ -1275,12 +1275,13 @@ Add a feature toggle to an active feature toggles list
 
 ##### Parameters
 
--   `featureToggleName`  
+-   `featureToggleName` **[string](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)** the name of the feature toogle
+-   `fromUrlQuery` **[boolean](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Boolean)** whether the feature toogle is sent from url query. Default is false. (optional, default `false`)
```
alexandra-simeonova, line 1279:
```suggestion
-   `featureToggleName` **[string](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)** the name of the feature toggle
-   `fromUrlQuery` **[boolean](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Boolean)** specifies whether the feature toggle is sent from URL query. The default is `false`.(optional, default `false`)
```

ndricimrr:
please run **npm run docu** after the docu commits, might be not properly synced

PR #2746, opened 2022-06-09
Add withOptions function for link manager

### LINKED ISSUE #2737: withOptions function for linkmanager
Add an new function `withOptions` to linkmanager with an argument to control route change behaviour.

First usecase to be implemented with this issue is add two options:
- preventHistoryEntry: replaceState instead of pushState should be used
- preventContextUpdate: after url change there will be no postMessage to iframe

e.g. `LuigiClient.linkManager().withOptions({preventContextUpdate:true, preventHistoryEntry: true}).navigate('/overview')`

stanleychh:
**Description**

Add a new function `withOptions` to link manager with an argument to control route change behaviour.

preventHistoryEntry: replaceState instead of pushState should be used
preventContextUpdate: after url change there will be no postMessage to iframe

e.g. `LuigiClient.linkManager().withOptions({preventContextUpdate:true, preventHistoryEntry: true}).navigate('/overview')`

**Related issue(s)**
Fixes #2737


core/src/services/routing.js
```diff
@@ -44,10 +44,13 @@ class RoutingClass {
   /**
     navigateTo used for navigation
     Triggers a frame reload if we are on the same route (eg. if we click on same navigation item again)
-    @param route string  absolute path of the new route
+    @param {string} route absolute path of the new route
+    @param {Object} options navigation options of keepBrowserHistory, navSync, preventContextUpdate
```
ndricimrr, line 48: Could you add here something similar to [this example parameter](https://github.com/SAP/luigi/blob/master/core/src/core-api/ux.js#L31) or [this example parameter](https://github.com/SAP/luigi/blob/master/core/src/core-api/ux.js#L31)

I would assume it would like like this:

```
@param {Object} options navigation options
@param {Object} options.keepBrowserHistory <<explanation here>>
@param {Object} options.navSync <<explanation here>>
@param {Object} options.preventContextUpdate <<explanation here>>
```

It would be pretty helpful when debugging something in the future.

client/src/linkManager.js
```diff
@@ -306,6 +308,30 @@ export class linkManager extends LuigiClientBase {
     return this;
   }
 
+  /**
+   * Sends options to the route. The parameters are used by the `navigate` function. Use it optionally in combination with any of the navigation functions and receive it as part of the context object in Luigi Client.
+   * @memberof linkManager
+   * @param {Object} options
+   * @returns {linkManager} link manager instance
```
ndricimrr, line 315:
```suggestion
   * @returns {linkManager} link manager instance
   * @since NEXTRELEASE
```
since its a new function

client/src/linkManager.js
```diff
@@ -306,6 +308,30 @@ export class linkManager extends LuigiClientBase {
     return this;
   }
 
+  /**
+   * Sends options to the route. The parameters are used by the `navigate` function. Use it optionally in combination with any of the navigation functions and receive it as part of the context object in Luigi Client.
```
ndricimrr, line 312:
```suggestion
   * Sets options to customise route changing behaviour. The parameters are used by the `navigate` function. Use it optionally in combination with any of the navigation functions and receive it as part of the context object in Luigi Client.
```

client/src/linkManager.js
```diff
@@ -306,6 +308,30 @@ export class linkManager extends LuigiClientBase {
...
+   * @memberof linkManager
+   * @param {Object} options
```
ndricimrr, line 314: could you add the same thing as the previous comment here. It would be pretty helpful for users when reading the docu of it:

```
@param {Object} options navigation options
@param {Object} options.keepBrowserHistory <<explanation here>>
@param {Object} options.navSync <<explanation here>>
@param {Object} options.preventContextUpdate <<explanation here>>
```

ndricimrr (commented):
So far tested and seems to be working well. 
Left some comments and suggestions.

ndricimrr (commented):
Could you re-generate the docu again with npm run docu and commit. there seems to be a mismatch

ndricimrr (approved):
LGTM

PR #2770, opened 2022-06-23
Responsive padding for the Shellbar Component

### LINKED ISSUE #2764: Introduce option to enable responsive shellbar paddings
Option should be configurable by setting 
`{ settings: { header: { responsivePaddings: true }}}`

If set to true, `fd-shellbar--responsive-paddings` should be added to class attribute of the shellbar (https://github.com/SAP/luigi/blob/5c7d50ea2192e301e8a747a31685d3e8b2bad761/core/src/navigation/TopNav.svelte#L308).

Note: requires release of https://github.com/SAP/fundamental-styles/pull/3698

UlianaMunich:
Fixes #2764 

Can be tested against "fundamental-styles" : "0.24.1-rc.7", this release contains fixes we need


docs/general-settings.md
```diff
@@ -160,6 +160,17 @@ Take a look at our [i18n](i18n.md) section for an implementation suggestion.
   },
 ```
 
+### header.responsiveShellbarPaddings
+- **type**: boolean
+- **description**: based on the browser window size the responsiv padding will be applied to the Shellbar Component when set to `true`.
```
wdoberschuetz, line 165: ```responsiv``` -> ```responsive```

alexandra-simeonova, line 165:
```suggestion
- **description**: based on the browser window size, a responsive padding will be applied to the Shellbar Component when set to `true`.
```

docs/general-settings.md
```diff
@@ -160,6 +160,17 @@ Take a look at our [i18n](i18n.md) section for an implementation suggestion.
...
+- **default**: by default, the parameter is set to `false` and the padding for Shellbar Component will stay the same and inherited from `.fd-shellbar` class.
```
alexandra-simeonova, line 166:
```suggestion
- **default**: by default, the parameter is set to `false`, meaning that padding for the Shellbar Component will stay the same as inherited from the `.fd-shellbar` class.
```

docs/general-settings.md
UlianaMunich: updated)

wdoberschuetz (approved):

Issue #2775, opened 2022-06-28
Update Docu once FD Styles NEXTRELEASE is done


UlianaMunich:
Once FD Styles release new version a Luigi Core neds to be updated with it. And once it is done a new description to the docu can be added:

### header.responsiveShellbarPaddings
- **type**: boolean
- **description**: based on the browser window size, a responsive padding will be applied to the Shellbar Component when set to `true`.
- **default**: by default, the parameter is set to `false`, meaning that padding for the Shellbar Component will stay the same as inherited from the `.fd-shellbar` class.
- **example**: 
```
 header: {  object / function / Promise
   responsiveTopNavPaddings: true
  },
```

PR #2776, opened 2022-06-28
Add documentation for responsive shellbar padding

alexandra-simeonova:
**Description**

- Add documentation for responsive padding (see also https://github.com/SAP/luigi/pull/2770)
- Documentation should be added after the next Fundamental Styles release

**Related issue(s)**
#2775


ndricimrr (approved):
Looks good 👍

docs/general-settings.md
```diff
@@ -161,6 +161,17 @@ Take a look at our [i18n](i18n.md) section for an implementation suggestion.
...
 ```
 
+### header.responsiveShellbarPaddings
+- **type**: boolean
+- **description**: based on the browser window size, a responsive padding will be applied to the Shellbar Component when set to `true`.
+- **default**: by default, the parameter is set to `false`, meaning that padding for the Shellbar Component will stay the same as inherited from the `.fd-shellbar` class.
+- **example**:
+```
+ header: {  object / function / Promise
+   responsiveTopNavPaddings: true
```
ndricimrr, line 171: https://github.com/SAP/luigi/blob/master/core/src/navigation/TopNav.svelte#L113-L115
```suggestion
   responsiveShellbarPaddings: true
```

PR #2800, opened 2022-07-06
Restructure documentation

### LINKED ISSUE #2796: Restructure documentation
**Description**

- Take advantage of the new collapsible navigation in the docs and create a better structure
- Improve existing documents 

**Attachments**

alexandra-simeonova:
**Description**

- Restructure documentation to avoid having to scroll through too many items in the same category
- Separate docs according to their topic
- Proposed menu looks like this:
Screen Shot 2022-07-06 at 09 12 18

**Related issue(s)**
Resolves #2796

### DIFF
--- a/docs/communication.md
+++ b/docs/communication.md
@@ -3,12 +3,12 @@
   "node": {
     "label": "Custom messages",
     "category": {
-      "label": "Advanced",
+      "label": "Luigi Client",
       "collapsible": true
     },
     "metaData": {
-      "categoryPosition": 4,
-      "position": 0
+      "categoryPosition": 5,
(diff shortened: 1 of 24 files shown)


docs/communication.md
```diff
@@ -3,12 +3,12 @@
   "node": {
     "label": "Custom messages",
```
ndricimrr, line 4: The custom messages/Communication page is **both** part of Core and Client section. But here I see it under the Luigi Client section/category. Maybe it would be more suitable to put it under Navigation category, or some other place you see fits better?

hardl (approved):
LGTM

PR #2780, opened 2022-06-29
PR for testing compatibility tests

viktorsperling:
**Description**

Changes proposed in this pull request:

- Solves #2759

**Related issue(s)**
#2759

PR #2861, opened 2022-08-10
Introduce js-test-application for e2e tests

JohannesDoberer:
Introduced a new vanilla js test application as replacement of using fiddle locally as test application.

The new test application has no default luigi config. A test has to call the application with specific config.


test/e2e.sh
```diff
@@ -39,10 +39,10 @@ runWebserver 8090
 WS_EXT_PID=$PID
 
 echo ""
-echo "Fiddle App"
-cd $BASE_DIR/../website/fiddle
-killWebserver 8080
```
ndricimrr, line 44: If we are getting rid of 8080, should we maybe then also replace it [here](https://github.com/SAP/luigi/blob/master/test/e2e-test-application/src/luigi-config/extended/userSettings.js#L92) and [here](https://github.com/SAP/luigi/blob/master/test/e2e-test-application/src/luigi-config/extended/userSettings.js#L105) with the new 4500?

ndricimrr: Apparently all the tests are green now. But only the usersettings one related to theming is failing. It seems that if you apply the change I mention above it should fix that test as well.

test/e2e.sh
```diff
@@ -39,10 +39,10 @@ runWebserver 8090
...
-runWebserver 8080 public /bundle.js
+echo "Js Test App"
```
ndricimrr, line 42: In addition if we are not using Fiddle anymore for the e2e tests than we might also have to change something in the **testCompatibility.sh** file. 
I remember vaguely that @viktorsperling was working on something there [fixing the broken compatibility tests](https://github.com/SAP/luigi/pull/2780/commits/b0d66673744fb02d948c98fe027242d469d86650). It then seems like we might need to add the new 4500 js-app in [there](https://github.com/SAP/luigi/blob/master/scripts/testCompatibility.sh#L58) as well so that the compatibility tests keep working.

ndricimrr (commented):
Looks good. Seems to be a pretty good fix for the tests. 👍 

Had some comments that might be worth taking a look at.

test/e2e.sh
JohannesDoberer: First we need to release this branch and after that we can adjust the script for the compatibility tests. 
Created a [followup ticket](#2872) for this.

ndricimrr (approved):
Looks good. Nice work 👍 ✅

PR #2894, opened 2022-09-08
Remove html, head and body tags

### LINKED ISSUE #2885: Remove html/head/body tags from web components examples
https://github.com/SAP/luigi/blob/master/website/fiddle/public/wc/ui5/ui5example.js
https://github.com/SAP/luigi/blob/master/website/fiddle/public/wc/luigiExampleWC.js
https://github.com/SAP/luigi/blob/master/website/fiddle/public/wc/list.js

wdoberschuetz:
**Description**
See #2885 

**Related issue(s)**
Resolves #2885


website/fiddle/public/wc/luigiExampleWC.js
```diff
@@ -4,36 +4,12 @@ import "./f_input.js";
...
-                    var val = ref.getAttribute(toggleAttrs[i]);
-                    if(val === 'true') {
-                        setElAttr(id, toggleAttrs[i], 'false');
-                    } else if (val === 'false') {
-                        setElAttr(id, toggleAttrs[i], 'true');
-                    }
-                }
-            }
-       }
-    </script>
```
ndricimrr, line 34: Do we need to remove the script as well ?

ndricimrr (commented):
Looks good. Just had one comment

website/fiddle/public/wc/luigiExampleWC.js
wdoberschuetz: Hmm, I looked at it and have found that it has no uses, I see now that it exists as well in ```docs\backup\web-component-to-be-used-in-expert-page.md``` as a part of an example. I guess it could stay or the example should be updated accordingly to what I have already removed.

Which one should it be, do you know? :)

ndricimrr: I think we'd rather keep this. It might be used in another example somewhere I would suppose.

ndricimrr (approved):
Looks good 👍

PR #2897, opened 2022-09-14
Update Example Angular version 12

ndricimrr:
**Description**

- Updates version to 12
**Related issue(s)**


wdoberschuetz (approved):
Would be cool if we could udpate to the newest version but no pressure :).

PR #2901, opened 2022-09-19
Plus sign not decoded correctly when adding and then getting nodeParams

### LINKED ISSUE #2896: Wrong encoding for node params
**Description**

When using `addNodeParams` from a MFE and then using `getNodeParams` afterwards from the same MFE, the parameters can be encoded wrongly and the output of  `getNodeParams`  is different from the input of `addNodeParams`.

**Expected result**

The output of `getNodeParams` is the same as the input of `addNodeParams`.

**Actual result**

Spaces in the input to `addNodeParams`  will be transformed to `+` in the output of `getNodeParams`. Might also affect other characters.

**Steps to reproduce**

1. Call `addNodeParams({q: "test abc"})`
2. Wait for context update 
3. Call `getNodeParams()` 
4. See the result: `{q: "test+abc"}`

Changing `shouldDesanitise` to `true` in `getNodeParams` has no effect.

**Troubleshooting**

dxp metadata-registry 476

wdoberschuetz:
- replace ```+``` with spaces before decodeURIComponent is used

**Related issue(s)**
Resolves #2896


ndricimrr (changes requested):
I tested it locally and it doesn't seem to yield the expected result. 
Screenshot 2022-10-17 at 14 00 44

When I get the node params back I should be getting the same thing that was added with the space and not the % encoded space. 

In addition I noticed the newly added **%2520** which kinda seems like double encoding from [this reference here](https://stackoverflow.com/a/16085190/8250726). I suppose that shouldn't be the case and that you are encoding twice when **adding** the node params; that might be redundant.

ndricimrr (commented):
There seem to be **21** file changes here and the majority of them not related to the issue. Please remove/undo all the unrelated changes
like **.gitignore**, multiple **package-lock.json**'s **package.json** changes **angular.json** changes etc.

core/src/utilities/helpers/routing-helpers.js
```diff
@@ -54,7 +48,7 @@ class RoutingHelpersClass {
   parseParams(paramsString) {
```
ndricimrr, line 48: Can you also add a test case for this scenario on: **core/test/utilities/helpers/routing-helpers.spec.js**
for the parseParams function.

PR #2930, opened 2022-10-13
History handler for modals

### LINKED ISSUE #2802: History entry for modals
When opening a modal micro frontend and showModalPathInUrl is true (https://docs.luigi-project.io/docs/navigation-parameters-reference/?section=showmodalpathinurl), a browser history entry should be created, so that if the user does a history back, the modal is closed but the main mfe will stay the same.

(extension-manager-189)

JohannesDoberer:
Fixes #2802


core/src/App.svelte
```diff
@@ -942,7 +945,7 @@
     const showModalPathInUrl = LuigiConfig.getConfigBooleanValue(
       'routing.showModalPathInUrl'
     );
-    if (showModalPathInUrl) {
+    if (showModalPathInUrl && mfModalList.length===1) {
```
ndricimrr, line 948: Can you add a comment here as to why you selected 1 specifically. It's a very specific reason and might be lost over time.

core/src/services/routing.js
```diff
@@ -659,10 +660,17 @@ class RoutingClass {
           url.hash = url.hash.slice(0, queryParamIndex);
         }
         url.hash = `${url.hash}${queryParamSeparator}${RoutingHelpers.encodeParams(params)}`;
+        if (!url.hash.includes('historyState')) {
+          url.hash += `&historyState=${history.length}`;
+        }
       } else {
-        url.search = `?${RoutingHelpers.encodeParams(params)}`;
+        const search = new URLSearchParams(url.search);
+        if (!search.get('historyState')) {
```
ndricimrr, line 668: Could this be an issue if you have 
**http://localhost:4200/projects/pr2/settings?~foo=historyState........&historyState=5**

core/src/utilities/helpers/generic-helpers.js
```diff
@@ -400,6 +400,28 @@ class GenericHelpersClass {
   isString(value) {
     return typeof value === 'string' || value instanceof String;
   }
+
+  removeKeyValueFromURL(urlObj, keyName) {
```
ndricimrr, line 404: Can you add a JSDOC type comment on this method. A bit hard to read what it does.

ndricimrr, line 404: Would be nice to have a unit test on this function. I for myself am not able to read all possible scenarios on this one.
(If we change this function later to something more simple its functionality should be the same, so won't need to re-write the test again).

 (Maybe to make the test easier just get **LuigiConfig.getConfigBooleanValue('routing.useHashRouting');** out of it and pass it as a parameter so you don't have to do extra mocks)

core/src/App.svelte
```diff
@@ -395,7 +395,10 @@
         preservedViews = [];
         Iframe.removeInactiveIframes(node);
       }
-      closeModal();
+      mfModalList.forEach((m, index) => {
```
ndricimrr, line 398: reminder: check if this works for multiple modals when hitting the back button

ndricimrr (commented):
Looks good in general. Left some comments.

core/src/services/routing.js
```diff
@@ -662,18 +665,21 @@ class RoutingClass {
       } else {
         url.search = `?${RoutingHelpers.encodeParams(params)}`;
       }
-      history.replaceState(window.state, '', url.href);
+      if (!sessionStorage.getItem('historyState')) {
```
ndricimrr, line 668: Is **!sessionStorage.getItem('historyState')** check necessary here ? 🤔 
We check for **(sessionStorage.getItem('historyState'))** in line 643 above. Seems like in this line we are sure that history state does not exist already, otherwise code execution wouldn't arrive here at all.

core/src/services/routing.js
```diff
@@ -640,6 +640,9 @@ class RoutingClass {
   }
 
   appendModalDataToUrl(modalPath, modalParams) {
+    if (sessionStorage.getItem('historyState')) {
```
ndricimrr, line 643: can you please add a one line comment here as to why we check for this historyState, what it means and why we need to return if it exists. I cannot understand. I guess I would forget in a month or two.

core/src/services/routing.js
```diff
@@ -662,18 +665,21 @@ class RoutingClass {
...
-      history.replaceState(window.state, '', url.href);
+      if (!sessionStorage.getItem('historyState')) {
+        sessionStorage.setItem('historyState', history.length);
+      }
+      history.pushState(window.state, '', url.href);
     }
   }
 
-  removeModalDataFromUrl() {
+  removeModalDataFromUrl(closeBtnPressed) {
```
ndricimrr, line 675: Could you please add a JSDOC description on this function

ndricimrr (commented):
Looking good. Left some small comments

core/src/services/routing.js
JohannesDoberer: 👍

core/src/services/routing.js
JohannesDoberer: Good point.. was wrong anyway..

core/src/services/routing.js
JohannesDoberer: 👍

core/src/services/routing.js
```diff
@@ -640,6 +640,9 @@ class RoutingClass {
   }
 
   appendModalDataToUrl(modalPath, modalParams) {
+    const url = new URL(location.href);
+    //necessary to jump to the last history entry
```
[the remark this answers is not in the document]
JohannesDoberer: Yes we need it. When we close the modal via close btn we are not at the end of the history.
If you open a modal again (and you are not at the end of the history) you'll have a wrong state and history.go(historyState-history.length) in removeModalDataFromUrl fn navigate to the wrong history entry.

ndricimrr (commented):
Left some comments.

core/src/services/routing.js
```diff
@@ -696,7 +709,20 @@ class RoutingClass {
...
+    if (historyState && closeBtnPressed) {
+      window.addEventListener(
+        'popstate',
+        e => {
+          history.pushState(window.state, '', url.href);
+          history.back();
+        },
+        { once: true }
+      );
+      history.go(historyState - history.length);
```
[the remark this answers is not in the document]
JohannesDoberer: No it's not always -1. When you navigate from modal to modal it is a higher negative value.

core/src/services/routing.js
```diff
@@ -651,7 +657,7 @@ class RoutingClass {
       if (modalParams && Object.keys(modalParams).length) {
         params[`${modalParamName}Params`] = JSON.stringify(modalParams);
       }
-      const url = new URL(location.href);
+      const url = urlObj;
```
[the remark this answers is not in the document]
JohannesDoberer: Don't change parameter itself just use the value

core/src/App.svelte
```diff
@@ -1509,7 +1519,7 @@
       if ('luigi.navigation.back' === e.data.msg) {
         const mfModalTopMostElement = mfModalList[mfModalList.length - 1];
         if (IframeHelpers.isMessageSource(e, mfModalTopMostElement && mfModalTopMostElement.modalIframe)) {
-          closeModal(mfModalList.length - 1);
+          closeModal(mfModalList.length - 1, true);
```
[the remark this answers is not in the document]
JohannesDoberer: **browser back** button is not the same like `luigi.navigation.back`

core/src/App.svelte
```diff
@@ -970,26 +975,29 @@
...
         const showModalPathInUrl = LuigiConfig.getConfigBooleanValue(
           'routing.showModalPathInUrl'
         );
-        if (showModalPathInUrl) {
-          Routing.removeModalDataFromUrl();
+        // only remove the modal path in URL when closing the first modal
+        if (showModalPathInUrl && mfModalList.length===1) {
+          Routing.removeModalDataFromUrl(closeBtnPressed);
         }
         resetMicrofrontendModalData(index);
```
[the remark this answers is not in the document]
JohannesDoberer: why not 👍

ndricimrr (approved):
Tested and working well 👍

PR #2933, opened 2022-10-16
Add tests for Fiori 3 Theme

rafalgamon:
**Description**

-Add tests for Fiori 3 Theme


core/src/UserSettingsEditor.svelte
```diff
@@ -214,6 +214,22 @@
     }
   }
 
+  function createLinkForStyle(name){
```
JohannesDoberer, line 217: Please add entries for the editor only via configuration.
https://docs.luigi-project.io/docs/user-settings?section=example

This code would be displayed also in luigi apps which are not using fiori theme.

PR #2935, opened 2022-10-18
Update angular support lib to angular v14

JohannesDoberer:


JohannesDoberer:
@luigi-project/client-support-angular@1.25.2-dev.202210181516

PR #2963, opened 2022-11-09
LuigiClient type declaration for web components

JohannesDoberer:


client/luigi-element.d.ts
```diff
@@ -0,0 +1,411 @@
...
+}
+
+/**
+ * Html string processing according to luigi functionality.
+ * Also useful in combination with LitElement VS Code plugins.
+ *
+ * @param {String} literal The literal to process.
+ * @returns {String} Returns the processed literal.
+ */
+export declare const html: (strings: TemplateStringsArray, ...values: unknown[]) => string;
```
wdoberschuetz, line 394: (question) In the luigi-element.js this method receives ```keys``` [luigi-element.js: L148](https://github.com/SAP/luigi/blob/master/client/src/luigi-element.js#L148). But it's declared as values here. Should it be like this? Maybe I'm overeager but there's no param for this as well. I presume it's all by design, if so, ignore my comment. :)

wdoberschuetz (approved):
LGTM

PR #2966, opened 2022-11-11
Remove unused dependencies & update angular example old dependencies

ndricimrr:
**Description**
Maintaining dependency alerts of dependencies that aren't even used should be avoided as much as possible.

- Removes some unused dependencies to keep the example minimal, and updates some versions 
- also Solves the #2962 bump for loader-utils 

**Related issue(s)**

Resolves #2962


wdoberschuetz (approved):
LGTM.

PR #2984, opened 2022-11-15
Fix unsaved changes modal not closing after browser back 

### LINKED ISSUE #2964: Unsaved changes modal does not disappear using browser back
How to reproduce:

1. Let’s assume user has some not saved data in the form and he clicks back arrow in the browser
2. ‘Unsaved Changes Detected’ dialog appears and user clicks ‘Yes’ (so he wants to loose his data in the form and go back)

Result:
Only url in the browser address bar changes but dialog is still visible.

Expected:

Confirmation modal should be closed.

ndricimrr:
**Description**
When going back the handling is done but the dirty bit is not set back to false leading into 'endless loop' where the modal is re-created again and again after clicking yes.

- Using **getUnsavedChangesModalPromise** instead, to properly set dirt state back to false.
- test
- go back on empty history

**Related issue(s)**

Fixes #2964


JohannesDoberer:
Needs clarificationb:
What should happen if the unsafed change modal is active and user clicks on _browser_ back button?
Currently modal stays active and the user is able to go back through the history (without route changes).

JohannesDoberer (approved):
👍 LGTM

PR #3062, opened 2022-12-13
History handling for modals v2

### LINKED ISSUE #3029: History handling for modals v2
Due to the limitation of history.length to 50 in chrome and firefox we decided to go with history.state object.
In history.state will stored the route before the modal was opened.

JohannesDoberer:
Fixes #3029


JohannesDoberer:
Closed because branch created not from master

PR #3072, opened 2022-12-21
History handling using history state

JohannesDoberer:


core/src/services/routing.js
```diff
@@ -740,19 +740,46 @@ class RoutingClass {
       url.search = finalUrl;
     }
     // only if close modal [X] is pressed
-    if (historyState && isClosedInternal) {
+    if (history.state && history.state.modalHistoryLength >= 0 && isClosedInternal) {
+      const modalHistoryLength = history.state.modalHistoryLength;
+      const path = history.state.pathBeforeHistory;
+      let isModalHistoryHigerAsHistoryLength = history.state.isModalHistoryHigerAsHistoryLength;
```
ndricimrr, line 746:
```suggestion
      let isModalHistoryHigherThanHistoryLength = history.state.isModalHistoryHigerAsHistoryLength;
```

core/src/services/routing.js
```diff
@@ -740,19 +740,46 @@ class RoutingClass {
...
       window.addEventListener(
         'popstate',
         e => {
-          history.pushState(window.state, '', url.href);
-          history.back();
+          if (isModalHistoryHigerAsHistoryLength) {
```
ndricimrr, line 750:
```suggestion
          if (isModalHistoryHigherThanHistoryLength) {
```

core/src/services/routing.js
```diff
@@ -740,19 +740,46 @@ class RoutingClass {
...
         { once: true }
       );
-      history.go(historyState - history.length);
+
+      if (history.state.historygap === history.length - history.state.modalHistoryLength) {
+        history.go(-history.state.modalHistoryLength);
+      } else {
+        if (history.state.modalHistoryLength > history.length) {
+          const historyMaxBack = history.length - 1;
+          isModalHistoryHigerAsHistoryLength = true;
```
ndricimrr, line 770:
```suggestion
          isModalHistoryHigherThanHistoryLength = true;
```

ndricimrr (commented):
Small typo

core/src/services/routing.js
ndricimrr, line 746: **history.state.isModalHistoryHigerAsHistoryLength**:

We access the property **isModalHistoryHigerAsHistoryLength** of history state, but seems we never set its value elsewhere or I missed it somehow ? Where does it come from 🤔

ndricimrr (approved):
LGTM 👍

PR #3136, opened 2023-01-27
hideTabNavAutomatically

JohannesDoberer:


core/src/services/routing.js
```diff
@@ -395,6 +390,14 @@ class RoutingClass {
         } else if (cnode.tabNav === false) {
           tabNavInherited = false;
           break;
+        } else if (GenericHelpers.isObject(cnode.tabNav)) {
+          if (cnode.tabNav.hideTabNavAutomatically === true && cnode.children.length === 1) {
+            tabNavInherited = false;
+            break;
+          } else {
```
ndricimrr, line 397: I think this **else** part needs some re-work:
For example with the following **tabNav** definition **tabNav** is enabled too, just because it is an **Object**: 
EXAMPLE: 

 ` tabNav: { blabla: false },` this is equal to  `tabNav: true` which I suppose we shouldn't allow
 Also `tabNav: {}` also enables **tabNav** which I suppose we should also not allow 🤔 ?

docs/navigation-advanced.md
```diff
@@ -423,8 +423,9 @@ appSwitcher = {
 Tab-style navigation in Luigi can be displayed directly above the micro frontend area, providing you with additional menu options. When you put tab navigation nodes into a [category](navigation-configuration.md#category), they will be rendered in a drop-down. Add this parameter to your configuration to create tab navigation nodes:
 
 ### tabNav
-- **type**: boolean
+- **type**: boolean or Object
 - **description**: renders the children of the node as a horizontal navigation bar. Sub-children are not supported. When you categorize nodes you will get a drop-down menu in the horizontal navigation.
+When you have only one child in the horizontal navigation bar you can specify an object instead of boolean with `{hideTabNavAutomatically:true}` and the horizontal navigation bar will be hidden automatically.
```
ndricimrr, line 428: It seems to suggest that when:
`tabNav: {hideTabNavAutomatically:true}`

it indirectly means that **tabNav** is automatically assumed  to be set **true**. 

I guess we should somehow state in docu that if you have the following:
`tabNav: {hideTabNavAutomatically:true}`

then it means that **tabNav** is inherently also activated when hideTabNavAutomatically = true

Otherwise it makes it ambiguous, like what happens when `tabNav: { hideTabNavAutomatically: false }`. Although I don't know how we could phrase it in words. Maybe @alexandra-simeonova can help with that.

ndricimrr (commented):
Working well. 
Found some edge cases worth looking at.

core/src/services/routing.js
```diff
@@ -395,6 +390,14 @@ class RoutingClass {
         } else if (cnode.tabNav === false) {
           tabNavInherited = false;
           break;
+        } else if (GenericHelpers.isObject(cnode.tabNav)) {
+          if (cnode.tabNav.hideTabNavAutomatically === true && cnode.children.length === 1) {
```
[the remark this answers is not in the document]
JohannesDoberer: Thx!

core/src/services/routing.js
JohannesDoberer: Yes, indeed. To optimistic ;).

alexandra-simeonova (approved):

ndricimrr (approved):

PR #3152, opened 2023-02-06
Css-fix: tabNav + breadcrumb

### LINKED ISSUE #3132: Tabnav position /w breadcrumb
Currently tabnav and breadcrumb are overlapping, if shown simultaneously.

Expected: if breadcrumbs are shown, tabnav and mfe container top (and potentially other dom elements, e.g. splitview) should shifted down by breadcrumbs height.

Could be done by following css (for tabnav and iframecontainer, other parts need to be checked as well):
```
.lui-breadcrumb #tabsContainer { 
    top: calc(var(--luigi__shellbar--height) + var(--luigi__breadcrumb--height));
} 

.lui-breadcrumb .iframeContainer.iframeContainerTabNav { 
    top: calc(var(--luigi__shellbar--height) + var(--luigi__breadcrumb--height) + 2.75rem);
}
```

Note: the '2.75rem' is the assumed height of the tabnav and defined as scss variable (https://github.com/SAP/luigi/blob/7a41a94764a73b322dee89d89faf93c2fb8a0a39/core/src/styles/_variables.scss#L5). It would be better to define it as a 'real' css variable so it can be overwritten in custom css.

JohannesDoberer:
Fixes #3132


core/src/App.svelte
```diff
@@ -2134,6 +2136,18 @@
...
+    top: calc(
+      var(--luigi__shellbar--height) + var(--luigi__breadcrumb--height)
+    );
+  }
+  :global(.lui-breadcrumb .iframeContainer.iframeContainerTabNav) {
+    top: calc(
+      var(--luigi__shellbar--height) + var(--luigi__breadcrumb--height) +
+        var(--luigi__horizontal-nav--height)
+    );
+    border: 1px solid red;
```
ndricimrr, line 2149: I believe this was added for testing only. I guess it should be removed then
```suggestion
```

ndricimrr (commented):
Looks good. Tested and found no issues. One minor suggestion on the code

ndricimrr (approved):

PR #3226, opened 2023-03-16
Prevent rerendering of wc

JohannesDoberer:
- Simple wc gets a context update without rerendering
- Compound views just not rerendered


core/src/services/routing.js
```diff
@@ -623,7 +622,26 @@ class RoutingClass {
     window.open(updatedExternalLink.url, updatedExternalLink.sameWindow ? '_self' : '_blank').focus();
   }
 
+  getGeneratedWCId(navNode) {
```
ndricimrr, line 625: Can you add some JSDoc comments for the function header here. The wc part is quite uncommented. would be nice to have some here

ndricimrr (approved):

Issue #3230, opened 2023-03-17
No confirmation dialog on browser reload when dirtyState is set to true


aaasarthak:
**Description**
When the `dirtyStatus` is set to true, the browser back button shows the confirmation dialog but the browser reload doesn't.

browser used - Chrome

**Expected result**

On click of browser reload also, it should show the confirmation dialog for unsaved changes

**Actual result**

No confirmation dialog

**Steps to reproduce**

- use `setDirtyStatus()` to set the dirty status of the client 
- Try reloading the page using browser reload button
- Observe that there is no confirmation dialog for unsaved data changes

JohannesDoberer:
Hello @aaasarthak ,
we expose a core-api function where you can check which dirty status is currently set and e.g. avoid reloading the page.
Let me know if this helps you.

PR #3236, opened 2023-03-24
Get dirty status

JohannesDoberer:
#3230


core/src/core-api/ux.js
```diff
@@ -166,6 +166,16 @@ class LuigiUX {
   removeBackdrop() {
     new linkManager().sendPostMessageToLuigiCore({ msg: 'luigi.remove-backdrop' });
   }
+
+  /**
+   * Returns the dirty status, which is set by the client. Default the dirty status is `false`.
```
alexandra-simeonova, line 171:
```suggestion
   * Returns the dirty status, which is set by the Client via [setDirtyStatus](luigi-client-api.md#setdirtystatus). By default, the dirty status is `false`.
```

ndricimrr (approved):

alexandra-simeonova (commented):
Normally, when you change the `ux.js`file, there should be autogenerated docu as well. I see in this case the docu changes were not reflected in the `luigi-core-api.md` file automatically. Can you try running `npm install` or `lerna run docu` as described [here](https://github.com/SAP/luigi/blob/main/docs/content-guidelines.md#API-documentation) in order to auto-generate the doc changes?

alexandra-simeonova (approved):

PR #3241, opened 2023-03-29
Fix confirmation modal on reject result

### LINKED ISSUE #3229: Console error on clicking on "No" in the confirmation dialog
**Description**

- I have a form in my client(microfrontend application), where if the user doesn't save before navigating I want to give a confirmation dialog for the unsaved changes.
- For which I am using `setDirtyStatus()` in the client application which gives the confirmation dialog if the user tries to navigate from the shellbar navigation options.
- But if we click on "No" in the confirmation dialog, it closes the dialog with a console error.

**Expected result**

It should not throw the console error on selecting "No" in the confirmation dialog

**Actual result**

It throws the console error

**Steps to reproduce**

- setDirtyStatus as true from your microfrontend
- Try navigating outside the application
- There will be a confirmation modal to tell that unsaved changes will be lost
- Select "No" on that model
- Observe that the dialog closes and application behavior is as expected, but there is an error in the console.

**Screenshots**

image
image

ndricimrr:
**Description**
Since there is nothing done upon confirmation modal rejection, we ignore rejection handlers. However they throw a harmless exception which pollutes the console. This PR adds empty handlers to avoid such exceptions from being thrown.

- Add rejection handler functions to avoid [unhandledrejection](https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event)

**Related issue(s)**

Fixes #3229


JohannesDoberer (approved):
LGTM

core/src/navigation/LogoTitle.svelte
```diff
@@ -100,25 +111,32 @@
...
+      },
+      () => {}
+    );
   }
 
   export function goToRoot() {
-    getUnsavedChangesModalPromise().then(() => {
-      Routing.navigateTo('/');
-    });
+    console.log('test 2');
```
ndricimrr, line 133:
```suggestion
```

PR #3246, opened 2023-03-29
Fix failing prettify setup

ndricimrr:
**Description**
Prettier Git diff checking was done with [a package](https://github.com/kandhavivekraj/git-changed-files#basebranch) which had 'master' as default branch, causing the prettifying process to fail as we switched to 'main'

- set default to main
- change to new cypress e2e test folder 

**Related issue(s)**


JohannesDoberer (approved):
👍

PR #3269, opened 2023-04-20
Fix node version to 18.15.0

ndricimrr:
**Description**

We set node 18 as Pipeline standard, and it picks the latest current LTS. But it seems some new node was released recently, and that causes an issue with the pipeline.

Failing for node 18.16.0
https://github.com/SAP/luigi/actions/runs/4753488852/jobs/8447292296#step:2:10

Succeding with node 18.15.0
https://github.com/SAP/luigi/actions/runs/4747489631/jobs/8432557852#step:2:10

This PR fixes the node version to 18.15.0 so we keep with a consistent node until we decide to upgrade

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -22,7 +22,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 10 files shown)


JohannesDoberer (approved):
👍 !!!

PR #3276, opened 2023-05-03
Check if showModalPathInUrl is defined

### LINKED ISSUE #3260: ModalPathData will not be removed from URL
Given:
Open a view in a modal and update the modal path in the URL `LuigiClient.linkManager().updateModalPathInternalNavigation('microfrontend')`. Additionally, ensure `showModalPathInUrl` is `false` or undefined.

Or use the lib feature
[updateModalDataPath=false](https://docs.luigi-project.io/docs/framework-support-libraries?section=autorouting-for-modals)
Close the modal via LuigiClient.uxManager().closeCurrentModal() or click on the "X" button.

Routing config:
```javascript
routing:{
        useHashRouting: true,
        showModalPathInUrl: false,
        modalPathParam: 'mymodal'
}
```

Path will be written to the URL and not removed when you close the modal.

Expected:
Should not be possible to update the path when `showModalPathInUrl` is not `true`

JohannesDoberer:
`updateModalDataPath`/`updateModalPathInternalNavigation` should only be possible if `showModalPathInUrl` is defined and set to `true`.
Fixes #3260

### DIFF
--- a/core/src/App.svelte
+++ b/core/src/App.svelte
@@ -1640,6 +1640,10 @@
       }
 
       if ('luigi.navigation.updateModalDataPath' === e.data.msg) {
+        if(!LuigiConfig.getConfigBooleanValue('routing.showModalPathInUrl')){
+          console.warn('"showModalPathInUrl" will be ignored. Property is not defined or not set to `true`.');
+          return;
+        }
         if (isSpecialIframe) {
           const route = GenericHelpers.addLeadingSlash(
             buildPath(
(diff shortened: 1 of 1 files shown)


core/src/App.svelte
```diff
@@ -1640,6 +1640,10 @@
       }
 
       if ('luigi.navigation.updateModalDataPath' === e.data.msg) {
+        if(!LuigiConfig.getConfigBooleanValue('routing.showModalPathInUrl')){
+          console.warn('"showModalPathInUrl" will be ignored. Property is not defined or not set to `true`.');
```
ndricimrr, line 1644:
```suggestion
          console.warn('Updating path of the modal ignored. Property "showModalPathInUrl" is not defined or not set to `true`.');
```
Maybe sth like this is better, since showModalPathInUrl is not the one ignored specifically

ndricimrr (approved):

PR #3339, opened 2023-06-29
Write updated modal settings to url

### LINKED ISSUE #3338: [low prio] incorrectly rendered rounding when setting the size via updateLuigiModalSettings
When setting the size via updateLuigiModalSettings to 'l', we receive an incorrectly rendered rounding in the modal

image

If I set the size to 'l' in the URI parameter, the rounding is rendered correctly.

Before:
image

After setting size to 'l' in URI parameter:
 
image

_To reproduce, open an arbitrary template in the Hyperspace Portal(1. Create a project or open one where you are a member 2. Click Components 3. Click Create 4. Select an arbitrary template)_

JohannesDoberer:
Fixes #3338

### DIFF
--- a/client/src/linkManager.js
+++ b/client/src/linkManager.js
@@ -170,18 +170,21 @@ export class linkManager extends LuigiClientBase {
   }
 
   /**
-   * Update current title and size of a modal.
+   * Update current title and size of a modal. If `routing.showModalPathInUrl` is set to true, the URL will be updated with the modal settingds data.
+   * In addition you can specify if a new history entry wil lbe created with the updated URL.
    * @memberof linkManager
    * @param {Object} updatedModalSettings possibility to update the active modal.
    * @param {Object} updatedModalSettings.title update the `title` of the active modal.
    * @param {Object} updatedModalSettings.size update the `size` of the active modal.
+   * @param {boolean} addHistoryEntry adds an entry in the history, by default it's `false`.
    * @example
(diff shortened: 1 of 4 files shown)


client/src/linkManager.js
```diff
@@ -170,18 +170,21 @@ export class linkManager extends LuigiClientBase {
   }
 
   /**
-   * Update current title and size of a modal.
+   * Update current title and size of a modal. If `routing.showModalPathInUrl` is set to true, the URL will be updated with the modal settingds data.
```
alexandra-simeonova, line 173:
```suggestion
   * Updates the current title and size of a modal. If `routing.showModalPathInUrl` is set to `true`, the URL will be updated with the modal settings data.
```

client/src/linkManager.js
```diff
@@ -170,18 +170,21 @@ export class linkManager extends LuigiClientBase {
...
+   * In addition you can specify if a new history entry wil lbe created with the updated URL.
```
alexandra-simeonova, line 174:
```suggestion
   * In addition, you can specify if a new history entry will be created with the updated URL.
```

core/src/Modal.svelte
```diff
@@ -147,6 +149,7 @@
       height = settingsHeight;
       width = settingsWidth;
     } else {
+      elem.classList.remove('lui-modal-fullscreen');
```
hardl, line 152: why only remove in the else block?

hardl (changes requested):

hardl (approved):
LGTM

PR #3352, opened 2023-07-14
onLoad functionality for mfe

### LINKED ISSUE #3344: Container: "onload" event for mfe's
some dedicated event when mfe is "ready", we have already sth. for iframes but not for wc

JohannesDoberer:
Resolves #3344


container/src/LuigiContainer.svelte
```diff
@@ -5,6 +5,8 @@
   export let context;
   export let label;
   export let webcomponent;
+  export let can_mfe_initialized;
+  export let mfe_initialized_ready;
```
ndricimrr, line 9: What does `can_mfe_initialized` mean. Also `mfe_initialized_ready`. I have a general idea but it'd be good to have small inline comments as we might forget later on.

container/src/LuigiContainer.svelte
```diff
@@ -5,6 +5,8 @@
   export let context;
   export let label;
   export let webcomponent;
+  export let can_mfe_initialized; // if `true` at LuigiContainer tag, LuigiContainer sends an event `initialzed` to mfe. Mfe is immediately ready.
+  export let mfe_initialized_ready; // will be set to true if `can_mfe_initialized` is set to `true` otherwise it will be `true` if webcomponent is attatch to the container.
```
ndricimrr, line 9: Could you also add this to `container/public/LuigiContainer.svelte.d.ts` with comments and typings like for the other properties

ndricimrr: and maybe add the documentation label afterwards so Alex can check it afterwards.

container/public/LuigiContainer.svelte.d.ts
```diff
@@ -19,6 +19,16 @@ export default class LuigiContainer extends HTMLElement {
...
 
+  /**
+   * If set to `true` a microfrontend will be immediately initialized. LuigiContainer sends an event `initialized` to the microfrontend.
+   */
+  can_mfe_initialized: boolean;
+
+  /**
+   * Will be set to `true` if `can_mfe_initialized` is set to `true` otherwise it will be `true` if the mfe webcomponent is attached to the container.
+   */
+  mfe_initialized_ready: boolean;
```
alexandra-simeonova, line 30:
```suggestion
   * If set to `true`, a micro frontend will be immediately initialized. LuigiContainer sends an event `initialized` to the micro frontend.
   */
  can_mfe_initialized: boolean;

  /**
   * Will be set to `true` if `can_mfe_initialized` is set to `true`. Otherwise, it will be `true` if the micro frontend webcomponent is attached to the container.
   */
  mfe_initialized_ready: boolean;
```

container/public/LuigiContainer.svelte.d.ts
```diff
@@ -19,6 +19,16 @@ export default class LuigiContainer extends HTMLElement {
    */
   webcomponent: boolean;
 
+  /**
+   * If set to `true` a microfrontend will be immediately initialized. LuigiContainer sends an event `initialized` to the microfrontend.
+   */
+  can_mfe_initialized: boolean;
+
+  /**
+   * Will be set to `true` if `can_mfe_initialized` is set to `true` otherwise it will be `true` if the mfe webcomponent is attached to the container.
```
ndricimrr, line 28: I find it a bit misleading. 

> **Will be set to true** if can_mfe_initialized is set to true **otherwise it will be `true`** if the mfe webcomponent is attached to the container.

        (can_mfe_initialized = true) => (mfe_initialized_ready = true)  then also: 
         <=> (can_mfe_initialized = false) => (mfe_initialized_ready = false) so aren't 
         can_mfe_initialized = mfe_initialized_ready so they are the same, then why use two?
         
Or maybe the wording is just not right.

ndricimrr, line 28: Could you also add that this WC related. I think it doesn't apply to iframe based container microfrontends, right ?. Something like: _To be used for webcomponent based microfrontends_.

container/src/LuigiContainer.svelte
```diff
@@ -94,12 +102,32 @@
...
+      mainComponent.addEventListener('wc_ready', () => {
+        if (
+          !(mainComponent as any)._luigi_mfe_webcomponent
+            ?.deferLuigiClientWCInit
+        ) {
+          mfe_initialized_ready = true;
+          dispatchLuigiEvent(Events.INITIALIZED, {});
+        }
+      });
+    }
```
ndricimrr, line 130: The whole thing seems a bit duplicated. https://github.com/SAP/luigi/issues/3344 seems to suggest that this feature should be about WC only, since it seems we handled it for iframes already. So I guess the whole thing should fit under **isWebComponent()** ?:
```
// here
if (isWebComponent) {
      mainComponent.innerHTML = '';
      webcomponentService.renderWebComponent(viewurl, mainComponent, ctx, {});
      mainComponent.addEventListener('wc_ready', () => {
          if (
            !(mainComponent as any)._luigi_mfe_webcomponent
              ?.deferLuigiClientWCInit
          ) {
            mfe_initialized_ready = true;
            dispatchLuigiEvent(Events.INITIALIZED, {});
          }
        });

        if (canMfeInitialized()) {
              mfe_initialized_ready = true;
              setTimeout(() => {
                    dispatchLuigiEvent(Events.INITIALIZED, {});
              });
        }
}
```

JohannesDoberer: Talked with Philipp about the implementation. It's fine.
Regarding "duplication":
`if(!(mainComponent as any)._luigi_mfe_webcomponent?.deferLuigiClientWCInit)` checks if webcomponents, which are extending LuigiElement, does not have the option to defer the LuigiClient initialization. In the suggested code the dispatchLuigiEvent would be called twice if a LuigiContainer tag has the property `can_mfe_initialized="true"` and `webcomponent="true` and the webcomponent has no option `deferLuigiClientWCInit` or it is set to false (which is the default).

container/public/LuigiContainer.svelte.d.ts
JohannesDoberer: ok.. maybe I need other names :)

ndricimrr (approved):
Looks good

alexandra-simeonova (approved):

container/src/LuigiCompoundContainer.svelte
```diff
@@ -1,8 +1,10 @@
 <svelte:options tag={null} accessors={true} />
 
 <script lang="ts">
-  // export let viewurl;
+  export let viewurl;
   export let context;
+  // if `true` at LuigiContainer tag, LuigiContainer sends an event `initialzed` to mfe. Mfe is immediately ready.
```
ndricimrr, line 6:
```suggestion
  // if `true` at LuigiContainer tag, LuigiContainer sends an event `initialized` to mfe. Mfe is immediately ready.
```

container/src/LuigiContainer.svelte
```diff
@@ -5,6 +5,8 @@
   export let context;
   export let label;
   export let webcomponent;
+  // if `true` at LuigiContainer tag, LuigiContainer sends an event `initialzed` to mfe. Mfe is immediately ready.
```
ndricimrr, line 8:
```suggestion
  // if `true` at LuigiContainer tag, LuigiContainer sends an event `initialized` to mfe. Mfe is immediately ready.
```

container/test-app/index.html
```diff
@@ -417,6 +513,36 @@
...
+      }
+      let readyTest3 = document.getElementById('calendar');
+      if (!readyTest3.isMfeInitialized) {
+        readyTest3.addEventListener(MFEventID.INITIALIZED, () => {
+          console.log('calendar is ready now! 3');
+        });
+      } else {
+        console.log('calendar immediately ready 3');
+      }
+      let dashboard2 = document.getElementById('dashboard2');
```
ndricimrr, line 540: This is getting very confusing. Could you change the id name to something more specific. 
I believe we will have to use them when we start making e2e-tests for container.

readyTest -> test-defer-init-false
readyTest2 -> test-defer-init-true
dashboard2 -> nested-wc-compound-container

container/public/LuigiCompoundContainer.svelte.d.ts
```diff
@@ -1,3 +1,13 @@
 export default class LuigiCompoundContainer extends HTMLElement {
   compoundConfig: any;
+
+  /**
+   * If set to `true`, a micro frontend will be immediately initialized. LuigiContainer sends an event `initialized` to the micro frontend.
+   */
+  instantInit: boolean;
```
ndricimrr, line 7: should we maybe make this **instantinit** also, same as in `LuigiContainer.svelte`

ndricimrr (commented):
Just some non-functional comments to maybe consider

container/test-app/nested.js
```diff
@@ -0,0 +1,28 @@
...
+          <header><slot name="header">header</slot></header>
+          <main style="flex: auto; overflow: auto"><slot name="content">content</slot><slot></slot></main>
+          <footer><slot name="footer">footer</slot></footer>
+        </section>
+      `;
+    this._shadowRoot.appendChild(template.content.cloneNode(true));
+  }
+
+  set context(ctx) {
+    console.log('ctx', ctx);
```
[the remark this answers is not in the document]
JohannesDoberer: removed because it's not necessary for this functionality

ndricimrr (approved):

PR #3366, opened 2023-07-26
fix incomplete context for tabheader mfe

hardl:
tabheader mfe received unprocessed node context instead of fully featured one (with inherited and replaced data)


ndricimrr (approved):
tested & working well so far

ndricimrr (approved):
Tested again, works when testing

PR #3376, opened 2023-07-31
Fix undefined Type-error in Container service

ndricimrr:
**Description**

- postMessage call below :
- 
https://github.com/SAP/luigi/blob/2f80ab9cc85706d78ba0d6e4bcb8c4e725ec5a57/container/src/services/container.service.ts#L78-L83

fails with:

```
Error:      TypeError: Cannot read properties of null (reading 'postMessage')
    at messageListener (container.service.ts:78:24)
```

because the check fails to account for null/undefined target/event here:
https://github.com/SAP/luigi/blob/2f80ab9cc85706d78ba0d6e4bcb8c4e725ec5a57/container/src/services/container.service.ts#L67

**Related issue(s)**

Issue appears while attempting to run tests on Cypress environment: 
Screenshot 2023-07-31 at 18 13 21

### DIFF
--- a/.github/workflows/mend.yml
+++ b/.github/workflows/mend.yml
@@ -26,4 +26,9 @@ jobs:
       - run: npm --prefix ./core ci ./core
       - run: npm --prefix ./website/landingpage/dev install ./website/landingpage/dev
       - run: npm --prefix ./website/fiddle install ./website/fiddle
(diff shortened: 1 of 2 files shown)


JohannesDoberer (approved):
LGTM

PR #3389, opened 2023-08-07
addNodeParams, getNodeParams and setAnchor for wc luigiClient

JohannesDoberer:


container/public/LuigiContainer.svelte.d.ts
```diff
@@ -34,6 +37,11 @@ export default class LuigiContainer extends HTMLElement {
    */
   active_feature_toggle_list: string[];
 
+  /**
+   * The parameters to be passed to the web-component-based micro frontend. Will not passed to the compound children.
```
alexandra-simeonova, line 41:
```suggestion
   * The parameters to be passed to the web-component-based micro frontend. Will not be passed to the compound children.
```

container/test-app/index.html
```diff
@@ -539,6 +542,36 @@
...
+          // if (searchParams.toString() !== '') {
+          //   localhash += `?${searchParams.toString()}`;
+          // }
+          // url.hash = localhash;
+          // event.detail.keepBrowserHistory
+          //   ? window.history.pushState({}, '', url)
+          //   : window.history.replaceState({}, '', url);
+          // document
+          //   .querySelector('[data-test-id="luigi-client-api-test-01"]')
+          //   .setAttribute('node_params', `${JSON.stringify(params)}`);
```
ndricimrr, line 573:
```suggestion
```

client/luigi-element.d.ts
```diff
@@ -414,4 +414,27 @@ export interface LuigiClient {
...
+   * @returns {Object} node parameters, where the object property name is the node parameter name without the prefix, and its value is the value of the node parameter. For example `{sort: 'asc', page: 3}`
+   * @memberof LuigiClient
+   */
+  getNodeParams: (shouldDesanitise: boolean) => Object;
+  /**
+   * Sends anchor to Luigi Core. The anchor will be added to the URL.
+   * @param {string} anchor
+   * @memberof LuigiClient
+   */
+  setAnchor: (anchor: string) => void;
```
ndricimrr, line 439: I tried to test this function on simpledev but I get a weird behavior. 

Luigi config: 
```
      {
          label: 'Modal WC',
          pathSegment: 'modalwc',
          hideFromNav: false,
          loadingIndicator: {
            enabled: false
          },
          context: {
            test: 12333
          },
          openNodeInModal: true,
          hideSideNav: false,
          webcomponent: true,
          viewUrl: '/helloWorldWC.js'
        }
```


WC: helloWorld.js
```
import { LuigiElement } from "./luigi-element.js";
export default class extends LuigiElement {
  render(ctx) {
    return '<div>test</div>';
  }
  onContextUpdate(ctx) {
    console.log('context update', ctx);
  }

  afterInit(ctx) {
       this.LuigiClient.setAnchor('A')
  }
}
```
**Prerequisite**: the main microfrontend opened should be iframe based and Anchor shouldn't already be set. 

When I try to open the **Modal WC**, with the **setAnchor** being run. The modal quickly opens and then closes directly after.  If I open modal second time then it works. I guess should be checked.

client/luigi-element.d.ts
```diff
@@ -414,4 +414,27 @@ export interface LuigiClient {
...
+  /**
+   * Returns the node parameters of the active URL.
+   * Node parameters are defined like URL query parameters but with a specific prefix allowing Luigi to pass them to the micro frontend view. The default prefix is **~** and you can use it in the following way: `https://my.luigi.app/home/products?~sort=asc&~page=3`.
+   * <!-- add-attribute:class:warning -->
+   * > **NOTE:** some special characters (`<`, `>`, `"`, `'`, `/`) in node parameters are HTML-encoded.
+   * @param {boolean} shouldDesanitise defines whether the specially encoded characters should be desanitised
+   * @returns {Object} node parameters, where the object property name is the node parameter name without the prefix, and its value is the value of the node parameter. For example `{sort: 'asc', page: 3}`
+   * @memberof LuigiClient
+   */
+  getNodeParams: (shouldDesanitise: boolean) => Object;
```
ndricimrr, line 433: Again the same config as in Modal WC, when I try to get the node params inside a modal, I do not get the node params. But when the modal is opened in normal WC and not a modal it works well. 

Example:
```
import { LuigiElement } from "./luigi-element.js";
export default class extends LuigiElement {
  render(ctx) {

    return '<div>test</div>';
  }
  onContextUpdate(ctx) {
    console.log('context update', ctx);
  }

  afterInit(ctx) {
    // <------------------- getNodeParams() returns undefined --------------------> 
    console.log('context afterInit', ctx, this.LuigiClient.getNodeParams());
  }
}
```

ndricimrr (commented):
code looks good, but while testing found some weird behavior in modal view

client/luigi-element.d.ts
JohannesDoberer: Good catch. It's related to a unexpected behavior that already exists. It's discussed with Philipp that we handle this in another issue. It's just not allowed in this issue to call setAnchor in a modal.  Also because calling setAnchor in a modal is not a normal use case.

ndricimrr (approved):
Tested again and looks good

PR #3393, opened 2023-08-10
Container unit test

hardl:
Adds unit test for container and refactor bundling process


ndricimrr (approved):
Also tested on Angular example. works as expected

PR #3405, opened 2023-08-18
Return empty object for getContext

### LINKED ISSUE #3398: client-support-angular: Provide a more accurate return type for LuigiContextService.getContext
**Description**

Within the `client-support-angular` library, the [`LuigiContextService.getContext`](https://github.com/SAP/luigi/blob/459e74bff56ba3337c825fdd9069e2f33ea9db1c/client-frameworks-support/client-support-angular/projects/client-support-angular/src/lib/service/luigi-context-service.ts#L11-L14) function specifies its return type as `Context`. However, as its documentation indicates, the returned value can be `null` or `undefined`, which should be reflected in the return type (making it `Context | null | undefined`).

**Reasons**

Doing so would make it less awkward to deal with in typescript.

JohannesDoberer:
Nightly 2.3.1-dev.202308221736

Fixes #3398


ndricimrr (approved):

JohannesDoberer:
Follwo-up ticket: https://github.com/SAP/luigi/issues/3418

PR #3414, opened 2023-08-28
Add getUserSettings for webcomponents

### LINKED ISSUE #3403: Add getUserSettings for WC LuigiClient

JohannesDoberer:
Fixes #3403


core/src/services/web-components.js
```diff
@@ -324,6 +327,26 @@ class WebComponentSvcClass {
       });
     });
   }
+
+  /**
+   * Get stored user settings for a specific user settings group
```
alexandra-simeonova, line 332:
```suggestion
   * Gets the stored user settings for a specific user settings group

```

core/src/App.svelte
```diff
@@ -310,6 +310,8 @@
         };
       },
       set: (obj) => {
+        const wc_container = document.querySelector('.wcContainer');
+        if(wc_container)wc_container.configChangedRequest = true;
```
ndricimrr, line 314:
```suggestion
        if (wc_container) wc_container.configChangedRequest = true;
```
```suggestion
        if(wc_container)wc_container.configChangedRequest = true;
```

core/src/services/web-components.js
```diff
@@ -324,6 +327,26 @@ class WebComponentSvcClass {
...
+   */
+  getUserSettingsForWc(wc) {
+    return new Promise(resolve => {
+      if (wc.userSettingsGroup) {
+        const userSettingsGroupName = wc.userSettingsGroup;
+        LuigiConfig.readUserSettings().then(storedUserSettingsData => {
+          const hasUserSettings =
+            userSettingsGroupName && typeof storedUserSettingsData === 'object' && storedUserSettingsData !== null;
+
+          let userSettings = hasUserSettings ? storedUserSettingsData[userSettingsGroupName] : null;
```
ndricimrr, line 344:
```suggestion
          const userSettings = hasUserSettings ? storedUserSettingsData[userSettingsGroupName] : null;
```
Maybe const since its not being reassigned

core/src/services/web-components.js
```diff
@@ -324,6 +327,26 @@ class WebComponentSvcClass {
...
+          resolve(userSettings);
+        });
+      }
```
ndricimrr, line 347: **if (wc.userSettingsGroup)**  statement above seems to only handle the promise if userSettingsGroup is defined. Should we also maybe handle the other scenario. Maybe this makes it easier for devs to debug their issues in the future. Thought of **reject('userSettingsGroup is not defined');** but maybe also empty rejection would be enough
```suggestion
      } else {
          reject();
      }
```

ndricimrr (approved):
Aside from the comments added looks good. Also tried it out and works as expected.

alexandra-simeonova (approved):

PR #3429, opened 2023-09-05
Fix getContextAsync in client lib for angular

JohannesDoberer:
Since getContext returns an empty object if no context is defined instead of null/undefined we need to check for empty context object in getContextAsync.


JohannesDoberer:
2.5.1-dev.20230950852

ndricimrr (approved):
Tested & working

PR #3322, opened 2023-06-09
Quick Fix Lerna bootstrap

ndricimrr:
**Description**

Changes proposed in this pull request:

- keep a working lerna version for now, until we migrate to newest lerna version

**Related issue(s)**

PR #3449, opened 2023-09-25
Replace luigiContextService & upgrade e2e app angular to 15

JohannesDoberer:


lerna.json
```diff
@@ -1,5 +1,5 @@
 {
-  "lerna": "3.15.0",
+  "lerna": "5.4.0",
```
ndricimrr, line 2: This might produce some issues since latest lerna can crash in the pipeline. We would need to make some kind of proper upgrade to the latest version sometime I guess. 
Latest version that I remember would work was 4.0.0 form [this PR](https://github.com/SAP/luigi/pull/3322/files)

client-frameworks-support/client-support-angular/package.json
```diff
@@ -49,5 +49,8 @@
   },
   "engines": {
     "node": ">=18"
+  },
+  "publishConfig": {
+    "directory": "dist/client-support-angular"
```
ndricimrr, line 54: I assumed this was mostly about the E2E test app, would this change also affect the release process of support-angular package. I am not sure. why was this needed here

JohannesDoberer: Yes, that's right. It's necessary otherwise the wrong package (this [one](https://github.com/SAP/luigi/tree/main/client-frameworks-support/client-support-angular) which is one level up) will be installed for the e2e-test-application.

Codewise it should not have an effect on the release, see [here](https://github.com/SAP/luigi/blob/main/scripts/npmPublish.sh#L147).
In addition I also did a nightly release which was successful.

ndricimrr (approved):

PR #3472, opened 2023-10-13
Support selfRegistered wc for container

### LINKED ISSUE #3463: Container: support for selfRegistered wc
tbd...

JohannesDoberer:
Resolves #3463


container/src/utilities/helpers.ts
```diff
@@ -7,9 +7,23 @@ export class GenericHelpersClass {
    * @param {any} functionToCheck function to check
    * @returns {boolean}
    */
-  isFunction (functionToCheck: any): boolean {
+  isFunction(functionToCheck: any): boolean {
     return functionToCheck && {}.toString.call(functionToCheck) === '[object Function]';
   }
+
+  /**
+   * Checks weather webcomponent is an attribute or property. In case of attribute it returns the parsed value.
```
alexandra-simeonova, line 15:
```suggestion
   * Checks whether web component is an attribute or property. In case of attribute, it returns the parsed value.
```

container/typings/LuigiCompoundContainer.svelte.d.ts
```diff
@@ -33,4 +40,14 @@ export default class LuigiCompoundContainer extends HTMLElement {
    * The anchor value to be passed to the web-component-based micro frontend.
    */
   anchor: string;
+
+  /**
+   * The following properties can be set for the webcomponent object. By default the webcomponent is set to true.
```
alexandra-simeonova, line 45:
```suggestion
   * The following properties can be set for the web component object. By default, the web component is set to true.
```

container/src/utilities/helpers.ts
```diff
@@ -7,9 +7,23 @@ export class GenericHelpersClass {
...
     return functionToCheck && {}.toString.call(functionToCheck) === '[object Function]';
   }
+
+  /**
+   * Checks whether web component is an attribute or property. In case of attribute, it returns the parsed value.
+   * @param webcomponent 
+   * @returns returns the parsed webcomponent value.
+   */
+  checkWebcomponentValue(webcomponent: any): object | boolean {
+    console.log('test');
```
ndricimrr, line 20:
```suggestion
```

container/src/utilities/helpers.ts
```diff
@@ -7,9 +7,23 @@ export class GenericHelpersClass {
...
+    if (typeof webcomponent === 'string') {
+      return JSON.parse(webcomponent);
+    } else if (typeof webcomponent === 'boolean' || typeof webcomponent === 'object') {
```
ndricimrr, line 23: Is boolean and string treated the same in this case. I thought ex.: `<element test="true"></element> `, true is always string even though looks like boolean. Maybe some context I am missing 🤔

container/typings/LuigiContainer.svelte.d.ts
```diff
@@ -6,6 +6,13 @@ export declare interface UserSettings {
   [key: string]: number | string | boolean;
 }
 
+export declare interface WebComponentSettings {
+  id?: any;
+  type?: string;
+  selfRegistered?: boolean;
+  tagName?: string;
```
ndricimrr, line 13: I am not sure about the '?' making all the attributes optional. Are we sure all of these are optional. Maybe some must be required for the minimal example to work or its not the case ?

ndricimrr (commented):
Looks good. Left some comments mostly for clarification

container/src/utilities/helpers.ts
JohannesDoberer: You are right, when you do the html structure it is always a string but you can also do it via javascript like:
```const luigiC = document.createElement('luigi-container');
luigiC.webcomponent={selfRegistered: true}
luigiC.viewurl='bla...'
...and so on
```
In that case it can be an object or boolean. That's why we need this check.

container/typings/LuigiContainer.svelte.d.ts
JohannesDoberer: I have checked it again. No property is required, each one is optional. Also an empty object works.

container/typings/LuigiContainer.svelte.d.ts
```diff
@@ -6,6 +6,13 @@ export declare interface UserSettings {
   [key: string]: number | string | boolean;
 }
 
+export declare interface WebComponentSettings {
```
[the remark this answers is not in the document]
JohannesDoberer: No it's not only for self-registered. It's also when you want to set a tagName by you own, give the webcomponent an id or when your wc is from type module.

ndricimrr (approved):
Looks good

PR #3486, opened 2023-10-26
Btp layout

hardl:


core/src/App.svelte
```diff
@@ -2623,6 +2790,16 @@
         left: calc(#{$leftNavWidthCollapsed} + #{$globalNavWidth});
       }
     }
+    :global(.lui-semiCollapsible:not(.semiCollapsed)) {
```
camelCaseChris, line 2793: Not sure if it still matters at this point - but I just saw this. On our side, we tried to avoid top-level global styles because they are not scoped and we had issues with styling conflicts.

You have to use top-level `:global` here because the class `lui-semiCollapsible` is set in code in this component. If there is a way to change that to a [`class:name` directive](https://svelte.dev/docs/element-directives#class-name) in the template you might be able to change this to something like `.lui-semiCollapsible:global(:not(.semiCollapsed))`

core/src/navigation/LeftNav.svelte
```diff
@@ -496,8 +579,374 @@
...
+      class="fd-navigation fd-navigation--vertical {hideNavComponent
+        ? 'hideNavComponent'
+        : ''} {footerText || semiCollapsibleButton
+        ? 'hasFooter'
+        : ''} {footerText && !semiCollapsibleButton ? 'hasOnlyFooterText' : ''}
+        {isSemiCollapsed
+          ? 'fd-navigation--snapped'
+          : ''}" 
+      role="navigation" 
+      style="width: var(--luigi__left-sidenav--width); height: 100%;">
```
camelCaseChris, line 593: This might be easier to understand:
```suggestion
  <div 
    class="fd-navigation fd-navigation--vertical" 
    class:hideNavComponent={hideNavComponent}
    class:hasFooter={footerText || semiCollapsibleButton}
    class:hasOnlyFooterText={footerText && !semiCollapsibleButton}
    class:fd-navigation--snapped={isSemiCollapsed}
    role="navigation" 
    style="width: var(--luigi__left-sidenav--width); height: 100%;">
```

core/src/App.svelte
```diff
@@ -105,6 +105,10 @@
   let breadcrumbsEnabled;
   let contextRequested = false;
   let loadingIndicatorTimeout;
+  let btpToolLayout = LuigiConfig.getConfigBooleanValue(
+      'settings.btpToolLayout'
+    ) && GenericHelpers.requestExperimentalFeature('btpToolLayout', true);
```
ndricimrr, line 110: let btpToolLayout = LuigiConfig.getConfigBooleanValue(
      'settings.btpToolLayout'
    ) && GenericHelpers.requestExperimentalFeature('btpToolLayout', **true**);
    
    here experimental is set to **TRUE**

core/src/navigation/LeftNav.svelte
```diff
@@ -184,6 +185,10 @@
   let store = getContext('store');
   let getTranslation = getContext('getTranslation');
   let addNavHrefForAnchor = false;
+  let btpToolLayout = LuigiConfig.getConfigBooleanValue(
+      'settings.btpToolLayout'
+    ) && GenericHelpers.requestExperimentalFeature('btpToolLayout', false);
```
ndricimrr, line 190: in App.svelte we check if experimental is **TRUE** , and here if it must be *false**. 
Shouldn't it be also TRUE here too, or am I missing something. 
The Left Nav shows up properly when testing with: 
```
settings.btpToolLayout: true
settings.experimental.btpToolLayout: true
```

And doesn't show at all for the following. 
```
settings.btpToolLayout: true
settings.experimental.btpToolLayout: false
```

So I am wondering if here we should have TRUE-TRUE as well 🤔

core/src/navigation/LeftNav.svelte
```diff
@@ -248,6 +253,61 @@
...
+      });
+      btpNavTopCnt.querySelector('.fd-navigation__list > .fd-navigation__list-item--overflow').style.display = 'none';
+    }
+  }
+
+  calculateNavEntries = () => {
+    const spacer = btpNavTopCnt.querySelector('.fd-navigation__list > .lui-spacer');
+    const moreUL = btpNavTopCnt.querySelector('.lui-moreItems');
+    const entries = btpNavTopCnt.querySelectorAll('.fd-navigation__list > .lui-nav-entry');
+    if(entries.length <= 0) {
```
ndricimrr, line 272: would this also make sense to be less then <= 1
`  if(entries.length <= 1) {`
Because below you have   `for(let i = entries.length -1; i > 0; i--) { ` which is only looping if entries.length >=2

core/src/utilities/helpers/generic-helpers.js
```diff
@@ -257,11 +258,11 @@ class GenericHelpersClass {
   }
 
   getInnerHeight /* istanbul ignore next */() {
-    return LuigiElements.isCustomLuigiContainer() ? LuigiElements.getLuigiContainer().clientHeight : window.innerHeight;
+    return IframeHelpers.getIframeContainer().clientHeight;
   }
 
   getContentAreaHeight /* istanbul ignore next */() {
-    const contentAreaHeight = this.getInnerHeight() - this.getShellbarHeight();
+    const contentAreaHeight = this.getInnerHeight() - IframeHelpers.getIframeContainer().getBoundingClientRect().top;
```
ndricimrr, line 265: I think these changes, might need changing, since this works well when you have a main microfrontend iframe based. But when Having a WebComponent open and opening splitview, it gets distorted:
Screenshot 2024-02-06 at 15 14 13

core/src/navigation/LeftNav.svelte
ndricimrr: unless you meant that for 1 entry present, no looping is done but rather display=flex is still set

core/src/navigation/LeftNav.svelte
hardl: see above, here no warning should be printed on console (one is enough ;) )

JohannesDoberer (approved):
LGTM 👍 !

PR #3494, opened 2023-10-30
Set view group data for wc api

JohannesDoberer:


core/src/services/web-components.js
```diff
@@ -85,6 +88,14 @@ class WebComponentSvcClass {
       },
       getUserSettings: async () => {
         return await this.getUserSettingsForWc(eventBusElement._luigi_node);
+      },
+      setViewGroupData: data => {
+        const vg = NavigationHelpers.findViewGroup(eventBusElement._luigi_node);
```
JohannesDoberer, line 93: Should not be allowed for compound children.

JohannesDoberer:
vg_live_settingsWC
```
export default class extends HTMLElement {
    constructor() {
        super();
        const btn = document.createElement('template');
        btn.innerHTML = `<button id="setViewGroupData">setRandomViewGroup</button>`;

        this._shadowRoot = this.attachShadow({
            mode: 'open',
            delegatesFocus: false
        });
        this._shadowRoot.appendChild(btn.content.cloneNode(true));

        this.$btn = this._shadowRoot.querySelector('#setViewGroupData');
        this.$btn.addEventListener('click', async () => {
            if (this.LuigiClient) {
                this.LuigiClient.setViewGroupData({
                    foo: '(' + Math.round(Math.random() * 100) + ')',
                    bar: '(' + Math.round(Math.random() * 100) + ')'
                });
            }
        });
    }

}
```

Node config
```
{
            pathSegment: 'mfe1Wc',
            label: 'MFE 1 {viewGroupData.foo} Wc',
            viewUrl: '/vg_live_settingsWC.js',
            viewGroup: 'vg1',
            webcomponent: true
          },
```

core/src/utilities/helpers/navigation-helpers.js
```diff
@@ -556,6 +556,39 @@ class NavigationHelpersClass {
...
+        }
+
+        return undefined;
+      } else {
+        return node.viewGroup;
+      }
+    } else if (node.parent) {
+      return this.findViewGroup(node.parent, originalNode || node);
+    }
+  }
```
ndricimrr, line 591: Could you add some JSDoc comments onto these 3 functions . I think since you worked with them here, you can explain them even better.

core/src/services/web-components.js
```diff
@@ -106,6 +106,14 @@ class WebComponentSvcClass {
       },
       getUserSettings: async () => {
         return await this.getUserSettingsForWc(eventBusElement._luigi_node);
+      },
+      setViewGroupData: data => {
```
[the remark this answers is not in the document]
ndricimrr: Ah didn’t see that. I think docu for that part is somehow not indexed when searching for it in the website. Might need to have a look at that later on🤔

ndricimrr (approved):

Issue #3504, opened 2023-11-03
Discrepancy between Installation Docs and Installer Script for Luigi Core VUE


paddomanno:
**Description**

The step-by-step installation instructions provided in the project's documentation for Vue differ significantly from the behavior of the installer script. This inconsistency is causing confusion and potentially leading to installation issues for users trying to set up the project.

I'm not sure about the details of how this affects the installation, but what I noticed is that the webpack config is different and that the script is missing the eslint config. They also use different es5/es6 versions and build commands.

**Expected result**

The step-by-step instructions provided in the documentation should accurately guide users through the installation process, and the installer script should reflect these instructions.

**Actual result**

There are significant differences in the installation process between the documentation and the installer script, which can be confusing and lead to potential issues during installation.

**Steps to reproduce**

1. Visit the [application setup documentation page](https://github.com/SAP/luigi/blob/main/docs/application-setup.md) for installation instructions.
2. Follow the [step-by-step instructions](https://github.com/SAP/luigi/blob/main/docs/application-setup.md#click-to-expand-3) provided for installing the project with Vue.
3. Run the [installer script](https://github.com/SAP/luigi/blob/main/scripts/setup/vue.sh) linked in the docs.
4. Compare the outcome and observed behavior with the documentation.

Diff: [on diffchecker.com](https://www.diffchecker.com/wImI1sEe/)

alexandra-simeonova:
Thank you for opening an issue! The content was updated in pull request #3507. Hence, I will close this issue.

PR #3507, opened 2023-11-06
Remove discrepancy between VUE Installation Docs and Installer Script

alexandra-simeonova:
**Description**

- Remove discrepancy between VUE Installation Docs and Installer Script
- Use updated version of the VUE script as it will be fixed in #3505 

**Related issue(s)**
#3504


ndricimrr (approved):
LGTM

PR #3515, opened 2023-11-06
Add name to root package

camelCaseChris:
**Description**

- When running `npm install` in the root folder (e.g. as instructed in CONTRIBUTING.md), the `name` properties in package-lock.json are changed from `sapluigi` to the name of the local folder of the repository. If the repository was cloned with default settings, that folder name will be `luigi`. The background is that npm will fall back to the folder name if no package name is given in package.json.

  This change sets the `name` property in package.json to `luigi` to avoid this behavior.

**Related issue(s)**

None


ndricimrr (approved):
👍

Issue #2497, opened 2022-01-24
Update Dependencies


JohannesDoberer:
TBD
Most of our dependencies are not up to date.

How do we update all these deps?
- lerna
- angular in e2e-app
...

How can we ensure to be up to date with our deps.?

stanleychh:
[npm-check-updates](https://www.npmjs.com/package/npm-check-updates) might help. Further investigation is needed.

ndricimrr:
Epic seems outdated. Duplicate Epic in #2587 already exists

PR #3518, opened 2023-11-07
[DRAFT] Remove Lerna

### LINKED ISSUE #3321: Lerna bootstrap legacy investigate
Lerna bootstrap seems to be deprecated. Find short/long term solutions 
https://lerna.js.org/docs/legacy-package-management

Further notes: 

### Regarding `npm workspaces`: 
1. Promising option but **not yet stable** enough. Many repos on GitHub seem to be struggling to move to npm workspaces as it seems to bring issues to dependency management for monorepos. (search "migrate to npm workspace" on GH)
2. Packages who moved to `npm workspaces` investigating on alternatives to move away again ...
3. **No built-in option to do one package before another** : has to be done manually [Source](https://earthly.dev/blog/npm-workspaces-monorepo/)
> If you want to make sure package-a is built before package-b, there’s no way to configure this with workspaces, and you have to manually build package-a before building package-b (and this has to be done every time package-a changes)
4. [workspaces is a no-go guide ? ](https://medium.com/@d.ts/how-to-use-npm-workspace-d155076da956) 
5. [Links](https://stackoverflow.com/a/69343375/8250726)

**Conclusion:** 
From the research done seems that there's plenty of experimentation with workspaces at the moment. So might need to wait out the wave of migrations, and once its more stable than maybe reconsider, since we can't afford many migrations 

**Alternatives:** 
**pnpm** - maintained by 1 person ? 
**yarn** - too much effort to migrate from npm to yarn ?

Best current bet seems to make own scripts to install every package's modules as needed. (We only have 9 packages that we release, not 100s.. yet).

camelCaseChris:
**Background**

The goal of this change is to ~~upgrade to [Lerna 7](https://lerna.js.org/docs/introduction)~~ remove Lerna. As of release 7, Lerna no longer manages dependencies. Instead, [it instructs to use the workspace features of existing package managers](https://lerna.js.org/docs/getting-started#adding-lerna-to-an-existing-repo).

That means that the workspaces (what Lerna called packages) of our monorepo and their dependencies are now managed by [npm workspaces](https://docs.npmjs.com/cli/v10/using-npm/workspaces) instead of Lerna.

~~What remains in the hands of Lerna is running scripts across workspaces. npm can do that, too, but [Lerna still offers advanced features](https://lerna.js.org/docs/features/run-tasks). And of course all other [Lerna commands](https://lerna.js.org/docs/api-reference/commands).~~

This leads to a few changes:
- Workspaces are configures in package.json, not lerna.json. Any folder configured there must also have its own package.json in order to be considered a workspace.
- npm commands support the [additional options `--workspace` and `--workspaces`](https://docs.npmjs.com/cli/v10/commands/npm-install#workspace). In short: commands affecting all or selected workspaces can all be executed from the root folder. Also note the section in the documentation linked above that describes the behavior if commands are run without those options.
  - This also means that [npm will ignore .npmrc files in workspace folders](https://github.com/npm/cli/issues/4605). It will issue warnings whenever it does that. That means that .npmrc files in workspace folders can be deleted if they equal the one in the root folder.
- npm creates **one** lockfile for all folders that are configured workspaces. There are no more lockfiles in the workspaces. In fact, any existing ones should be deleted.
  - **Note:** Not every folder containing a package.json is considered a workspace! That is only true if that folder is also included in the `workspaces` property in the root package.json. Folders that are not configured as workspaces can still be their own "npm islands" with isolated package.json, lockfile, .npmrc etc. But any scripts inside package.json of such isolated will not be picked up by `npm run ... --workspaces`.
- npm installs most of the dependencies in the **root** node_modules folder.
  - The workspaces also seem to receive selected dependencies into their local node_modules folders but fewer than before. I did not find the exact rules that define when a dependency is installed in a workspace instead of in the root. If there is documentation, I did not find it. I can guess a few guidelines by looking at the choices npm makes but it does not seem to be super straightforward. In the end, we are not supposed to care; npm is supposed to manage it.
  - The effect out of this that is relevant for us is that some dependencies are now installed in different places. Scripts relying on certain dependency locations may have to be adapted.

**Changes so far:**

- Moved workspace configuration from lerna.json to the root package.json.
- Deleted lerna.json.
- Uninstalled lerna
- Deleted lockfiles in workspaces.
  - There are still lockfiles in some folders, which, however, are not workspaces. **This might need a review, see below.**
- Adapted .gitignore files to cover *.LICENSE.txt files created during the `bundle` script inside the plugins folder.
- Deleted .npmrc files in workspace folders. They all had the same content as the root file.
- Fixing errors in tests and other scripts:
  - Changed path to relevant node_modules in plugins/auth/src/auth-oidc/webpack-extra.config.js. Root-cause: assumed location of packages in a specific node_modules folder was no longer correct.
  - Fix issue with conflicting imported @types/* packages in client-frameworks-support/testing-utilities/tsconfig.json. Likely root-cause: nice more dependencies are now installed in the root node_modules, more installed @types/* packages may now be relevant for TypeScript. The types in conflict were for jasmine and chai.
  - Fixed mounted path to node_modules in test/e2e-js-test-application/package.json. Root-cause: assumed location of packages in a specific node_modules folder was no longer correct.
  - Fixed a Cypress test that failed because a different version of fundamental-styles was applied. The test checked for the color of a button. The button applied a CSS variable but the test checked for the final color value. I changed the test to test that the button had the CSS variable applied for its color instead of the final value.

**Still to be done:**
- Make entire e2e test work.
- Adapt scripts running on GitHub.
- Test if the `release` script still works. The dependency to `lerna` is gone but `lerna-changelog` and its configuration are still there. There is no configuration or documentation that states that `lerna-changelog` requires `lerna` but we have to check and adapt the release process if needed.
- Adapt documentation.
- Test deployment/release due to possibly changed npm configuration.
- Probably more things?

**Tests status**
- Check that `npm run bundle --workspaces --if-present` runs the same scripts as `npx lerna run bundle`.
  - @luigi-project/client-support-ui5:bundle -> :white_check_mark:
  - @luigi-project/testing-utilities:bundle -> :white_check_mark:
  - luigi-client-private:bundle -> :white_check_mark:
  - auth-plugins:bundle -> :white_check_mark:
  - @luigi-project/container-private:bundle -> :white_check_mark:
  - luigi-core-private:bundle -> :white_check_mark:
  - @luigi-project/client-support-angular:bundle -> :white_check_mark:
  - **Notes:**
    1. When we remove folders from the list of workspaces, we have to compare them with this list so that we can catch scripts that may no longer be executed by the above npm command.
    1. npm shows the full console output of the scripts, as opposed to Lerna. Some of our bundle scripts log a high number of warnings, which looks very messy. We may want to clean this up, especially if consumers run those scripts, too.
- Check commit hooks -> :white_check_mark:
- Check unit tests -> :white_check_mark:
- Check e2e tests
  - Run all `bundle` scripts with `npx lerna run bundle`
    - testing-utilities -> :white_check_mark: initial errors fixed
    - plugins -> :white_check_mark: initial errors fixed
    - others -> :white_check_mark:
  - Starting local server in e2e-test-application -> :white_check_mark:
  - Starting local server in e2e-js-test-application -> :white_check_mark:
  - Run `npm run e2e:run:js` in e2e-test-application
    - js-test-app-theming.cy.js -> :white_check_mark: initial errors fixed
    - Others -> :white_check_mark:
  - Run `npm run e2e:run:angular` in e2e-test-application
    - microfrontends.cy.js -> :construction: 1 failing, to be checked
    - user_settings_dialog.cy.js -> :construction: 2 failing, to be checked
    - Others -> :white_check_mark:
  - Run `npm run e2e:run:external` in e2e-test-application
    - external-micro-frontend.cy.js -> :construction: 1 failing, to be checked

**Related issue(s)**
Fixes #3321 

**Outlook**
- This can be a starting point for taking another look at dependency updates, see #2497.
- This also makes it easier to switch to a workflow using `npm ci` rather than `npm install`. To be discussed.


camelCaseChris:
Update: added list of remaining lockfiles and description of additional workspace options for npm commands.

ndricimrr:
extra notes. 
Currently we use lerna for 3 reasons. 
1. lerna bootstrap ... (install package dependencies)
2. lerna bundle (bundle each package in the repo)
3. lerna changelog (Generate Changelog for Release Process)
4. (something else I missed)

Reason **1** is gone since lerna asks us to do it ourselves. This was previously very helpful since it was also helping with linking dependencies with our local environment (i.e.: e2e-test-application/package.json has @luigi-project/client defined but it should instead use client/public distribution when testing client code locally )
Reason **3**   seems like it can be replaced by maybe a one-purpose changelog generation npm packge. ([example 1](https://www.npmjs.com/package/@semantic-release/release-notes-generator) , etc)
Reason 2 could be done manually.Needs further investigation.

ndricimrr:
Handled in #3527

PR #3520, opened 2023-11-08
commented out all ie11 references

### LINKED ISSUE #3511: Remove IE11 logic
We used to have certain extra logic to support ie11 in the past but since ie11 is deprecated we deleted most of our ie11 logic.
Seems like there is still some left throughout the repo that needs to be removed.

VincentUllal:
**Description**

- ...
- ...
- ...

**Related issue(s)**
Fixes #3511


ndricimrr (commented):
Looks good. I think its safe to delete all the commented out parts

PR #3524, opened 2023-11-10
Update width and height on updateModalSettings

JohannesDoberer:


core/src/Modal.svelte
```diff
@@ -256,9 +256,12 @@
       if (e.data.updatedModalSettings.size) {
         settings.size = e.data.updatedModalSettings.size;
         await setModalSize();
+      }else if(e.data.updatedModalSettings.width && e.data.updatedModalSettings.height){
+        settings.height = e.data.updateModalDataInUrl.height;
+        settings.width = e.data.updateModalDataInUrl.width;
       }
       if(LuigiConfig.getConfigBooleanValue('routing.showModalPathInUrl')){
-        Routing.updateModalDataInUrl(RoutingHelpers.getModalPathFromPath(), {'title':settings.title, 'size':settings.size}, e.data.addHistoryEntry);   
+        Routing.updateModalDataInUrl(RoutingHelpers.getModalPathFromPath(), {'title':settings.title, 'size':settings.size, 'height': settings.height, 'width': settings.width}, e.data.addHistoryEntry);
```
ndricimrr, line 264:
```suggestion
        Routing.updateModalDataInUrl(RoutingHelpers.getModalPathFromPath(), {'title': settings.title, 'size': settings.size, 'height': settings.height, 'width': settings.width}, e.data.addHistoryEntry);   
```

core/src/Modal.svelte
```diff
@@ -256,9 +256,12 @@
       if (e.data.updatedModalSettings.size) {
         settings.size = e.data.updatedModalSettings.size;
         await setModalSize();
+      }else if(e.data.updatedModalSettings.width && e.data.updatedModalSettings.height){
+        settings.height = e.data.updateModalDataInUrl.height;
+        settings.width = e.data.updateModalDataInUrl.width;
```
ndricimrr, line 261:
```
settings.height = e.data.updateModalDataInUrl.height;
settings.width = e.data.updateModalDataInUrl.width;
```

This is a little bit confusing to me. 

I can see that **e.data.updatedModalSettings.width** comes from [here](https://github.com/SAP/luigi/blob/main/client/src/linkManager.js#L181)

but where does **e.data.updateModalDataInUrl.height** comes from ? Is it maybe a new property you are introducing that the user sends with the postmessage from client, since updateModalDataInUrl doesn't exist in the repo anywhere else.

ndricimrr: Also since **e.data.updateModalDataInUrl** is not checked if it exists, then doing: 
`settings.height = e.data.updateModalDataInUrl.height;` would fail with, _... trying to access height of undefined_ I think

JohannesDoberer: Thx.. 
c&p mistake

ndricimrr (approved):
👍

PR #3534, opened 2023-11-17
Simplify Plain JS example

ndricimrr:
**Description**

- Too many useless packages used when the example is supposed to be plain. Removed

_setup bash script needs to be tested after merge to main, since downloaded assets will be available when this PR is merged_


VincentUllal (approved):
lgtm

PR #3539, opened 2023-11-22
Prevent multiple eventListener when creating modals

### LINKED ISSUE #3536: Pressing ESC with two overlaying modals leads to error
This issue only occurs in our application. We have tested that it does not always occur. In our app, I open a modal A on top of another modal B. Pressing ESC closes modal A. Reopen the closed modal A. When reopening the modal A, it appears without content.

**Expected result**

The reopened modal should have content inside.

**Actual result**

The reopened modal does not have content.

I aligned about this issue with Johannes Doberer.

JohannesDoberer:
Fixes #3536


core/src/App.svelte
```diff
@@ -1977,9 +1978,15 @@
     configTag = LuigiConfig.getConfigValue('tag');
     isHeaderDisabled = LuigiConfig.getConfigValue('settings.header.disabled');
   });
+
+  const handleKeyDown = (event) => {
+    if (event.keyCode === KEYCODE_ESC && mfModalList && mfModalList.length>0) {
+      closeModal(mfModalList.length-1);
+    }
```
ndricimrr, line 1985:
```suggestion
    if (event.keyCode === KEYCODE_ESC && mfModalList && mfModalList.length > 0) {
      closeModal(mfModalList.length - 1);
    }
```

ndricimrr (approved):
small comment, otherwise looks good

PR #3542, opened 2023-11-27
Create Extra_Resources.md

systemdesign42:
Add an article on Micro Frontends

### DIFF
--- a/docs/Extra_Resources.md
+++ b/docs/Extra_Resources.md
@@ -1,0 +1,1 @@
+- [Everything You Need to Know About Micro Frontends](https://newsletter.systemdesign.one/p/micro-frontends)


alexandra-simeonova:
Hi @systemdesign42,
There is no specific document where we collect extra resources. However, we have a [Twitter](https://twitter.com/luigiprojectio) page where I can post a link to the article you shared. Would that be ok?

alexandra-simeonova:
Thank you, the post can be found here: https://twitter.com/luigiprojectio/status/1730179486245630140

We are also discussing where we could add a dedicated "Further reading" or "Extra resources" page. 🙂 This can be tracked via the following issue: https://github.com/SAP/luigi/issues/3545

PR #3544, opened 2023-11-29
Fix addNavHrefs for CtxSwitcher and TopNavDropDown

JohannesDoberer:
addNavHrefs does not work for ContextSwitcher and TopNavDropDown


ndricimrr (approved):
Tested & Works well

PR #3550, opened 2023-12-01
npm run lint only when changes in container

### LINKED ISSUE #3531: Container ESlint commit hook takes too long
```
> @luigi-project/container-private@1.0.0 lint
> eslint src/**/*.*
```

Lines above suggest that we are running eslint on container code each time we commit. This is nice, but if you are not working on container code it is not nice :/ since it takes too long to just do one commit. 

Could be that the delay is caused by something else. 

Investigate if that is causing the delay and use some kind of diff check to only run container lint if /container folder has any changes. 
It would be nice to do this without installing any new packages. 

Hint: This line could work but not tested properly for all edge cases: 
`git diff --exit-code --name-only | grep '^container/' && npm run lint --prefix=container || true`

VincentUllal:
**Description**

- ...
- ...
- ...

**Related issue(s)**

Fixes #3531

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
    },
    "husky": {
       "hooks": {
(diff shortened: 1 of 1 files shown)


ndricimrr (approved):
works well 👍

PR #3564, opened 2023-12-05
Fix resizing when drawer opens/closes

### LINKED ISSUE #3150: Splitview Mf won't be resized if drawer overlap is false
Given:
splitview + drawer *overlap=false* does not resize the splitview. 
Only the "main" microfrontend will be resized when drawer will be opened.

Expected:
Both microfrontends (main and splitview) will be resized.

camelCaseChris:
**Description**

Not sure if my approach makes sense; please comment.

I rewrote `resizeMicrofrontendIframe` so that:
- It does not only resize the content IFrame but also some elements for the split view and the horizontal navigation. (The breadcrumbs are also mentioned in #3150 but as far as I could see, they are above the drawer and should therefore not be resized?)
- The resizing is done as far up in the DOM as possible. So e.g. not on `<iframe>` elements but on their wrapper elements.
- It now also supports a flag for resetting all the widths, which is now used when the drawer is closed again.
- It only influences the `width` styles, so that it does not conflict with other style properties that are set e.g. for the split view elements.

The name `resizeMicrofrontendIframe` is probably not fitting any longer but I did not want to change it before validating the overall approach.

**Related issue(s)**

Fixes #3150.


JohannesDoberer:
Looks good,  but when I tested it, I found some issues.
If you open the drawer first and then the splitview a wrong width is calculated. Additionally a calculation is done when you open the splitview and/or collapse (drawer must be active).

Regarding the breadcrumbs:
I think a follow up issue would be necessary for the breadcrumbs, because if breadcrumbs are active, in my opinion the drawer should go up to the TopNav and not be below the breadcrumbs and depending on whether the drawer should overlap or not, the width of the breadcrumbs should be changed
But that's for @hardl to decide.

camelCaseChris:
> Looks good, but when I tested it, I found some issues. If you open the drawer first and then the splitview a wrong width is calculated. Additionally a calculation is done when you open the splitview and/or collapse (drawer must be active).

Good catch, should be fixed now. I clicked through several sequences of steps, did not find any other issues. I also tried to reduce unnecessary operations each time the function is called, which happens a lot. There is still some overhead but since the number of containers to resize is not always the same, I need to check each of them individually.

> Regarding the breadcrumbs: I think a follow up issue would be necessary for the breadcrumbs, because if breadcrumbs are active, in my opinion the drawer should go up to the TopNav and not be below the breadcrumbs and depending on whether the drawer should overlap or not, the width of the breadcrumbs should be changed But that's for @hardl to decide.

True. I was also wondering if the whole page might even be done as a grid layout, or stacked flex layout depending on intended browser support. That might make the whole resizing easier. But I do not know the history so this might be a bad idea. It would also be a rather big rewrite.

PR #3571, opened 2023-12-07
Fix splitview closing

### LINKED ISSUE #3361: Splitview will be removed if model will be closed
How to reproduce:
- Go to fiddle
- `LuigiClient.linkManager().openAsSplitView('/home/overview')`
- `LuigiClient.linkManager().openAsModal('/home/overview')`
- close the model

Result:
- splitview will be removed as well

Expected:
- splitview should stay

camelCaseChris:
**Description**

I could not reproduce the bug, behavior on my side was as expected: closing the modal did not close the split view. Maybe the issue was fixed by some other change in the meantime.

In any case, I added a test to make sure that this does not happen again in the future.

**Related issue(s)**

Fixes #3361.


test/e2e-test-application/cypress/e2e/tests/1-angular/navigation-4.cy.js
```diff
@@ -290,4 +290,33 @@ describe('Navigation', () => {
...
+
+      // Modal is open
+      cy.get('div[data-testid="modal-mf"]').should('exist');
+
+      // Close modal
+      cy.get('button[data-testid="lui-modal-index-0"]').click();
+
+      cy.get('#splitViewContainer').should('exist');
+      cy.get('#splitViewDragger').should('exist');
+      cy.get('div[data-testid="modal-mf"]').should('not.exist');
```
ndricimrr, line 319: This works, but it could generate flaky tests. 

It would be better if you put this section inside of the **then** callback above to avoid a race condition when should('exist') is called before the modal/splitview is open

ndricimrr (commented):
Looks good. Small comment

test/e2e-test-application/cypress/e2e/tests/1-angular/navigation-4.cy.js
camelCaseChris: Good catch, thanks!

ndricimrr (approved):
Looks good

PR #3572, opened 2023-12-07
calc max-height for tabNav dropdown

### LINKED ISSUE #3388: max-height and scroll for tabnav dropdown
the following works in a particular scenario, needs to be checked against  other view setups (e.g. no shellbar, with tabnavHeader,...)
```
.luigi-tabsContainer .fd-popover__body {
    max-height: calc(100vh - calc(var(--luigi__shellbar--height) + var(--luigi__horizontal-nav--live-height, var(--luigi__horizontal-nav--height))));
    overflow-y: auto;
}
```

JohannesDoberer:
fixes #3388

### DIFF
--- a/core/src/navigation/TabNav.svelte
+++ b/core/src/navigation/TabNav.svelte
@@ -1,11 +1,22 @@
 <script>
-  import { afterUpdate, beforeUpdate, createEventDispatcher, onMount, getContext, onDestroy } from 'svelte';
+  import {
+    afterUpdate,
+    beforeUpdate,
+    createEventDispatcher,
+    onMount,
+    getContext,
+    onDestroy,
+  } from 'svelte';
   import { Navigation } from './services/navigation';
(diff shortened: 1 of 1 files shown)


test/e2e-test-application/cypress/e2e/tests/1-angular/globalSearch.cy.js
```diff
@@ -58,7 +58,7 @@ describe('Global Search', () => {
   });
 
   describe('Get results', () => {
-    it('Type something and get results', () => {
+    it.only('Type something and get results', () => {
```
camelCaseChris, line 61: Remove the `.only`?

camelCaseChris (commented):
Tried it out locally, all worked as I would expect. Tried with/without breadcrumbs, with different tabNav settings, with additional drop-downs in the tabNav and also in mobile simulation. Scrolling worked.

Just one question in the code...

camelCaseChris (approved):

PR #3581, opened 2023-12-11
3574 migrate docs to svelte4

### LINKED ISSUE #3574: Migrate docs to Svelte 4

camelCaseChris:
**Description**

- Upgrade from Svelte 3 to Svelte 4.
- Bump of other dependencies that are directly or indirectly related to Svelte(Kit).
- Changes to tsconfig to remove warnings and errors. Included upgrade from TS 4 to TS 5.

Comments/concerns:
- I cleaned up unused SvelteKit adapters, including the one for Netlify. This works locally but we should check if the deployment of the site is somehow impacted by this.
- At one point it was not possible to update (or even delete) dependencies one by one because of looping conflicts in peer dependencies. The only way I could fix that was by deleting package-lock.json and node_modules, bulk-change package.json and then do `npm install` again. If you run into any issues when doing a regular `npm install`, doing `npm ci` might be needed.
- I kept getting several kinds or warnings and errors around tsconfig.json. After several changes it now looks a lot different. With the current state, `npm run check` works fine as well as the `build`, `core` and `client` scripts. tsconfig.json uses the generated config file of SvelteKit again. With the bumps to SvelteKit and TS, running `npm client` once will make some updates to that generated config file.

Not included:
- Bump of other dependencies.

**Related issue(s)**

Fixes #3574.


website/docs/package.json
```diff
@@ -17,7 +17,7 @@
 		"lint": "prettier --plugin-search-dir . --check .",
 		"format": "prettier --plugin-search-dir . --write .",
```
ndricimrr, line 18: From what I can see locally, we make no use of **lint** or **format**. Prettifying is done on the commit hook anyway. So could you remove these two lint and format scripts as well as the prettier dependency down below, so we don't have to maintain its versions in the future

camelCaseChris: Done

ndricimrr (approved):
Tested and working as expected.

camelCaseChris:
Test deployment seems to have worked fine. Link to the test deployment is available in the list of checks.

ndricimrr (approved):
Deployed Site looks good too 👍

PR #3585, opened 2023-12-12
Refactor check webcomponent function

### LINKED ISSUE #3566: Correct webcomponent function return value
**Goal**
Function should return either:  **object | boolean** , no other values

**Description**
https://github.com/SAP/luigi/blob/3e3d636fff1bf65abb054edff2ffe3b0f6b15bde/container/src/utilities/helpers.ts#L27-L40

Focus on line **29**

In **checkWebcomponentValue()** function in **container/src/utilities/helpers.ts** we parse a string value and expect it to be JSON object. if it is an object we handle it as is and return **parsedValue** as an object. 

However this could fail in two ways:
1. User misconfigures webcomponent config and JSON.parse produces error for non-parseable String: `Uncaught SyntaxError: Unexpected token ... is not valid JSON`.  
      - In this case the Error should be caught and proper warning be given to the user in the console.
2. JSON.parse('123456') or JSON.parse('"Any-String"') would return a String '123456' or "Any-String"
      - In such cases when parse result is not an object the default return value should be a boolean **true**. (It could be argued later that a console information warning here would be beneficial or not)

Other TODOs:
- Adjust the unit tests for **checkWebcomponentValue** function 
- **webcomponent: any** => could be typed to **webcomponent: object | boolean | string**
- Update JSON doc of function parameter to explain what the parameters could be. 
- If possible also add a Cypress test for this scenario

JohannesDoberer:
Fixes #3566
In this PR:
No value validation for properties, like `webcomponent={selfregisteredWrongValue=true}` etc.
Strings will be JSON parsed and if there is any parsing error, the app will throw an error and wc will not be shown.
Function will only return boolean or object.
No standard/default return type.

Additional fix: context can be set as property and no validation as well. If there is a parse error wc will not be shown.


container/src/utilities/helpers.ts
```diff
@@ -21,22 +21,32 @@ export class GenericHelpersClass {
...
+  /**
+  * Resolves the context to an object. If the context is a string, it attempts to parse
+  * it as JSON. If parsing fails JSON parse error will be thrown.
+  * @param {object | string} context - The context to be resolved.
+  * @returns {object} The resolved context as an object.
+  */
+  resolveContext(context: object | string): object {
+    return context
+      ? typeof context === 'string'
+        ? JSON.parse(context)
```
ndricimrr, line 46: It certainly doesn't hurt to have **Json.parse** fail as we can catch this in our unit tests easily as a native JSON thrown error. 

However I think this pollutes the Console with seemingly **unhandled error messages**, making it look like the error is on our side. 
It would look better if we actually handle the thrown error with a **try/catch** statement and a default warning and default **{} return value** on failure instead of letting **JSON.parse** throw whatever error wants.

container/src/utilities/helpers.ts
```diff
@@ -21,22 +21,32 @@ export class GenericHelpersClass {
...
+   * @returns {object | boolean} webcomponent returns the parsed webcomponent value.
    */
-  checkWebcomponentValue(webcomponent: any): object | boolean {
+  checkWebcomponentValue(webcomponent: object | boolean | string): object | boolean {
     if (typeof webcomponent === 'string') {
-      let parsedValue = JSON.parse(webcomponent);
-      if (this.isObject(parsedValue)) {
-        if ('selfRegistered' in parsedValue) {
-          parsedValue.selfRegistered = parsedValue.selfRegistered === 'true';
-        }
```
ndricimrr, line 33: I don't know much about this **selfRegistered** property check. Is it not needed anymore that it was deleted here

container/src/utilities/helpers.ts
```diff
@@ -21,22 +21,32 @@ export class GenericHelpersClass {
...
-      }
-      return parsedValue;
+      return JSON.parse(webcomponent);
```
ndricimrr, line 29: I would add the same comment here about handling it same as in **resolveContext** with a try/catch

ndricimrr (commented):
Looks good. Just some small comments.

Think it should be standard to **try/catch** the JSON parse thrown errors although it might look ugly at first.

The [pipeline seems to be failing](https://github.com/SAP/luigi/actions/runs/7209274676/job/19640014602) for some reason, and I think Cypress does not like it very much when we don't try/catch errors as expected/ 

```
 SyntaxError: The following error originated from your application code, not from Cypress. It was caused by an unhandled promise rejection.
 
 https://on.cypress.io/uncaught-exception-from-application
      at JSON.parse (<anonymous>)
```

container/src/utilities/helpers.ts
JohannesDoberer: The reason why I want to avoid the try/catch is that the luigiContainer or luigiCompoundContainer should fail if there is an error. When we do try/catch the wc will shown although there is a wrong context value. This can lead to unforeseen behaviour, wc is there but with a wrong context. In addition besides the property/attribute _webcomponent_ all the others will not validated as well.
For example if you have somthing like this:
```
<luigi-container id="lc" data-test-id="luigi-container" client-permissions="test"></luigi-container>
```
you will get an invalid json as well.

container/src/utilities/helpers.ts
JohannesDoberer: no validation of any value

container/src/utilities/helpers.ts
JohannesDoberer: see _context_

container/src/utilities/helpers.ts
ndricimrr: I see 🤔 . In that case we could also do a custom throw Error to have the same intended fail behaviour but make it easier for devs to find where they did wrong. Something like:
```
try {

} catch(Error e) {
 console.error("Failed because of X specific reason");
 throw Error(e);
}
```
But I guess we'd have to discuss it first if its too ugly or too much detailed

JohannesDoberer: In that case we need to implement it for every property/attribute, because we can't do it for context and webcomponent only.

Let's discuss it in the daily.

ndricimrr (approved):
Looks good

ndricimrr:
One additional thing:
Do we have to change something on the [typings file](https://github.com/SAP/luigi/blob/main/container/typings/LuigiContainer.svelte.d.ts#L31-L38) to keep it in sync with the inputs of the checkWC.. function or not needed.

ndricimrr (approved):

PR #3590, opened 2023-12-14
1603 lazy loading option for compound web component mode

### LINKED ISSUE #1603: LazyLoading for web components MFEs
In web component compound views, load MFEs only if they are within current visible viewport.

camelCaseChris:
**Description**

- There is an option to activate lazy loading mode for a compound web component.
- With lazy loading, after rendering the container for each compound item, the item is not immediately rendered. Instead, the container is added to an `IntersectionObserver` that triggers the rendering when the container is visible. To support this, each compound item container is rendered with a temporary height, so that not all containers are immediately visible in the viewport.
  - The temporary height is removed from the container after attaching the web component to the compound item.
  - The container might in some cases switch back to a height to 0, e.g. if the attached web component renders its contents asynchronously. But this is nothing we can solve but needs to be documented and is then up to the UI developers to (if even needed) adapt to through configuration or changes to the web components.

Discussion points:
- Configuration? Current proposal:
  - [x] `compound.lazyLoadingOptions` (object) options for lazy loading of the compound.
    - [x] `enabled` (boolean) to activate lazy loading mode.
    - [x] `temporaryContainerHeight` (CSS length string) height to apply as a default for all children of the compound. If not specified, a default fallback is used, currently "500px".
    - [x] `intersectionRootMargin` (CSS length string), to be passed on to the IntersectionObserver
  - [x] `compound.children[].layoutConfig.temporaryContainerHeight` (string) CSS height to apply to the individual children of the compound. Overrides the compound default.
- Lazy loading for nested compounds?
  - Yes
  - [x] To be checked: does IntersectionObserver work for things inside shadow DOM?
- Lazy loading for custom renderers?
  - Setting the temporary height to the compound item containers may conflict with what custom renderers are doing. How to handle this?
  - For now: "solve by documentation + configuration".
    - Documentation: describe what lazy load is doing.
    - [x] Configuration: offer an option `compound.lazyLoadingOptions.noTemporaryContainerHeight` to suppress the setting/removing of the temporary height to the compound item containers.
- Central place to store constants, like the default fallback for `compound.temporaryContainerHeight`?
  - In class
- Standards for error handling?
  - Foreseen, developer errors: log to console
  - Relevant for end-users: internal alert mechanism
- [x] Clean up console logging and comments

Known issues:
- [x] Even with big temporary heights for the compound item containers, it can happen that surprisingly many items are rendered right away. This needs further investigation.

**Related issue(s)**
Fixes #1603.


core/src/services/web-components.js
```diff
@@ -309,22 +430,57 @@ class WebComponentSvcClass {
...
+  /**
+   * Responsible for rendering web component compounds based on a renderer or a nesting
+   * micro frontend.
+   *
+   * @param {*} navNode the navigation node defining the compound
+   * @param {*} wc_container the web component container dom element
+   * @param {*} context the luigi node context
+   */
+  renderWebComponentCompound(navNode, wc_container, extendedContext) {
+    const useLazyLoading = navNode.compound?.lazyLoadingOptions?.enabled === true;
```
JohannesDoberer, line 461: Can you please document this settings object `lazyLoadingOptions`.
E.g [here](https://github.com/SAP/luigi/blob/main/docs/navigation-parameters-reference.md#compound) and maybe it is worth to mention this object in this [document](https://github.com/SAP/luigi/blob/main/docs/web-component.md#compound-web-components) as well.

core/src/services/web-components.js
```diff
@@ -286,20 +345,82 @@ class WebComponentSvcClass {
...
+      } else {
+        console.error('Could not find WC container data', {
+          for: compoundItemContainer
+        });
+      }
+      observer.unobserve(compoundItemContainer);
+    });
+  }
+
+  /**
```
JohannesDoberer, line 377: Please add a description to the function what it does.

core/src/services/web-components.js
```diff
@@ -286,20 +345,82 @@ class WebComponentSvcClass {
...
+
+    const temporaryContainerHeight =
+      compoundItemSettings.layoutConfig?.temporaryContainerHeight ||
+      compoundSettings.lazyLoadingOptions?.temporaryContainerHeight ||
+      DEFAULT_TEMPORARY_HEIGHT;
+
+    compoundItemContainer.style.height = temporaryContainerHeight;
+  }
+
+  /**
```
JohannesDoberer, line 400: Please add a description to the function what it does.

JohannesDoberer (commented):
It looks good to me!
Good Job 👍 .
Small comments regarding documentation.

core/src/services/web-components.js
camelCaseChris: Done...

core/src/services/web-components.js
camelCaseChris: Done...

core/src/services/web-components.js
camelCaseChris: Done...

JohannesDoberer (approved):
LGTM 👍 !

PR #3593, opened 2023-12-15
refactor todos and removed outdated todos

### LINKED ISSUE #3532: Resolve TODOs in repo
Search for `// TODO` or `//TODO` in the repo and investigate which TODOs can be resolved starting from easiest->hardest.
Can be split into separate tickets.

VincentUllal:
**Description**

Draft PR

**Related issue(s)**

Fixes #3532

### DIFF
--- a/core/src/App.svelte
+++ b/core/src/App.svelte
@@ -280,7 +280,6 @@
     });
   };
 
-  //TODO refactor
   const getComponentWrapper = () => {
     return {
       get: () => {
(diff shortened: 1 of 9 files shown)


core/src/App.svelte
```diff
@@ -366,14 +333,8 @@
...
-            } else if (prop === 'tabNav') {
-              tabNav = obj.tabNav;
-            } else if (prop === 'isNavigateBack') {
-              isNavigateBack = obj.isNavigateBack;
-            } else if (prop === 'goBackContext') {
-              goBackContext = obj.goBackContext;
-            } else if (prop === 'isNavigationSyncEnabled') {
-              isNavigationSyncEnabled = obj.isNavigationSyncEnabled;
+            } else {
+              eval(prop + " = " + JSON.stringify(obj[prop]));
```
ndricimrr, line 337: This is a very interesting solution. Definitely cleaner. However, we want refrain from using the **eval** function altogether as it introduces issues with Content Security Policies. To allow this properly we need to enable [**'unsafe-eval'**](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy#unsafe-eval)

We had major issues with this before in the past trying to remove all direct and indirect usage of **eval** https://github.com/SAP/luigi/issues/1458

ndricimrr (commented):
Only one comment. Rest looks okay. Will check again. Maybe remove the eval solution and look for another TODO instead if it can't be done without eval

PR #3595, opened 2023-12-18
Update ts container typings

### LINKED ISSUE #3592: Update Container Typescript Typings
Container TS Typings seem to be out of sync with the current implementation.

Check the implementation and make sure that whatever function/property is exposed should be reflected on the typings TS Files. 

Also check if existing typings are in sync with the current implementation or something changed.

This issue could help with #3591

ndricimrr:
**Description**

- Add all typings
- Refactor some JSDoc

**Related issue(s)**

Resolves #3592


JohannesDoberer (approved):
LGTM

PR #3602, opened 2024-01-09
Add Container API to docu

### LINKED ISSUE #3591: Extend Container API Documentation
Currently we document Core/Client API functions at length in our docu website. 

Need to have something similar for Container API. Maybe through the same process or different if not possible.

_Starting Point would be the Typescript Typings:_
[LuigiContainer.svelte.d.ts](https://github.com/SAP/luigi/blob/main/container/typings/LuigiContainer.svelte.d.ts)
[LuigiCompoundContainer.svelte.d.ts](https://github.com/SAP/luigi/blob/main/container/typings/LuigiCompoundContainer.svelte.d.ts)

ndricimrr:
**Description**

- Utilise automatic docu generation from code

local run tips:
```
// Generate .md content that goes to /docs folder
cd scripts
npm run docu

// generate new docu.json html-ified content
cd website/docs
npm run build
npm run core

cd website/docs
npm run preview
```

**Related issue(s)**

Resolves #3591


container/typings/LuigiCompoundContainer.svelte.d.ts
```diff
@@ -1,7 +1,3 @@
-export declare interface UserSettings {
-  [key: string]: number | string | boolean;
-}
-
```
ndricimrr, line 4: removed since it doesn't seem to be used anywhere else but here, and it can be represented by an 'Object' too as well. Same for other interfaces in ...Container.d.ts

container/typings/LuigiCompoundContainer.svelte.d.ts
```diff
@@ -11,12 +7,17 @@ export declare interface WebComponentSettings {
...
+   */
+  viewurl: string;
+
   /**
    * The configuration for the compound microfrontend
    * Take a look at https://docs.luigi-project.io/docs/navigation-parameters-reference/?section=compound
    * 
    */
-  compoundConfig: any;
+  compoundConfig: Object;
```
camelCaseChris, line 20: According to the TS docs, this might not be what we want:
https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#number-string-boolean-symbol-and-object

`object` seems to be the recommendation, meaning "anything that is not a primitive".

Same could be valid for the other uses of `Object` in this file.

ndricimrr: Yes I think I know what you refer to, but I believe in this context code is not directly depending on TS typings. Main aim of this file is to help with Documentation Generation. Changing Object -> object only changes the O from Capital to lowercase: 

Screenshot 2024-01-12 at 11 10 35

The context as to what the **object** is we leave to the developer, as we don’t want to enforce such a strict opinionated typescript typings for docu. (for which devs might even have their other own opinions)

Also all of our docu is built upon **Object** notation and that has never caused a misunderstanding. 
(ex.: https://docs.luigi-project.io/docs/luigi-core-api/?section=parameters ). Changing Object-object here would cause a asymmetry with rest of docu. 

The context in which Object->object change would be relevant would be if we need to have super strict opinionated  typings enforced for some functionality reason, for which we would probably also change code or flood the console with warnings (Given object does not meet requirements), which we currently are against doing.

Also, as a historical background, this `.d.ts` file here is primarily just an extra (duplicate) file that’s used for generating documentation and Code Editor suggestions for Typescript users of container package at most. We created and curate these **.d.ts** files manually since generating TS type declarations from our Svelte based container code was too complicated and not possible at the time.

+, it doesn’t affect the functionality whatsoever, as that is left to Svelte code on LuigiContainer.svelte. So Object vs object here would produce no runtime/build errors. In even more depth, the type of these variables boils down to a Javascript Object notation ( check [here](https://github.com/SAP/luigi/blob/main/container/src/LuigiCompoundContainer.svelte#L9)  , so below Typescript territory

JohannesDoberer (approved):
LGTM 👍 . Maybe also the `since x.x.x` would make sense but we can also handle it in another issue.

Issue #3513, opened 2023-11-06
Investigate flaky test wait call removal


ndricimrr:
Adding waits to certain cypress tests seems to do the job at times, but the problems reoccurs some time after. 
It's recommended to steer away from the cy.wait and similar functions. 
Investigate a way to make the test run without the wait calls. 
(low prio for now)

camelCaseChris:
This is not done yet. One more change to go.

PR #3603, opened 2024-01-10
Experiment: Reduce waits and avoidable API calls in Cypress

camelCaseChris:
**Description**

Using `luigi-client-link-manager-features-4.cy.js` as a pilot, this change experiments with reducing `cy.wait()` and other avoidable API calls in your Cypress tests **while not making tests flaky**.

Examples for avoidable API calls:
- `.should('exist')` before an action.
- `cy.window()` without using the window.
- Complex access to shadow DOM inside `.then` callbacks. I saw them cause problems locally in some cases because of `then` not being repeatable. Cypress has the `shadow()` function that seems to work just fine.

We should start the test suite for this PR a couple of times to get a feeling for the outcome.

**Related issue(s)**
Part of #3513


camelCaseChris:
I covered two files by now and did maybe 7 or 8 test runs with this PR. So far so good. I take the PR out of draft to get reviews on the actual changes. Depending on further test runs we can decide to keep going with or stop this.

test/e2e-test-application/cypress/e2e/tests/1-angular/luigi-client-link-manager-features-4.cy.js
```diff
@@ -1,16 +1,13 @@
...
   describe('Drawer', () => {
     let $iframeBody;
+
     beforeEach(() => {
       cy.visitLoggedIn('/projects/pr2');
       cy.getIframeBody().then(result => {
         $iframeBody = result;
         cy.goToLinkManagerMethods($iframeBody);
-        cy.expectPathToBe('/projects/pr2');
+        // cy.expectPathToBe('/projects/pr2');
```
ndricimrr, line 10: was there a benefit to commenting this out or did it seem to be unnecessary anymore to check path 
If we don't need this at all here, and its better without, maybe remove it altogether. 🤔

camelCaseChris: Very true, thanks. The reason for removing it was that `goToLinkManagerMethods` already does that assertion so there is no reason to do it again right after.

test/e2e-test-application/cypress/e2e/tests/1-angular/luigi-client-link-manager-features-4.cy.js
```diff
@@ -227,110 +192,56 @@ describe('Luigi Client linkManager Webcomponent, Drawer', () => {
...
-        cy.wait(700);
-        cy.get('.wcContainer>div>div>*').then(container => {
-          cy.wait(500);
-          const root = container.children().prevObject[0].shadowRoot;
-          const wcContent = root.querySelector('button').innerText;
-
-          expect(wcContent).to.equal('Start');
-        });
-      });
+      cy.get('@btn').should('have.text', 'Start');
```
ndricimrr, line 216: Just as a question. What is this `@btn` notation. I haven't seen it before 
Also below `@timer ` below 🤔 . Seems to be quite a shorter solution indeed.

camelCaseChris: Cypress calls those aliases. When used with HTML elements, this lets you save and replay queries:
https://docs.cypress.io/guides/core-concepts/variables-and-aliases#Elements

Useful stuff and also makes it easier to use `.should()` instead of `.then()`. At least in a few cases.

The aliases used here are created in the `beforeEach` hook.

ndricimrr (approved):

PR #3606, opened 2024-01-15
Update ui5-support-lib.js

felix960799:
Adds the possibility to add the boolean data.preventBrowserHistory option to the manifest If set to true, navigating inside the same route will prevent the browser history

We use routes that build up on selecting different options. If we click back on the browser arrow, we need to click multiple times until we reach the inital start point e.g. Dashboard again

if the flag is not set, there are no side effects to current behaviours

**Description**

- ...
- ...
- ...

**Related issue(s)**


JohannesDoberer (approved):
LGTM 👍 !
Thx for the contribution!

PR #3607, opened 2024-01-15
Handle Svelte Compiler Warnings

### LINKED ISSUE #3563: Improve Core Code Quality from Svelte Warnings
After introducing Vite we get **nice** warnings about new svelte code recomendations when running `npm run build` in `core/`

We might as well solve these issues to avoid having these warnings there instead of just enforcing a non-verbose vite build.

Looking at the warnings a lot of them are very small suggestions, that shouldn't do any harm. So we can start from the easier ones. 

Following is generated summary that is more legible and helps to separate these issue into mini-tasks:

**Note:** Lots of _good first issues_ can come out of this. 

In issues for which the warnings make no sense for our use case, need to develop a solution, simple enough to disable the warnings if it just makes no sense. ex: "Unused export property" issues might be one of them.

### Accessibility Issues (72 Issues):

**Issue Type**                                     | **File**                                          | **Occurrences(s)**      | **Error/Warning Message**
-----------------------------------------------------|---------------------------------------------------|------------------|-----------------------------
A11y: Missing href attribute for <a> element        | TopNav.svelte, TabNav.svelte, LeftNav.svelte, SplitView.svelte, Authorization.svelte, AuthorizationSimpleProfileMenu.svelte, GlobalSearch.svelte                    | 13          | A11y: `<a>` element should have an href attribute
A11y: Missing keyboard event for clickable element  | TabNav.svelte, LeftNav.svelte, TopNav.svelte, GlobalNav.svelte, SplitView.svelte, etc                    | 37 | A11y: visible, non-interactive elements with an on:click event must be accompanied by an on:keydown, on:keyup, or on:keypress event.
A11y: Noninteractive element with nonnegative tabIndex | TopNav.svelte, etc                  | 13 | A11y: Noninteractive element cannot have nonnegative tabIndex value
A11y: Redundant role 'list'                          |  Authorization.svelte , UserSettingsDialog.svelte                        | 2          | A11y: Redundant role 'list'
A11y: Avoid using autofocus                         | GlobalSearch.svelte              |  2  | A11y: Avoid using autofocus
$: has no effect outside of the top-level            |  App.svelte              | 5         | $: has no effect outside of the top-level

### Unused CSS Selectors (17 Occurrences):

**File**                                           | **Occurrence(s)**      | **Unused Selector**
----------------------------------------------------|------------------|------------------------
App.svelte                   |           5   | -
ContextSwitcher.svelte                |   3    | -
BadgeCounter.svelte                 |     4  | -
UserSettingsEditor.svelte  |  1 | -
LeftNav.svelte | 3 | -
TopNav.svelte | 1  | -

### Unused exports (7 Issues):

 **File**                                          | **Occurence(s)**      | **Error/Warning Message**
---------------------------------------------------|------------------|-----------------------------
 TopNav.svelte                      | 1            | TopNav has unused export property 'urlAuthError'. If it is for external reference only, please consider using `export const urlAuthError`
 UserSettingsEditor.svelte                     |  1          | UserSettingsEditor has unused export property 'userSettingsGroupKey'. If it is for external reference only, please consider using `export const userSettingsGroupKey`
 UserSettingsDialog.svelte                     |  1          |  UserSettingsDialog has unused export property 'schemaObj'. If it is for external reference only, please consider using `export const schemaObj`
GlobalSearch.svelte                |   1       | GlobalSearch has unused export property 'displayCustomSearchResult'. If it is for external reference only, please consider using `export const displayCustomSearchResult`
 GlobalSearchCentered.svelte   |  1 | GlobalSearchCentered has unused export property 'displayCustomSearchResult'. If it is for external reference only, please consider using `export const displayCustomSearchResult`
 Breadcrumb.svelte  |  2 | Breadcrumb has unused export property 'hideNavComponent'. If it is for external reference only, please consider using `export const hideNavComponent` +  Breadcrumb has unused export property 'responsiveNavSetting'. If it is for external reference only, please consider using `export const responsiveNavSetting`

Total Issues: 16

camelCaseChris:
**Description**

- The existing SCSS settings in the Svelte configuration caused the plug-in in VS Code to throw syntax errors. That kept it from showing the actual Svelte warnings. I adapted the SCSS settings so that they also work with the VS Code plug-in.
- Ignored all warnings about key events missing for non-interactive elements with click events. Implementing those key events would be a larger change to the overall keyboard usability of the core.
- After checking, removed all cases of unused component props.
- Ignored all warnings about `tabindex` on non-interactive elements. Fixing them would require changing the markup, which is a larger change.
- Fixed warnings about unused CSS selectors. Those warnings indeed **had an impact** because Svelte will exclude such rules from the compiled CSS. In other words, the components had CSS that looked like it would be applied but was not.
- Ignored all warnings about not using the `autofocus` attribute. Improving focus management is outside of the scope of this change.
- Fixed warnings about redundant ARIA roles by removing those roles.
- Fixed a warning about a missing ARIA role by adding that role.
- Ignored warnings about invalid values for `href`.
- Fixed warnings about usage of `$:` outside of root level. Those were compiled 1:1 as labeled statements. Removed the labels because they had no effect.
- Fixed deprecation warnings about using `type` over `lang` in Svelte components.

**Related issue(s)**
Fixes #3563


core/src/App.svelte
```diff
@@ -2126,7 +2125,6 @@
       on:handleSearchNavigation={handleSearchNavigation}
       bind:isSearchFieldVisible
       bind:displaySearchResult
-      bind:displayCustomSearchResult
```
ndricimrr, line 2129: This seems to not be needed to me as well as it doesn't seem to be used.
But maybe @JohannesDoberer thinks it should stay for some reason, or is it just an 'outdated remove todo' from the past?

core/src/navigation/TopNav.svelte
```diff
@@ -855,13 +861,6 @@
     }
   }
 
-  img {
-    &.fd-top-nav__icon {
-      height: 16px;
-      min-width: auto;
-    }
-  }
```
ndricimrr, line 863: Are we sure this is not being used ? Will also run & test but thought maybe its removed by mistake. 🤔

ndricimrr (commented):
Really nice work.
The warnings are gone and build seems pretty clean 👍 
Some minor comment only.

core/src/navigation/TopNav.svelte
camelCaseChris: I think I removed it because a) Svelte marked it as unused, based on the local checks within this component, so it was not included in the compiled code anyway and b) I found no indication in code that there can be images with that class.. That is why I did not wrap it into a `:global` rule like in other cases. But thanks for double-checking, I may have overlooked something!

core/src/App.svelte
JohannesDoberer: @ndricimrr 
How do you come to the conclusion that `displayCustomSearchResult ` is not needed?
I can see it is used in TopNav and will be shared with GlobalSearch and GlobalSearchCentered or I'm completely wrong?

camelCaseChris: That is true and in GlobalSearch(Centered) it seemed to do nothing, which is why I removed them. In one of them it came up as a Svelte warning, which is why I started looking. I think the question here is if I overlooked something and that prop does have a purpose.

JohannesDoberer: ah. Sorry my mistake.
Yes it can be removed.

ndricimrr (approved):
Spotted no issues

PR #3609, opened 2024-01-17
Add next version tag for container

### LINKED ISSUE #3608: Add 'since' flag to the container api reference docu

ndricimrr:
**Description**

- Adds a script to enable having NEXT_RELEASE tag when releasing new features/api calls for container
- This is not the same as core/client/etc release process. Since the container has its own release cycle a new custom script is added here. NEXTRELEASE was used there, and NEXT_RELEASE here since container and core/client have different version numbers.

**Related issue(s)**

Fixes #3608


JohannesDoberer (approved):
LGTM 👍

Issue #3611, opened 2024-01-19
Refactor package.json for testing-utilities


JohannesDoberer:
Currently the client-frameworks-support/testing-utilities/package.json will be copied to client-frameworks-support/testing-utilities/dist including devDependencies and scripts which are not needed in dist.

In addition the paths in `main` and `types` are not correct in package.json when it is located in dist.

```
 "main": "dist/index.js",
...
  "types": "dist/index.d.ts",
```
These causes an build error in angular applications which are using new builder
`"builder": "@angular-devkit/build-angular:application",`

Error message: 
```
An unhandled exception occurred: Failed to resolve entry for package "@luigi-project/testing-utilities". The package may have incorrect main/module/exports specified in its package.json.
See "/private/var/folders/9q/hsvwy2hj7h108y_xzdqf9p700000gn/T/ng-rltXps/angular-errors.log" for further details.
```

PR #3613, opened 2024-01-19
Update peerDependencies for client-support-angular

JohannesDoberer:
Angular 14 is not longer supported.
Added Angular 17 as peerDependency


JohannesDoberer:
blocked by #3611

PR #3623, opened 2024-01-24
Refactor testing-utilities

JohannesDoberer:


JohannesDoberer:
Can be tested with following versions:
```
""@angular/animations": "^17.1.0",
.....
"@luigi-project/client-support-angular": "2.8.1-dev.202401241555",
    "@luigi-project/testing-utilities": "2.8.1-dev.202401241555",
```

client-frameworks-support/testing-utilities/dist/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@luigi-project/testing-utilities",
+  "description": "Luigi testing utilities for standalone testing of microfrontends",
+  "license": "Apache-2.0",
+  "main": "index.js",
+  "types": "index.d.ts",
```
ndricimrr, line 6: Now there are two package.json's 
1. testing-utilities/package.json
2. testing-utilities/**dist**/package.json

Shouldn't the **dist** folder usually be in `.gitignore` .
When this package is published, first `npm run build` will push all to **/dist** folder, and then the output build will be released from here targeting the dist folder : https://github.com/SAP/luigi/blob/44f357891a7bfccfe39d8828e51f83cfd27401ca/scripts/npmPublish.sh#L151-L154

client-frameworks-support/testing-utilities/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
```
ndricimrr, line 6: Maybe instead of the other `/dist/package.json`, Change here /dist/index.js to /index.js and to put the dist folder in gitignore altogether 🤔

client-frameworks-support/testing-utilities/dist/package.json
JohannesDoberer: It's not necessarily that dist should be completeley in gitignore. E.g. imo core/public, client/public and client-support-ui5 also have an package.json which is not ignored and the public folder is equivalent to dist.

The reason why I want to avoid to copy the package.json is that script commands should not be in published package.json. In addition the devDependencies should be in the public package.json.

I don't get your point about the npmPublish completely. Maybe offline :)

ndricimrr: ahh true I forgot about that. I'll try to test and let you know.

ndricimrr (approved):
Tested and works as expected

client-frameworks-support/client-support-angular/projects/client-support-angular/package.json
```diff
@@ -7,8 +7,8 @@
     "url": "ssh://github.com/SAP/luigi.git"
   },
   "peerDependencies": {
-    "@angular/common": "^14.0.0 || ^15.0.0 || ^16.0.0",
-    "@angular/core": "^14.0.0 || ^15.0.0 || ^16.0.0",
+    "@angular/common": "^15.0.0 || ^16.0.0 || ^17.0.0",
```
JohannesDoberer, line 10: add ^14.0.0

PR #3646, opened 2024-02-13
3605 angular routing dynamic params

### LINKED ISSUE #3068: Add LuigiRouteStrategy to e2e-test-app
**Description**

**Reasons**

**Attachments**

### LINKED ISSUE #3605: Dynamic params with LuigiAutoRoutingService
**Description**

The problem occurs when we use LuigiAutoRoutingService, specifically when using dynamic parameters with child routes. In this scenario, they do not work as expected; dynamic parameters only function as intended when defined within the same route, not in the parent route.

In this example, the use of ':id' is not working as expected in the child route.
```
{
	path: “:id”
	data: {
		luigiRoute: “:id”
	}
	children: [
	{
		path: “general”
		data: {
			luigiRoute: “:id/general”
		}
	}	
	]
}
```

This code works as expected:
```
{
	path: “:id/general”
	data: {
		luigiRoute: “:id/general”
	}
}
```

**Expected result**

Param is always replaced in the luigiRoute path

**Actual result**

Param is not replaced in child route

camelCaseChris:
**Description**

- Make the unit tests for `client-support-angular` work.
- Add basic unit tests for `LuigiAutoRoutingService`.
- Change e2e Angular tests to use `LuigiAutoRoutingService` instead of its own custom routing service.

**Related issue(s)**
Fixes #3605 
Fixes #3068


JohannesDoberer (approved):
LGTM 👍 !

PR #3654, opened 2024-02-19
Cypress support testing utilities

JohannesDoberer:


client-frameworks-support/client-support-angular/projects/client-support-angular/src/lib/luigi-mock/luigi-mock.module.ts
```diff
@@ -23,155 +23,4 @@ import { LuigiMockEngine } from '@luigi-project/testing-utilities';
...
-      // Hide the added DOM element to avoid interferring/overlapping with other elements during testing.
-      luigiVisualizationContainer.setAttribute('style', 'display:none');
-      document.body.appendChild(luigiVisualizationContainer);
-    }
-    const line: HTMLDivElement = document.createElement('div');
-    line.textContent = data;
-    luigiVisualizationContainer.appendChild(line);
-  }
-}
+export class LuigiMockModule { }
```
ndricimrr, line 26: I don't remember why this class was ever needed to begin with. Seems like MockEngine below does the same thing - has the same code. So why not delete this file altogether if it has nothing inside 🤔

JohannesDoberer: Currently it's not possible to delete the whole module because it loads [LuigiMockEngine.initPostMessageHook](https://github.com/SAP/luigi/pull/3654/files#diff-3c43c8edc309f76af69c1c9a1b7f9b82077d54090261cd3a123e1456b2999460L9)
Cannot move this "header" to testing-utilities because there is no angular available.

JohannesDoberer: ...and it would be a breaking change when we change the loading mechanism of luigi mock engine

docs/luigi-testing-utilities.md
```diff
@@ -21,9 +21,10 @@ meta -->
 The [Luigi Testing Utilities](https://github.com/SAP/luigi/tree/main/client-frameworks-support/testing-utilities) are a set of auxiliary functions used to enhance the user experience while testing Luigi-based micro frontends. The functions abstract away Luigi-specific logic from the tester so that it is easier for them to mock and assert Luigi functionality. 
 
 ## LuigiMockUtil 
-This class contains certain utility helper functions needed when writing [protractor-based](https://www.npmjs.com/package/protractor) e2e tests. You can simply import this module into you project and then use an instance of it to test micro frontend functionality. 
+This class contains certain utility helper functions needed when writing [protractor-based](https://www.npmjs.com/package/protractor) e2e tests. You can simply import this module into you project and then use an instance of it to test micro frontend functionality.
+With version 2.9.0 this class can also used for [cypress-based](https://www.cypress.io/) e2e tests.
```
smahati, line 25:
```suggestion
With version 2.9.0 this class can also be used for [cypress-based](https://www.cypress.io/) e2e tests.
```
missing word

ndricimrr (approved):

PR #3655, opened 2024-02-19
Fix split view sizing with BTP layout

### LINKED ISSUE #3639: Fix splitview calculation and add Cypress test Splitview size check
Add a test to check that the splitview is created properly. 
Scenarios must take into account webcomponent main view as well as iframe based main view. 

Following is an example of what could go wrong: 

splitview

camelCaseChris:
**Description**

- Fix split view container/dragger being sized incorrectly. Background: the split view is positioned absolutely, with a left offset set to the width of the left navigation. There is no positioned element between the split view and the root element, so the offset is relative to the root. In BTP mode, the vertical navigation is also positioned, so the offset is then wrong. It has to be 0 instead.
- New test for the above change. Proposing a new test file for the E2E JS tests, assuming that there will be more tests for the BTP layout.

**Related issue(s)**
Fixes #3639


JohannesDoberer (approved):
LGTM 👍 !!

PR #3676, opened 2024-03-12
Luigi container and Compound Container API docs update

smahati:
**Description**

- Cosmetic changes to the documentation

### DIFF
--- a/container/typings/LuigiCompoundContainer.svelte.d.ts
+++ b/container/typings/LuigiCompoundContainer.svelte.d.ts
@@ -68,14 +68,14 @@ export default class LuigiCompoundContainer extends HTMLElement {
 
   /**
    * The hasBack value to be passed to the compound micro frontend.
-   * You can use it to mark if there is one or more preserved views. Useful when needing to show a back button. 
+   * It indicates that there is one or more preserved views. Useful when you need to show a back button. 
    * @since NEXT_RELEASE
    */
   hasBack: string;
 
   /**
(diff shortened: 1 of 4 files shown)


ndricimrr (changes requested):
This particular markdown file is not to be manually edited. 

We document new features/functions with **JSDoc** comment header descriptions and then use  our custom made scripts to generate and keep that in sync with the respective .md files. 
(Some .md files which don't relate to function descriptions, like tutorials, examples, guides, faq etc, are of course manually edited in the /docs folder where all .md files reside.)

In this case we only update the respective **JSDoc** Comments , so you should make your suggested changes instead on the in-code **JSDOC** instead : https://github.com/SAP/luigi/blob/main/container/typings/LuigiCompoundContainer.svelte.d.ts

Afterwards update the markdown file using the documentation script:
```
cd script
npm run docu
```
This should then also do an update to the .md files. 

Afterwards just commit the result.

ndricimrr (commented):
We have no strict rules regarding PR title names, but try to keep a concise description of what the PR does. 
This makes it much easier when going back in Commit history of the main branch as to what each PR did.

smahati (commented):
updated the jsdoc comments :)

ndricimrr (approved):
Looks good 👍 . I put it into To Accept column. When its ready to merge I will notify you.

ndricimrr:
You may merge now @smahati ✅

PR #3679, opened 2024-03-13
Fix loading indicator not going away when fast clicking multiple times to a nav node 

hardl:
clear timeout before setting it


ndricimrr (approved):
Able to reproduce and tested. Works as expected

PR #3680, opened 2024-03-15
Fix global nav focus issue

hardl:


ndricimrr (approved):
Also tested and seems to work as expected.

PR #3690, opened 2024-04-05
Add the container examples link to docu website side navigation

### LINKED ISSUE #3687: Add Container Examples to Docu Website Example Category
Screenshot 2024-04-03 at 16 06 43

Somewhere on this section, maybe at the bottom, we could add an External Link to the Container Examples to give more exposure to the Container.

Investigate how the current external links are added there, and add the "Container Examples" link as well.

Linking to this place: https://github.com/SAP/luigi/tree/main/container/examples 

Note: [Luigi Docu Page](https://docs.luigi-project.io/docs/getting-started) is in itself a Luigi Microfrontend based application. So the **external** option would be a Luigi node configuration. Search in our Docu how to add a new node on the left navigation that is shown with the external link icon, and add it appropriately.

smahati:
**Description**

- Added an external link to the container examples page in the sidebar.
- Added a note to check the path for working on the documentation as GitHub shows README files from nested folders and it's not clear which folder is being referred to.
- Changed some examples to show a different link being added to a category.

**Related issue(s)**

Resolves #3687

### DIFF
--- a/docs/navigation-configuration.md
+++ b/docs/navigation-configuration.md
@@ -155,10 +155,10 @@ To define all subsequent nodes, use the category label:
 {
   category: 'Links',
   externalLink: {
-    url: 'http://www.luigi-project.io',
+    url: 'http://www.luigi-project.io/docs/navigation-configuration',
     sameWindow: false
   },
-  label: 'Click here to visit the Luigi homepage',
+  label: 'Click here to visit the Luigi navigation doc',
 },
 ...
 ```
@@ -183,10 +183,10 @@ To define all subsequent nodes, use the category label:
 {
   category: 'anyId',
   externalLink: {
-    url: 'http://www.luigi-project.io',
+    url: 'http://www.luigi-project.io/docs/navigation-configuration',
     sameWindow: false
   },
-  label: 'Click here to visit the Luigi homepage',
+  label: 'Click here to visit the Luigi navigation doc',
 },
 ...
 ```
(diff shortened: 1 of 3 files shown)


docs/navigation-configuration.md
```diff
@@ -155,10 +155,10 @@ To define all subsequent nodes, use the category label:
 {
   category: 'Links',
   externalLink: {
-    url: 'http://www.luigi-project.io',
+    url: 'http://www.luigi-project.io/docs/navigation-configuration',
```
ndricimrr, line 158: The changes in this file are not needed. 

This part is a documentation of the feature for **externalLink**s. It is just an example configuration, not something that will affect the website build. 
For this example we want to have it simple and abstract. Very custom links like docs/navigation-configuration add confusion here. This is more like www.example.com kind of an example link that we have used.

ndricimrr (commented):
One suggestion. The rest looks good. Nice work

docs/navigation-configuration.md
smahati: Since we add another node with the same URL I thought it might be better, I can remove this or should I use something like www.example.com?

ndricimrr: You have already added the new node here **website/docs/static/public/navigation-nodes.json**.
That JSON file you changed will change how the documentation website is built. 

This _navigation-configuration.md_ here does not affect how the website is built. This is just a simple code example of how the feature is supposed to work. 

I would just revert this MD file here to what it was before and then it should be okay. No changes are needed here.

ndricimrr: I am reading the whole file, yes makes sense 👍 🤔

ndricimrr (approved):
Looks good 👍

Issue #3658, opened 2024-02-20
Improve webcomponent HTML attribute for Luigi Container


ndricimrr:
`<luigi-container webcomponent="false"></luigi-container>`
is often misused to have webcomponent feature set to false. However, due to [Boolean attribute for HTML](https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML) this will count as true, and false would be simply not setting the attribute at all. 

Document this properly and/or make it more intuitive for developers to use

PR #3695, opened 2024-04-11
WC html attribute for container

VincentUllal:
**Related issue(s)**

"Resolves" #3658


JohannesDoberer (approved):
LGTM 👍 !
Note: I would recommend to use the `strict equality operator`.

PR #3697, opened 2024-04-15
Light dom container

hardl:
no-shadow option for luigi container


container/src/LuigiContainer.svelte
```diff
@@ -216,12 +135,29 @@
...
...webcomponent' }, locale: { type: 'String', reflect: false, attribute: 'locale' }, theme: { type: 'String', reflect: false, attribute: 'theme' }, activeFeatureToggleList: { type: 'Array', reflect: false, attribute: 'active-feature-toggle-list' }, skipInitCheck: { type: 'Boolean', reflect: false, attribute: 'skip-init-check' }, nodeParams: { type: 'Object', reflect: false, attribute: 'node-params' }, userSettings: { type: 'Object', reflect: false, attribute: 'user-settings' }, anchor: { type: 'String', reflect: false, attribute: 'anchor' }, searchParams: { type: 'Object', reflect: false, attribute: 'search-params' }, pathParams: { type: 'Object', reflect: false, attribute: 'path-params' }, clientPermissions: { type: 'Object', reflect: false, attribute: 'client-permissions' }, dirtyStatus: { type: 'Boolean', reflect: false, attribute: 'dirty-status' }, hasBack: { type: 'Boolean', reflect: false, attribute: 'has-back' }, documentTitle: { type: 'String', reflect: false, attribute: 'document-title' }, allowRules: { type: 'Array', reflect: false, attribute: 'allow-rules' }, sandboxRules: { type: 'Array', reflect: false, attribute: 'sandbox-rules' } }, extend: customElementConstructor => {
```
ndricimrr, line 139: Was this format done for a reason in one line 🤔

ndricimrr: ok so I see that the prettier for svelte moved the whole section here automatically. and kept the think in a one line

container/cypress/e2e/test-app/wc/create-dynamic-wc-container.cy.js
```diff
@@ -78,4 +78,48 @@ describe('Create luigi-container dynamically', () => {
...
+    const htmlCode = `
+      <luigi-container data-test-id="luigi-container" id="lc">
+      </luigi-container>
+      <script>${scriptCode}</script>
+    `;
+    const stub = cy.stub();
+    cy.on('window:alert', stub);
+    cy.visit(tetsPage);
+    cy.get('.content').invoke('append', htmlCode);
+    // cy.get('[data-test-id="luigi-container"]').shadow().should('exist');
```
ndricimrr, line 120:
```suggestion
```

container/typings/LuigiContainer.svelte.d.ts
```diff
@@ -73,6 +73,12 @@ export default class LuigiContainer extends HTMLElement {
    */
   nodeParams: Object;
 
+  /**
+   * If set to true, luigi container webcomponent don't have a shadow dom.
```
ndricimrr, line 77: >  * If set to true, luigi container webcomponent don't have a shadow dom.

 Maybe sth like: 
 _If set to true, luigi container webcomponent will not be rendered in a shadow DOM_
 update the docu part also, guess its done pre-commit anyway

container/src/LuigiContainer.svelte
ndricimrr: I think its currently present and being used in v1 as per our [config](https://github.com/SAP/luigi/blob/main/prettier_config.json#L31), just with less sort filters. Newer versions I think [include the options ](https://github.com/sveltejs/prettier-plugin-svelte/blob/master/src/options.ts).

I think order though is not so important, but if we can with a small config change fix the one-liner would be easier to read.

walmazacn: I decided to exclude files with 'svelte:options' from being pretiffied - this fix is already commited to this PR
FYI @JohannesDoberer

walmazacn (approved):
LGTM 👍

ndricimrr (approved):
Also tested out and works well 👍

container/typings/LuigiContainer.svelte.d.ts
```diff
@@ -73,6 +73,12 @@ export default class LuigiContainer extends HTMLElement {
    */
   nodeParams: Object;
 
+  /**
+   * If set to true, the Luigi container webcomponent will not be rendered in a shadow DOM.
```
smahati, line 77:
```suggestion
   * If set to true, the Luigi container webcomponent will not use the shadow DOM for rendering.
```

walmazacn:
LGTM - all tests are passing and build is successful 👍

walmazacn:
Everything looks good after latest changes

PR #3699, opened 2024-04-16
Restructure container tests

### LINKED ISSUE #3562: Extend Container Integration Tests
+ Come out with a proper structure for integration tests files as it is currently not optimal
+ Add more integration tests
+ Issues like "timing issue when setting props" #3497 are hard to capture, maybe also add a test for this case if easy to do

### LINKED ISSUE #3720: Add Cypress Pipeline Record option for better debugging
Sometimes running Cypress on GithubActions pipeline gives a different behavior from running it locally, due to different enivronments present or any other reasons.

Cypress tests currently run headless on our CI and there is no way to see what happened. 

Enable the record feature for the different Cypress tests that we have. 

Starting with container and then the e2e-test-application.

Need to store record key on Github on our side.

ndricimrr:
**Description**

- Restructure container tests to make it easier to write Cypress tests in the future. 
- Add ReadMe's to make it easier to understand structure
...

**Follow-Up:** There is a long list of features/API calls/etc that exist in the **test-app** now, which are not added to the Cypress test. This PR aims to add some structure and separate concerns (iframe/wc/compound) since they were all mixed up in one large file. Also to get rid of the tailwind dashboard. As a Follow Up may reorganize the the JS webcomponent files and html test files gradually as we continue to increase Container E2E Test Coverage. 

**Related issue(s)**

Resolves #3562
Resolves #3720


bootstrap.js
```diff
@@ -6,19 +6,19 @@ const { exec } = require('child_process');
...
-  'client-frameworks-support/testing-utilities',
-  'client-frameworks-support/client-support-ui5',
-  'container',
-  'plugins',
-  'scripts',
-  'test/e2e-test-application',
-  'test/e2e-test-application/externalMf',
-  'test/e2e-js-test-application',
-  'client-frameworks-support/testing-utilities/test'
+  // './',
```
walmazacn, line 9: What's the point of keeping these lines commented? You may remove them as well as they're visible in git history anyway

bundle.js
```diff
@@ -2,16 +2,16 @@ const { exec } = require('child_process');
...
 // Array of folder names
 const foldersToBundle = [
-  'core',
-  'client',
-  'client-frameworks-support/testing-utilities',
-  'client-frameworks-support/client-support-angular',
-  'client-frameworks-support/client-support-ui5',
-  'container',
-  'plugins'
+  // 'core',
```
walmazacn, line 5: Same as previous comment

bootstrap.js
ndricimrr: sorry, it was not mentioned . This is a DRAFT PR, mistakenly turned into a PR, so it is not yet ready for review.

ndricimrr: still in Progress. I will ping you when Ready 👍

container/test-app/compound/compoundClientAPI.html
```diff
@@ -0,0 +1,219 @@
+<html>
+<head>
+  <meta name="viewport" content="width=device-width,initial-scale=1" />
+
+    <script type="module">
+        import '../bundle.js';
+        import MFEventID from '../bundle.js';
```
walmazacn, line 7: Isn't this import redundant as there's another one in line 44?

walmazacn (approved):
LGTM - I run tests locally and all of them are successful; same with localhost pages (2222 and 8080) - no errors in the console :tada:

container/test-app/index.html
```diff
@@ -1,670 +1,76 @@
...
-          // event: MFEventID.PathExistsEvent can be used as an event type to get the callback function
-          event.callback(true);
-        });
-
-        // setViewGroup Data listener
-        luigiContainer.addEventListener(MFEventID.SET_VIEW_GROUP_DATA_REQUEST, event => {
-          console.log('Set View Group Data Request received', event.detail, event);
-        });
-      });
-      let testDeferInitFalse = document.getElementById('test-defer-init-false');
```
JohannesDoberer, line 644: We have no example for this flag in the new test suite. Should we add one?

JohannesDoberer (approved):
LGTM 👍

container/test-app/index.html
ndricimrr: Yes. this one and maybe others too. I will create a Follow-Up dedicated to only adding tests for the new suite. This PR is already too big.

PR #3719, opened 2024-05-06
A11y for left nav if btp layout is enabled

JohannesDoberer:


ndricimrr (approved):
Tested with/without btp layout enabled, both working fine

PR #3728, opened 2024-05-16
Adjusts docu generation

### LINKED ISSUE #3708: Adjust docu generation to not trigger on all container changes
When trying to add new tests to **/container** the docu generation is being ran each time [here](https://github.com/SAP/luigi/blob/6c2a80f4158435bc49829a5e7a65d3c893fdef71/package.json#L65). 

Docu generation should only be triggered when only the relevant container docu parts are affected. 

Figure out which parts/files of **/container** folder are used for the docu generation part and only trigger docu generation when they are changed. Currently it is triggered on all changes. 

Relevant change to be made is most likely here:
https://github.com/SAP/luigi/blob/6c2a80f4158435bc49829a5e7a65d3c893fdef71/scripts/hooks/generate-docu.sh#L9

walmazacn:
**Description**

- Docu generation should only be triggered when only files from container source are affected

**Related issue(s)**

Resolves #3708

### DIFF
--- a/scripts/hooks/generate-docu.sh
+++ b/scripts/hooks/generate-docu.sh
@@ -6,7 +6,7 @@
 # ./scripts/hooks/generate-docu.sh
 
 function check_and_generate_docu() {
-  staged_changes=$(git diff --cached --name-only --diff-filter=ACM | grep -e "core/src/core-api" -e "client/src" -e "container/src" | wc -l)
+  staged_changes=$(git diff --cached --name-only --diff-filter=ACM | grep -e "core/src/core-api" -e "client/src" -e "container/typings" | wc -l)
   if [[ "$staged_changes" != *"0"* ]]; then
     echo "Changes in .js/.ts found. Building docu"
     cd scripts
--- a/scripts/hooks/generate-docu.sh
+++ b/scripts/hooks/generate-docu.sh
@@ -1,14 +1,14 @@
 #!/bin/bash
 
-# Checks if there are changed core-api or luigi-client js files
-# and generates docu if necessary
+# Checks if there are changed core-api, luigi-client or luigi-container
+# js/ts files and generates docu if necessary
 # Run this script only from Luigi root folder, else git adding won't work.
 # ./scripts/hooks/generate-docu.sh
 
 function check_and_generate_docu() {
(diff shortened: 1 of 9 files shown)


scripts/hooks/generate-docu.sh
```diff
@@ -1,14 +1,14 @@
...
-# Checks if there are changed core-api or luigi-client js files
-# and generates docu if necessary
+# Checks if there are changed core-api, luigi-client or luigi-container
+# js/ts files and generates docu if necessary
 # Run this script only from Luigi root folder, else git adding won't work.
 # ./scripts/hooks/generate-docu.sh
 
 function check_and_generate_docu() {
-  staged_changes=$(git diff --cached --name-only --diff-filter=ACM | grep -e "core/src/core-api" -e "client/src" -e "container"| wc -l)
+  staged_changes=$(git diff --cached --name-only --diff-filter=ACM | grep -e "core/src/core-api" -e "client/src" -e "container/src" | wc -l)
```
ndricimrr, line 9: From the following lines where the docu generation script is written:
https://github.com/SAP/luigi/blob/53e757dbca4de4403de3f66500ccec4bb30582ef/scripts/package.json#L7-L11

It seems like for auto-generating Luigi Container package documentation we need to check if changes are done in the `container/typings/LuigiContainer.svelte.d.ts` and `container/typings/LuigiCompoundContainer.svelte.d.ts` files. 

So instead of `container/src` change it to `container/typings` folder and that should work. make sure to test afterwards.

You can test if this works by:
 - Changing something in the two typings files mentioned above.
 - Trying to commit something
 - A [pre-commit hook](https://github.com/SAP/luigi/blob/main/package.json#L65) will be ran before the commit is actually performed. Check your Terminal if the  `npm run docu` command is also ran. Also, you should be able to see changes performed on the respective markdown files [here](https://github.com/SAP/luigi/blob/main/docs/luigi-container-api.md) and [here](https://github.com/SAP/luigi/blob/main/docs/luigi-compound-container-api.md)
 
 Try the same thing with changing another random file in container/src. In that case, upon committing `npm run docu` command should not run in your local terminal. 
 
 
 (If the pre-commit hook does not run in your machine, make sure to run npm run install on root folder of the repo and try again)

walmazacn: You're right - thank you for clarification. Fix is already applied

ndricimrr (approved):

PR #3734, opened 2024-05-20
Adds alternative for legacy OIDC client

### LINKED ISSUE #2446: Alternative for oidc-client
In our current implementation we are using _oidc-client_ v1.11.5. This project is not longer maintained.
Follow-up project is https://github.com/authts/oidc-client-ts

Keep the old oidc-client library in _/plugins/auth/src/auth-oidc/index.js_ that we also can support the implicit flow.
It should be possible that the user can choose which plugin/library he want's to use.

walmazacn:
**Description**

- adds new 'oidc-client-ts' package to support PKCE
- adjusts scripts and docs to reflect those changes

**Related issue(s)**

Resolves #2446


plugins/auth/src/auth-oauth2/index.js
```diff
@@ -4,17 +4,15 @@ import { Helpers } from '../helpers';
 export default class oAuth2ImplicitGrant {
   constructor(settings = {}) {
     const defaultSettings = {
+      accessTokenExpiringNotificationTimeInSeconds: 60,
```
JohannesDoberer, line 7: Was it necessary to rename this property as well? If not, needs to be clarified with Philipp, because of breaking change.

plugins/auth/src/auth-oidc/index.js
```diff
@@ -1,18 +1,17 @@
...
       loadUserInfo: false,
-      automaticSilentRenew: false,
-      accessTokenExpiringNotificationTime: 60,
-      thirdPartyCookiesScriptLocation: '',
-      logoutUrl: window.location.origin + '/logout.html',
+      post_logout_redirect_uri: window.location.origin + '/logout.html',
+      redirect_uri: window.location.origin,
+      response_type: 'code',
       silent_redirect_uri: window.location.origin + '/assets/auth-oidc/silent-callback.html'
     };
```
JohannesDoberer, line 17: Logout does not work for me. Getting the following error:
```
extendedConfiguration.bundle.js:2 Uncaught (in promise) TypeError: this.client.createSignoutRequest is not a function
    at e.value (extendedConfiguration.bundle.js:2:78458)
    at K2.logout (auth-layer.js:146:32)
    at pe (AuthorizationSimpleProfileMenu.svelte:180:18)
    at AuthorizationSimpleProfileMenu.svelte:163:9
   ```
_createSignoutRequest_ does not exist on UserManager (anymore)
Line 90

**Review by JohannesDoberer — COMMENTED** [review]
We need to clarify with Philipp if the renaming of the old one to legacy and the new one get's the old name.

File: plugins/auth/src/auth-oauth2/index.js (hunk 1 above)
Comment (walmazacn) [reply plugins/auth/src/auth-oauth2/index.js:7]: You're right - let me fix it

File: plugins/auth/src/auth-oidc/index.js (hunk 2 above)
Comment (walmazacn) [reply plugins/auth/src/auth-oidc/index.js:17]: Let me check what's going on

Comment (walmazacn) [reply plugins/auth/src/auth-oidc/index.js:17]: There was small change for 'UserManager' API in 'oidc-client-ts' - I fixed it

**Review by JohannesDoberer — APPROVED** [review]
Tested and works 👍 
Let's clarify the open points regarding renaming

File: scripts/npmPublish.sh [hunk 3]
```diff
@@ -138,6 +138,12 @@ else
   fi
 
   # Luigi Oidc Plugin
+  if ( prepublishCheck "plugins/auth/public/auth-oidc-pkce" ); then
```
JohannesDoberer, line 141: Please remove it from the general release flow.
We wanted to create an own release cycle like container, client-support-ui5 or cra-release.

plugins/auth/public/auth-oidc-pkce/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "tag": "luigi-plugin-auth-oidc-pkce"
+  },
+  "keywords": [
+    "luigi",
+    "core",
+    "oidc",
+    "openid connect",
+    "provider"
+  ],
+  "version": "2.13.0"
```
JohannesDoberer, line 20: own release cycle starting at 1.0.0

JohannesDoberer (changes requested):

plugins/auth/public/auth-oidc-pkce/package.json
walmazacn: fixed

JohannesDoberer (approved):
LGTM 👍

scripts/tools/release-cli/release-cli.js
```diff
@@ -40,6 +40,7 @@ const pkgJsonPaths = {
   client: path.resolve(base, 'client', 'public', 'package.json'),
   authOAuth2: path.resolve(base, 'plugins', 'auth', 'public', 'auth-oauth2', 'package.json'),
   authOIDC: path.resolve(base, 'plugins', 'auth', 'public', 'auth-oidc', 'package.json'),
+  authOIDC_pkce: path.resolve(base, 'plugins', 'auth', 'public', 'auth-oidc-pkce', 'package.json'),
```
JohannesDoberer, line 43: It's not needed, because we have an own release cylce for this package, which means this script is not responsable for the release of auth-oidc-pkce.

scripts/tools/publish-nightly.js
```diff
@@ -39,6 +39,7 @@ const packagePaths = {
   client: ['client'],
   oauth2: ['plugins', 'auth', 'src', 'auth-oauth2'],
   oidc: ['plugins', 'auth', 'src', 'auth-oidc'],
+  oidc_pkce: ['plugins', 'auth', 'src', 'auth-oidc-pkce'],
```
JohannesDoberer, line 42: not needed because of different release cycle

scripts/tools/publish-nightly.js
```diff
@@ -47,6 +48,7 @@ const publishPaths = {
   client: ['client', 'public'],
   oauth2: ['plugins', 'auth', 'public', 'auth-oauth2'],
   oidc: ['plugins', 'auth', 'public', 'auth-oidc'],
+  oidc_pkce: ['plugins', 'auth', 'public', 'auth-oidc-pkce'],
```
JohannesDoberer, line 51: not needed because of different release cylce

JohannesDoberer (changes requested):

plugins/auth/public/auth-oidc-pkce/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "tag": "luigi-plugin-auth-oidc-pkce"
+  },
+  "keywords": [
+    "luigi",
+    "core",
+    "oidc",
+    "openid connect",
+    "provider"
+  ],
+  "version": "1.0.0"
```
JohannesDoberer, line 20: I think we need a different version for a beta/rc release

JohannesDoberer (approved):
LGTM 👍

PR #3744, opened 2024-05-22
Fixes getUserSettings method

### LINKED ISSUE #3434: getUserSettings should return empty object instead of null
Currently `getUserSettings` returns null but it should return empty object.
Fix for classic luigi-client and luigi-client wc.

walmazacn:
**Description**

- fixes fixes 'getUserSettings' method
- fixes related unit tests

**Related issue(s)**

Resolves #3434


JohannesDoberer (approved):
LGTM 👍

PR #3745, opened 2024-05-23
Adds prettier commit hook for svelte files

### LINKED ISSUE #2946: Add prettier commit hook to svelte files
Svelte files are not automatically prettified on commit, leading to inconsistent prettifying that depends on VS Code local styling plugins. 

Install **prettier-plugin-svelte** on **core/package.json** and using the prettier CLI command should be enough to prettify automatically:

`prettier --write --plugin-search-dir=. ./**/*.svelte` // limit to core/src initially

walmazacn:
**Description**

- adds prettier commit hook for svelte files

**Related issue(s)**

Resolves #2946


package.json
```diff
@@ -39,7 +39,8 @@
       "lodash.template": ">=4.5.0",
       "meow": "8.1.2",
       "mixin-deep": ">=1.3.2",
-      "prettier": "^1.15.3",
+      "prettier": "^3.2.5",
```
ndricimrr, line 42: **Prettier 1** -> **Prettier 3** is a major version upgrade. ⚠️ 

Before doing such change you must check what breaking changes there were between v1 and v3 before upgrading.
 Need to make sure it doesn't cause any issues.
Code might still compile with no problem. 

Problem is when you create a new PR, that changes just 1 code line, you also get 100s of lines of non-sense code styling ([Example](https://github.com/SAP/luigi/pull/3745/files#diff-4bfc2a7ab3f5e1777cc276378c9b396cbf38bae43de4418980a50412c86f9b70) ) . This makes code  reviews difficult as you can't see easily what code changes were made by Developer and which by prettier commit hook. 

Need to assess first what "damage" the new prettier version causes by running prettier on all JS/TS relevant code files on the repo, and see how many file changes it makes. 

Too see the effects of the new prettier version, run:
`npm run full-code-quality-prettier`

on root folder. And check with `git status` , how many files were affected. 

_Note_: This is not  just svelte-specific comment, as prettier affects HTML, JS and TS files in our repo.

ndricimrr (changes requested):
This issue has been stale for long time because there were concerns back then that needed to be thoroughly researched before moving with a solution as it wasn't very straight forward. 

Aside from the comments left under this thread 'prettier-plugin-svelte' seems like a good candidate for prettifying our Svelte files and it does its job. 

The changes of this plugin though need to be examined thoroughly. 

Please test the scripts locally beforehand.

Major changes spotted ⚠️ Needs more investigation if there are more: 
1. New prettier switches all Svelte Files from singleQuoation to Double Quotation
```
  import Breadcrumb from './navigation/Breadcrumb.svelte';
  import Breadcrumb from "./navigation/Breadcrumb.svelte";
```
  We prefer singleQuotation instead. 
  
2. New Prettier adds quotation marks to curly brackets { }

```
   <TabNav
        pathData={navigationPath}
        {pathParams}
        on:handleClick={handleNavClick}
        {resizeTabNavToggle} />
```
        
  becomes with new Prettier version
        
```
<TabNav
        pathData="{navigationPath}"
        {pathParams}
        on:handleClick="{handleNavClick}"
        {resizeTabNavToggle} />
```
The effects of such style change need to be investigated thoroughly. Check Svelte and Prettier-Plugin-Svelte community first. 
(example: If   `pathData="{navigationPath}"` becomes invalid syntax we would have to change them back manually)

Starting point: 
https://github.com/sveltejs/svelte/issues/7925#issuecomment-1273685939
https://github.com/sveltejs/prettier-plugin-svelte/issues/348 
https://github.com/sveltejs/svelte/issues/7925

package.json
walmazacn: You're definitely right I shouldn't introduce breaking change that would affect many files in this PR - I downgraded both prettier and svelte plugin to version one. After running 'full-code-quality-prettier' command only svelte files are affected (having your code preferences in mind)

walmazacn: I checked all the options, but this plugin (or Prettier itself) doesn't allow to ignore specific language feature (like '@const' tag) - the incorrect plugin behavior was fixed in version 2.6.0:
https://github.com/sveltejs/prettier-plugin-svelte/blob/master/CHANGELOG.md#260

That's why I think we have to use comments for ignoring formatting just like I proposed and we can remove them when we decide to upgrade Prettier at least to version 2 (same goes for the plugin)

ndricimrr (approved):
Looks good 👍 . 

Would be nice to apply all prettier changes in one follow-up PR, right after merging this one. 

Since it would change lots of `.svelte `files, we would try to merge the follow-up a little quicker, so there's less Merge conflicts (in case someone else is working on some .svelte files at the moment)

package.json
ndricimrr: ok. then we stick to the comments 👍

PR #3748, opened 2024-05-23
Fiddle theming example

JohannesDoberer:


website/fiddle/src/App.svelte
```diff
@@ -48,6 +48,8 @@
     document.head.appendChild(style);
     
     const core = document.createElement('script');
+    // core.setAttribute('type', 'module');
```
walmazacn, line 51: I think it's not needed - at least if this PR is ready for review ;)

walmazacn, line 51: I feel this part can be removed - unless it's needed for some further work

walmazacn (approved):
Everything looks good

ndricimrr (approved):

PR #970, opened 2019-11-15
Third party cookie check

JohannesDoberer:

PR #180, opened 2018-10-30
Token refresh

y-kkamil:
**Description**

Changes proposed in this pull request:

Add possibility to silently refresh token using OpenID Connect `prompt=none` param.

* For now, if authorisation is set to something different than OpenID Connect and/or silent token refresh is disabled, user will be redirected to the logout page without any warning - not good if he has some unsaved work. There is a follow-up for that - #179

**Related issue(s)**
See also #179

PR #3754, opened 2024-05-27
Removes third party cookie check

### LINKED ISSUE #3641: Remove third-party-cookie check functionality
...as they will be blocked anyway by all major browsers.

Potentially a breaking change -> https://docs.luigi-project.io/docs/general-settings/?section=third-party-cookies-support-check

walmazacn:
**Description**

- removes third party cookie check
- adjusts related documentation

**Related issue(s)**

Resolves #3641


docs/authorization-configuration.md
```diff
@@ -136,26 +136,21 @@ auth: {
...
+When Luigi fails to renew the token and then logs the user out, it adds the `?reason=tokenExpired` query parameter to the logout page redirect URL. Use this parameter to set a logout page.
 
-When Luigi fails to renew the token and then logs the user out, it adds the `?reason=tokenExpired&thirdPartyCookies=[VALUE]` query parameters to the logout page redirect URL. Luigi replaces **[VALUE]**  with one of these options:
-- `disabled` means that third party cookies are disabled.
-- `enabled` means that the browser supports third party cookies.
-- `not_checked` means that the script was not provided in **thirdPartyCookiesScriptLocation** or it could not be loaded.
-
-Use these parameters to set a logout page.
+Notice - starting from version 3.0 Luigi doesn't support third-party cookies in any way.
```
ndricimrr, line 153: I like this Notice. But I feel in this way it would be overlooked. We usually have some custom warning box for such cases. Check screenshot below and respective docu part to do sth similar 

Screenshot 2024-05-29 at 17 12 31

client/src/lifecycleManager.js
```diff
@@ -162,11 +162,9 @@ class LifecycleManager extends LuigiClientBase {
...
     }
     if (luigiCookie === 'luigiCookie=true') {
-      window.parent.postMessage({ msg: 'luigi.third-party-cookie', tpc }, '*');
       document.cookie = luigiCookieKey + '=; Max-Age=-99999999; SameSite=None; Secure';
     } else {
       tpc = 'disabled';
-      window.parent.postMessage({ msg: 'luigi.third-party-cookie', tpc }, '*');
       console.warn('Third party cookies are not supported!');
     }
   }
```
ndricimrr, line 170: This whole function here, seems like it was created on this PR
https://github.com/SAP/luigi/pull/970/files long time ago. 
I am guessing that since we remove third party cookie check functionality, we undo all that was done in that PR.
so then remove all the _tpcCheck function and its usage altogether. 

That would make sense, but lets also discuss it in daily further. (ex: Should we remove it altogether/Should we keep a console warning anyway that third party cookies are not supported, or do we let Chrome log its own warnings )

docs/general-settings.md
```diff
@@ -21,6 +21,8 @@ meta -->
 
 The configuration file contains a section called **Settings** where you can configure additional Luigi options.
 
+Notice - starting from version 3.0 Luigi doesn't support third-party cookies in any way.
```
ndricimrr, line 24: Adding the same like in the screenshot shown above would be nice.

plugins/auth/src/auth-oidc/index.js
```diff
@@ -131,12 +131,7 @@ export default class openIdConnect {
         case 'login_required':
         case 'account_selection_required':
         case 'consent_required': // possible cause: disabled third party cookies in the browser
```
ndricimrr, line 133: // possible cause: disabled third party cookies in the browser
consent_required case seems to have been added due to 'disabled third party cookies in the browser' . Does this mean that we do not need this case, or the adjustment below should be enough ?

ndricimrr (commented):
Left some comments.

Would ask for @smahati to review all these **docu** changed files here as well.

plugins/auth/src/auth-oidc/index.js
walmazacn: In my opinion we should leave this case and just adjust the action - to be sure backward compatibility is kept

JohannesDoberer:
Check if older core with new clients and wise versa are running w/o errors.

walmazacn:
> Check if older core with new clients and wise versa are running w/o errors

I checked everything once more and in my opinion no API method is affected directly - I just moved code chunks related to 3rd party cookie check. In other words, after these changes core should work with client as expected no matter which version is used

docs/authorization-configuration.md
```diff
@@ -136,26 +136,21 @@ auth: {
 - **default**: `60` seconds
 #### thirdPartyCookiesScriptLocation
 - **type**: string
-- **description**: is the URL to the page containing third-party cookies support check. For details, see [Third-party cookies and silent token refresh section](#Third-party-cookies-and-silent-token-refresh).
+- **description**: is the URL to the page containing third-party cookies support check. For details, see [Silent token refresh section](#silent-token-refresh).
```
smahati, line 139: is it still the url to the "page containing third-party cookies support check"?

docs/authorization-configuration.md
```diff
@@ -136,26 +136,21 @@ auth: {
...
 
-### Third-party cookies and silent token refresh
+### Silent token refresh
 
 The OpenID Connect configuration allows you to specify the **automaticSilentRenew** option. When set to `true`, Luigi attempts to automatically renew the token in the background before it expires. Be aware that this mechanism requires the browser to support [third-party cookies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#Third-party_cookies).
 
-To detect whether the user's browser supports the mechanism, use the script in the [`third-party-cookies`](https://github.com/SAP/luigi/tree/main/core/third-party-cookies) catalog. Deploy this file on a domain different from your main application's and set **thirdPartyCookiesScriptLocation** to the `init.html` file. During initialization, Luigi detects the cookies support and produces a warning in the console if cookies are disabled in the user's browser.
+When Luigi fails to renew the token and then logs the user out, it adds the `?reason=tokenExpired` query parameter to the logout page redirect URL. Use this parameter to set a logout page.
```
smahati, line 151:
```suggestion
When Luigi fails to renew the token it logs the user out and  adds the `?reason=tokenExpired` query parameter to the logout page redirect URL. Use this parameter to set a logout page.
```
how should this parameter be used to set the logout page? Luigi adds the token expired parameter and the second one does not exist anymore.

smahati (changes requested):

smahati:
I assume that the requirement of third party cookie browser support isn't affected by this change as it is not handled by Luigi, if not do we need to make some changes here?

https://github.com/SAP/luigi/pull/3754/files#diff-96a643e5e4b2ef94118dc8efd3f937f762fa3d2ef6fdcd06249d16713c8b14f2L149

docs/authorization-configuration.md
smahati: yeah, I read the previous version to see if I could find out how it works but it wasn't clear. It may mean something like set's the _parameter_ for the logout page and not the page itself.

smahati: @ndricimrr I've made a suggestion that I was talking about in the daily. Please check and confirm.

docs/authorization-configuration.md
```diff
@@ -134,28 +134,20 @@ auth: {
...
 #### profileStorageInterceptorFn
 - **type**: function
 - **description**: provides a function to mutate the profile values of the [JSON Web Token (JWT)](https://jwt.io) before it gets stored in Luigi. It allows the removal of values like **email** to comply with data privacy restrictions. Since it is async, you could use it to enrich the profile data, but it should not get mixed up with **userInfoFn** which is specifically designed to define user information.
 
-### Third-party cookies and silent token refresh
+### Silent token refresh
 
 The OpenID Connect configuration allows you to specify the **automaticSilentRenew** option. When set to `true`, Luigi attempts to automatically renew the token in the background before it expires. Be aware that this mechanism requires the browser to support [third-party cookies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#Third-party_cookies).
```
ndricimrr, line 146: I went through this silent token part section once more. And I think the following line:

> The OpenID Connect configuration allows you to specify the **automaticSilentRenew** option. When set to `true`, Luigi attempts to automatically renew the token in the background before it expires. Be aware that this mechanism requires the browser to support [third-party cookies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#Third-party_cookies).

The text indicates that **third party cookies**(TPC) and **silent renew** come together as a 'package'. So if we remove **TPC** check and all support for **TPC** it makes no sense to keep the **automaticSilentRenew** option altogether. 

We were able so far to offer this functionality ourselves by leveraging **TPC**. If **TPC** is gone then we cannot offer the functionality anymore. 

I think then the whole **### Silent token refresh** can go altogether, and with it the silent token refresh functionality. 

Looking back, this Warning on here, indicates that without TPC, silent renewal might not work. [Check Warning](https://github.com/SAP/luigi/pull/180/commits/79ecacebdbe16f84dc113a3516a68b7c6b8ec8e9#diff-373c23867778f0930e352885d2def2a1a72f90ad22320dc173e3c421c6dd621eR6)  on the [Pull Request that first added Token Refresh ](https://github.com/SAP/luigi/pull/180/files). 

I would guess in this case developers would have to implement that on their own. Will ask this again in the daily.

ndricimrr (approved):
From the discussion looks good as it is 👍

docs/authorization-configuration.md
walmazacn: @smahati I applied your suggestion - please verify

hardl (changes requested):
Please revert all changes in /container/*, as it is outside of luigi core release cycle, so having it in a 3.0 branch can be confusing.

PR #3760, opened 2024-05-29
fix flaky compatibility test

hardl:
potential fix under the assumption that the failing is caused by redirect taking too long and therefore the 
cy.get('body') is invoked on the main luigi page before redirect to the login page


ndricimrr (approved):
looks promising. Maybe just merge it directly and we see in main how it goes 🤔

PR #3761, opened 2024-05-29
Fixes issue with set anchor method

### LINKED ISSUE #3399: setAnchor for wc mf in modal
dbd

Notice:
Modal will be closed because handleRouteChange calls closeModal.
CloseModal checks the dirty state on iframe mf which is async which means resetMicrofrontendModalData will be called later. We do not have this for wc.

Due to the async resetMicrofrontendModalData, the function `await this.shouldShowModalPathInUrl();` will be callled earlier and modal data url won't be lost.

walmazacn:
**Description**

- fixes issue with set anchor method for WC MF in modal

**Related issue(s)**

Resolves #3399


ndricimrr:
I checked the issue and it seems like this originated from [this comment ](https://github.com/SAP/luigi/pull/3389#discussion_r1291324999)which helped reproduce the issue. 

From the commit log that followed thereafter that comment it seems like it was decided that we just do not support setAnchor functionality for modals (and possibly other special microfrontends like splitview, drawer etc) at the moment as further investigation on how to do it is needed.

We have introduced the isSpecialMF predicate for this [here](https://github.com/SAP/luigi/pull/3389/commits/5c8d446c684a02c67b21333c8c96cd922cf152db#diff-74a3439759e183a2afe1666d7a453ed72049a67086d2b1d2b2c1d980f771644cR66). In such a case if you follow the reproduce steps, the setAnchor function call is simply not reached because we disabled it. 

So I think the changes on this PR do not address that currently, and will need to change the issue to be a feature request than a bug. Discuss it in daily

ndricimrr (commented):
Tested some edge cases and found some potential scenario where this could break

core/src/App.svelte
```diff
@@ -924,6 +929,10 @@
    * @param goBackContext the goBack context that is passed through when closing the modal
    */
   const closeModal = (index, isClosedInternal, goBackContext) => {
+    if (mfModalList?.length && !mfModalList[index]) {
+      index = mfModalList.length - 1;
+    }
```
ndricimrr, line 934: the **closeModal** function here, must '_close the modal given the respective modal index._' as described in the description. The given index should be given correctly. If this method receives an incorrect index, the function should fail. This lets us know, that we call this function with incorrect parameters somewhere. I would remove this check here and find the issue someplace else, where needed.

core/src/App.svelte
```diff
@@ -375,13 +375,18 @@
...
+
       // TODO: check if bookmarkable modal is interferring here
       if (!isValidBackRoute(pv, path)) {
         preservedViews = [];
         Iframe.removeInactiveIframes(node);
       }
-      for (let i = mfModalList.length; i--; ) {
-        closeModal(i);
+
+      if (!mfModalList.some((item) => item.modalWC && item.mfModal?.displayed)) {
```
ndricimrr, line 385: This condition is a bit hard to read directly. If I can read it correctly this means that:
```
"if in the list of modals there is at least an item that is WC-based and is currently at the top of the stack=displayed"
         don't do anything
else 
        close all modals
```
        

 Maybe the developer had another scenario where they have WC Based wc and displayed and want to close the modals also, not 'do nothing'.
 
I feel like in this way, can always find a failing scenario. For example, now, this fails to close all modals:
```
Luigi.navigation().openAsModal('/test/modalwc', { size:'l', keepPrevious: true});
Luigi.navigation().openAsModal('/test/modalwc2', { size:'m', keepPrevious: true});
Luigi.navigation().openAsModal('/test/overview', { size:'s', keepPrevious: true});
// navigate away
Luigi.navigation().navigate('/test/overview');
```

Question is though, instead of solving scenario after scenario, maybe rethink, what do we want to do when setAnchor is called inside a modal WC. 
Maybe closing all modals and leaving the anchor at the main navigation could also be valid. I am not sure, I'll ask in daily again, since its not a common request.

walmazacn: After my latest fix the scenario above will not fail, as all modals are closed after navigating away (separate process). The condition you're mentioning happens after route is updated, so it won't break anything as all modals are closed already

PR #3764, opened 2024-06-04
cra-template dependencies update

JohannesDoberer:


walmazacn (approved):
I tested it locally and it seems to be working fine

cra-template/template.json
```diff
@@ -1,28 +1,29 @@
...
-        "serve": "webpack-dev-server --mode production"
-      }
+  "package": {
+    "dependencies": {
+      "@luigi-project/client": "^2.13.0",
+      "@luigi-project/core": "^2.13.0",
+      "react": "^18.3.1",
+      "react-dom": "^18.3.1",
+      "react-router-dom": "^6.23.1",
+      "react-scripts": "5.0.1"
```
ndricimrr, line 9: If this react-scripts dependency was not needed before, I would remove it. Unless removing it brakes something. We dont use them in the scripts section it seems

ndricimrr (commented):
Also need to update the package.json version of the cra-template since we would need to release it after merging this PR. To avoid creating another PR, I think its ok if you just bump the version in this PR as well. 
https://github.com/SAP/luigi/blob/main/cra-template/package.json#L4

I would guess to 1.0.2

ndricimrr (approved):

PR #3775, opened 2024-06-06
Removes not needed warning message

### LINKED ISSUE #3767: Luigi Container: warning is logged in the browser console for non-luigi window messages
**Description**
Luigi container listens to every message from the microfrontend iframe and process it as a potential Luigi Container event. In our application some 3rd party scripts actively use `window.postMessage`. Since Luigi tries to process all the messages, there is a warning logged in the browser console every time there is a message not known by Luigi.

**Expected result**
No warning in the browser console. If Luigi container cannot recognize the message received, there is no need for `console.warn` here - https://github.com/SAP/luigi/blob/main/container/src/services/container.service.ts#L180

**Actual result**
Warning messages in the browser console:
Screenshot 2024-06-05 at 16 35 01

**Steps to reproduce**
Can be reproduced by sending any custom message from iframe to a host application. Custom message should not contain `event.data.msg` or its value should be something different from well-known Luigi Container events listed in https://github.com/SAP/luigi/blob/main/container/src/constants/internal-communication.ts

**Troubleshooting**
`-`

walmazacn:
**Description**

- removes not needed warning message
- updates related unit test

**Related issue(s)**

Resolves #3767


container/src/services/container.service.ts
```diff
@@ -69,122 +69,126 @@ export class ContainerService {
...
-                this.dispatch(Events.UPDATE_MODAL_PATH_DATA_REQUEST, targetCnt, event);
-                break;
-              case LuigiInternalMessageID.CHECK_PATH_EXISTS_REQUEST:
-                this.dispatch(Events.CHECK_PATH_EXISTS_REQUEST, targetCnt, event);
-                break;
-              case LuigiInternalMessageID.SET_DIRTY_STATUS_REQUEST:
-                this.dispatch(Events.SET_DIRTY_STATUS_REQUEST, targetCnt, event);
-                break;
-              default:
-                console.warn('Functionality not yet implemented: ', msg);
```
ndricimrr, line 180: I would expect only this **console.warn** here to be affected/removed in this PR to solve the referenced issue. 

Refactoring is good and **needed** but better leave it to another Issue/PR as this area of code here is very often used and we want to avoid accidentally introducing a breaking change or bug. This needs to be throughly tested that it works the same and that would take some time. For such refactoring I would leave it as is for now.

ndricimrr (approved):
Looks good

PR #3780, opened 2024-06-10
Get current route

JohannesDoberer:
added App.svelte as internal variable to Luigi


core/src/core-api/_internalLinkManager.js
```diff
@@ -87,13 +88,23 @@ export class linkManager extends LuigiCoreAPIBase {
...
 
   fromVirtualTreeRoot() {
     this.options.fromContext = null;
     this.options.fromClosestContext = false;
     this.options.fromVirtualTreeRoot = true;
+    this.options.fromParent = false;
+    return this;
+  }
+
+  fromParent() {
```
ndricimrr, line 103: Why the need to add fromParent or just to have it in sync with the rest 🤔

core/src/core-api/_internalLinkManager.js
```diff
@@ -125,6 +136,10 @@ export class linkManager extends LuigiCoreAPIBase {
     });
   }
 
+  getCurrentRoute() {
```
ndricimrr, line 139: would be nice to just have a small JSDoc on this fuction only. More specifically to explain how we get to the **buildPathForGetCurrentRoute** function through _app -> ctx -> props -> . Interesting syntax but cumbersome for someone reading the line for the first time.

ndricimrr (commented):
Seems to be added so far for Core-Api , Core-WC-clientAPI, but not Container WC-based ClientAPI [here](https://github.com/SAP/luigi/blob/main/container/src/services/webcomponents.service.ts#L91). Do we plan to add that there too ?

core/src/core-api/_internalLinkManager.js
JohannesDoberer: Yes, to keep it in sync.
Tested different scenarios with our current functionality like:
_...fromClosestContext().getCurrentRoute()_
_...fromContext('ctx').getCurrentRoute()_
and realized that getParent().getCurrentRoute is not supported.

ndricimrr (approved):
Looks good

Issue #3620, opened 2024-01-24
Algolia Search not indexing part of docu


ndricimrr:
**Description**

When searching for some parts of documentation, such as api functions, they don't show up on search even though they are part of the Docu files.

**Reasons**

Unknown

**Attachments**

<img width="525" alt="Screenshot 2024-01-24 at 12 03 49" src="https://github.com/SAP/luigi/assets/18335489/d01b351e-8ab3-4bb1-8970-2c3d7251c1ce">
<img width="918" alt="Screenshot 2024-01-24 at 12 04 10" src="https://github.com/SAP/luigi/assets/18335489/251cd364-535b-43cb-827c-cfebc4f89e47">
<img width="918" alt="Screenshot 2024-01-24 at 12 04 10" src="https://github.com/SAP/luigi/assets/18335489/aa5af737-474b-4228-a709-f104bed4a324">

PR #3783, opened 2024-06-12
Fix Algolia Search not indexing part of docu

amilewskaa:

- add filters parameter with `type:content` to search method (as Algolia Docsearch v3 does not support `only_content_level`)
- update appId and apiKey for algoliaSearch - app has configured Algolia Crawler to automatically update records once a week
- provide check for `hit.content` existance, if not - return empty string to avoid unexpected errors

**Related issue(s)**
#3620


JohannesDoberer (approved):
LGTM 👍

PR #3788, opened 2024-06-20
Adds Cypress tests for 'updateContext' method

### LINKED ISSUE #3786: Add Container Cypress Test for updateContext
We often forget to test both iframe and wc functionality, so need to make sure they always work even after updating functions. 

- [x] Add test for updateContext WC - based
- [x] Add test for updateContext iFrame based
- [x] Add test for compound scenario

Use existing test-app structure and avoid adding extra unnecessary files if not needed. 

https://docs.luigi-project.io/docs/luigi-container-api?section=updatecontext

--

This issue is part of https://github.com/SAP/luigi/issues/3757

walmazacn:
**Description**

- adds test for updateContext WC based
- adds test for updateContext iFrame based
- adds test for compound scenario

**Related issue(s)**

Resolves #3786


container/src/LuigiCompoundContainer.svelte
```diff
@@ -103,6 +103,9 @@
     thisComponent.updateContext = (contextObj: any, internal?: any) => {
       mainComponent._luigi_mfe_webcomponent.context = contextObj;
     };
+    thisComponent.getContext = (): any => {
+      return mainComponent?._luigi_mfe_webcomponent?.context || {};
+    };
```
ndricimrr, line 108: I would avoid introducing a new feature in this Test-only PR. This **getContext** function would then also have to be documented and shipped as a new feature. You can create a new Issue and ask if it makes sense to add it at this point. Thereafter you may have a separate PR for that. Since we havent added it till now, there probably were no requests so far. 

Another potential issue is clashing with this Light DOM Container PR [change here](https://github.com/SAP/luigi/pull/3697/files#diff-9ef147170f1d5c78725fc1fe4994ed87b810dde97fbc0f2b91ef886fe136db4eR114)- where mainComponent will depend on the **noShadow** option

This might work but there might be another way to grab the context. 
I believe since updateContext was made available for CompoundContainer, there should have already been a way to get the context. I am thinking some way through container.getAttribute('context') maybe or another form like **get/set context(ctx)** WC functions  . If there isnt a very simple way then you can skip adding this test for compound for now.

container/test-app/iframe/microfrontend.html
```diff
@@ -102,9 +102,14 @@ <h1 id="title">Multi purpose demo page</h1>
...
-          'LuigiClient.getContext()() called' + JSON.stringify(LuigiClient.getContext())
-        );
+        LuigiClient.sendCustomMessage({
+          id: 'my.contextMessage',
+          ...LuigiClient.getContext()
+        });
+        LuigiClient.uxManager().showAlert({
+          text: `LuigiClient.getContext() called ${JSON.stringify(this.ctx)}`,
+          type: 'info'
+        });
```
ndricimrr, line 112: I believe the sendCustomMessage for transmitting context back is enough. You can remove the showAlert

container/test-app/wc/clientAPI.html
```diff
@@ -43,6 +45,8 @@ <h3>
       import MFEventID from '../bundle.js';
 
       [...document.querySelectorAll('luigi-container')].forEach(luigiContainer => {
+        luigiContainer.init();
```
ndricimrr, line 48: This line might be excessive. The luigi container in this file would need `.init() `function to be initialised only if it had the **defer-init** attribute set to true. If this is needed regardless, then we might have a bug.

ndricimrr (commented):
Tests look good. 

Some other comments left

container/src/LuigiCompoundContainer.svelte
```diff
@@ -108,7 +108,17 @@
...
+        // handle multiple webcomponents on page
+        const wrappers = (thisComponent.getNoShadow() ? thisComponent : mainComponent)._luigi_mfe_webcomponent.querySelectorAll('div[style^="grid-row"]');
+        wrappers?.forEach((item) => {
+          if (item._luigi_mfe_webcomponent) {
+            item._luigi_mfe_webcomponent.context = contextObj;
+          }
+        });
+      } else {
+        (thisComponent.getNoShadow() ? thisComponent : mainComponent)._luigi_mfe_webcomponent.context = contextObj;
+      }
```
ndricimrr, line 121: was there a bug with updateContext not working on Compound Container? If it is, could you create a bug issue and handle it separately there. Then this Cypress Pr can be merged/reviewied easier

ndricimrr (commented):
Looks good. Please move the bugfix to a new PR if applicable

container/src/LuigiCompoundContainer.svelte
walmazacn: In this case '_luigi_mfe_webcomponent' points to direct child of MAIN element - which is DIV element wrapping compound item containers created through 'createCompoundItemContainer' method. Each of these containers have some WC for which context should be updated - instead of adding 'context' prop to the DIV wrapper. Check attached screenshot for reference

Zrzut ekranu 2024-07-02 105131

container/cypress/e2e/test-app/compound/wc-compound-container.cy.js
```diff
@@ -1,26 +1,29 @@
 describe('Compound Container Tests', () => {
   describe('LuigiClient API - LuigiCompoundContainer', () => {
+    const container = '[data-test-id="luigi-client-api-test-compound-01"]';
```
ndricimrr, line 3:
```suggestion
    const containerSelector = '[data-test-id="luigi-client-api-test-compound-01"]';
```

This name fits better I think

ndricimrr (approved):

PR #3790, opened 2024-06-21
Enables Prettier formatting in TS files

### LINKED ISSUE #3789: Investigate re-enabling prettier commit hook for Typescript files
Due to some linting and pipeline issues in the past we had to disable the TS linting from the prettier config. 
https://github.com/SAP/luigi/pull/3423

Investigate if we can turn this on again and re-enable it so that the Typescript files can be prettified automatically again.

Make sure it doesn't affect the Pipeline and that we do not need to do any major Version changes on the pipeline as that would require a lot of work. 

If there is an easy solution then re-enable it and apply all the Typescript prettifying in one quick PR to avoid merge conflicts. 

If its too much work and not worth it for now, feel free to close the issue.

walmazacn:
**Description**

- enabled Prettier formatting in TS files

**Related issue(s)**

Resolves #3789


JohannesDoberer (approved):
LGTM 👍 !

PR #3796, opened 2024-06-27
Left nav subcategories

hardl:
Note: this is a PR that contains code for a experimental feature, hence no docu or tests


core/src/navigation/LeftNav.svelte
```diff
@@ -273,12 +280,65 @@
     semiCollapsibleButton = LuigiConfig.getConfigValue('settings.responsiveNavigation') === 'semiCollapsible';
   });
 
+  function convertEntriesToToolLayout(entries) {
+    if (btpToolLayout) {
```
walmazacn, line 284: early return?

JohannesDoberer (approved):
Tested and works 👍

core/src/navigation/LeftNav.svelte
hardl: Since there are different opinions out there if early return is good or bad, I would leave it like it is ;)

Issue #2415, opened 2021-12-08
Move duplicated fn of GS to helpers


JohannesDoberer:
Currently we support GlobalSearch and experimental ft GlobalSearchCentered.
These two components have a lot of same functions.
Reduce redundancy and move it to helpers function and provide tests for it.
Follow-up of #2328

PR #3802, opened 2024-07-03
Move duplicated fn of gs to helpers

amilewskaa:
**Description**

- GlobalSearch and GlobalSearchCentered code refactor to reduce duplicated code
- Unit test for GlobalSearchHelperClass
- Ranamed variables: 
       `luigiCustomSearchRenderer__slot`-> `customSearchItemRendererSlot`
       `luigiCustomSearchRenderer__slotContainer`-> `customSearchItemRendererSlotContainer`

**Related issue(s)**
#2415


core/src/utilities/helpers/global-search-helpers.js
```diff
@@ -1,13 +1,195 @@
...
+  renderCustomSearchItem(item, slotContainer, index) {
+    setTimeout(() => {
+      search.searchProvider.customSearchResultItemRenderer(item, slotContainer.children[index], searchApiObj);
+    });
+    return '';
+  }
+
+  onKeyUp({ keyCode }, displaySearchResult) {
+    if (this.search && this.search.searchProvider) {
+      if (GenericHelpers.isFunction(this.search.searchProvider.onEnter) && keyCode === KEYCODE_ENTER) {
```
walmazacn, line 74: For multiple conditions `switch` statement seems to be a better choice

core/src/utilities/helpers/global-search-helpers.js
```diff
@@ -1,13 +1,195 @@
...
+    }
+  }
+
+  getSearchPlaceholder() {
+    const searchProvider = this.search.searchProvider;
+    if (!searchProvider || !searchProvider.inputPlaceholder) {
+      return undefined;
+    }
+    const currentLocale = LuigiI18N.getCurrentLocale();
+    if (GenericHelpers.isFunction(searchProvider.inputPlaceholder)) {
```
walmazacn, line 50: For multiple conditions switch statement seems to be a better choice

core/src/utilities/helpers/global-search-helpers.js
```diff
@@ -1,13 +1,195 @@
...
+
+  onSearchResultItemSelected(searchResultItem) {
+    if (this.search && GenericHelpers.isFunction(this.search.searchProvider.onSearchResultItemSelected)) {
+      this.search.searchProvider.onSearchResultItemSelected(searchResultItem);
+    } else if (GenericHelpers.isFunction(this.search.searchProvider.onEscape) && event.keyCode === KEYCODE_ESC) {
+      this.search.searchProvider.onEscape();
+    }
+  }
+
+  handleKeydown(result, { keyCode }, inputElement, luigiCustomSearchItemRenderer__slotContainer) {
```
walmazacn, line 140: Some alternative name for 'luigiCustomSearchItemRenderer__slotContainer' argument? Perhaps 'slotContainer' is good enough?

amilewskaa: To me original variable name is better, as it also provides the information what type of renderer is used.

core/src/utilities/helpers/global-search-helpers.js
amilewskaa: I can change that to switch, but I don't see much benefit from that.
@JohannesDoberer Please advise which approach do you prefer :)

core/src/utilities/helpers/global-search-helpers.js
amilewskaa: Same as: https://github.com/SAP/luigi/pull/3802#discussion_r1664636537

core/src/utilities/helpers/global-search-helpers.js
JohannesDoberer: I agree with both sites :)
Indeed this variable is a bit long and also the underscore is maybe outdated. On the other hand when you only read 'slotContainer‘ you don’t know what the variable is standing for.
What we definitely do not need is 'luigi' because it is obvious and just an internal variable and we can somehow get rid of the underscore.
If 'customSearchItemRendererSlotContainer' is a better name... i don't know either :)

core/src/utilities/helpers/global-search-helpers.js
```diff
@@ -1,13 +1,195 @@
...
+
+  renderCustomSearchItem(item, slotContainer, index) {
+    setTimeout(() => {
+      search.searchProvider.customSearchResultItemRenderer(item, slotContainer.children[index], searchApiObj);
+    });
+    return '';
+  }
+
+  onKeyUp({ keyCode }, displaySearchResult) {
+    if (this.search && this.search.searchProvider) {
```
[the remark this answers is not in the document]
JohannesDoberer: Since there are different opinions out there if early return is good or bad, I would leave it like it is ;)

core/src/utilities/helpers/global-search-helpers.js
walmazacn: OK - let's leave it as is. In the future we can also consider config object instead of multiple args to make it more robust :)

amilewskaa: Changed:
`luigiCustomSearchItemRenderer__slotContainer` -> `customSearchItemRendererSlotContainer`
`luigiCustomSearchItemRenderer__slot` - `customSearchItemRendererSlot`

core/src/utilities/helpers/global-search-helpers.js
walmazacn: OK - let's leave it as is

core/src/utilities/helpers/global-search-helpers.js
walmazacn: OK - let's leave it as is

walmazacn (approved):
LGTM 👍

JohannesDoberer (commented):
LGTM besides the small suggestion regarding camelCase

JohannesDoberer (approved):
LGTM 👍

PR #3806, opened 2024-07-04
missing hint for webcomponent prop for LuigiContainer

JohannesDoberer:
Added the note which is also described for webcomponent in luigi core http://localhost:4000/docs/navigation-parameters-reference?section=webcomponent


walmazacn (approved):
LGTM 👍

PR #3807, opened 2024-07-05
Adjust Container release NEXT_RELEASE string

### LINKED ISSUE #3804: Release script for classic luigi replaces nextrelease placeholder in container
When we run **./tools/release-cli/replaceInAllFiles.sh "NEXTRELEASE" "X.XX.X"** like described in the **./scripts/tools/release-cli/release-cli.js** script, the md files in container will also be touched.
This should not be the case because container has its own release cycle.

ndricimrr:
**Description**

- Changes container next release tag to NEXT_RELEASE_CONTAINER to avoid confusion 
**Related issue(s)**

Fixes #3804


walmazacn (approved):
LGTM 👍

PR #3841, opened 2024-08-01
Fix css issues for btptool layout

JohannesDoberer:
Can be checked with nightly version 2.14.3-dev.202408191043 in fiddle (for example)


walmazacn (approved):
Looks good 👍

PR #3844, opened 2024-08-05
Adds test for sandbox rules

### LINKED ISSUE #3842: Test sandboxRules for LuigiContainer
Add an e2e-test for LuigiContainer when an iframe should be displayed with custom sandboxRules.
```
<luigi-container viewUrl="/multipurpose.html" sandboxRules='["allow-scripts", "allow-same-origin"]'></luigi-container>
```
Test if the iframe in the dom has the specified sandboxRules

--

This issue is part of https://github.com/SAP/luigi/issues/3757

walmazacn:
**Description**

- adds test for sandbox rules in Luigi container

**Related issue(s)**

Resolves #3842


container/src/LuigiContainer.svelte
```diff
@@ -121,6 +121,10 @@
 
   const initialize = (thisComponent: any) => {
     if (!containerInitialized) {
+      if (!sandboxRules && thisComponent.getAttribute('sandboxRules')) {
+        sandboxRules = JSON.parse(thisComponent.getAttribute('sandboxRules'));
+      }
+
```
ndricimrr, line 127: I was expecting only added Tests in this PR. How come we need to add these lines. 

The **sandboxRules** feature was already added previously [here](https://github.com/SAP/luigi/pull/3650).

What do these lines help with? 

If this is a discovered bug, then a Bug PR for this would be more than welcome. Its much easier to look back at Git history when Reviewing in the future, but also helps spread the word that there was some issue with  this feature at some point in time.

container/test-app/iframe/microfrontend-sandbox-rules.html
```diff
@@ -0,0 +1,14 @@
+<!-- This microfrontend HTML file is used to test sandbox rules functionality -->
```
ndricimrr, line 1: I can see you have added this file  [microfrontend-sandbox-rules.html](https://github.com/SAP/luigi/pull/3844/files#diff-1bccb4094c8078d729a629df50f0d7a6f7f2bc1e2d9525c96ae05b6bbb55dbbd) and I assume that we might need to add more as needed in the future . 

It seems like we might be "running out of space" as shown in the picture below, as we keep adding more and more containers: 
Screenshot 2024-08-07 at 15 42 56

To keep things separated to avoid the previous clutter that we had to restructure, I would expect to have another category for such types of functionality. 
Screenshot 2024-08-07 at 15 44 11

So for example we can add an <Add new Category Here> where we put the sandbox rules test, defer init test, and maybe soon also the allowRules test. I don't know what we can call this category, but it would be nice to start separating now, then keep restructuring in the future.

container/cypress/e2e/test-app/iframe/iframe-settings.cy.js
```diff
@@ -0,0 +1,54 @@
...
+          const $body = iframe.contents().find('body');
+          cy.wrap($body)
+            .contains('defer-init test for iframes')
+            .should('exist');
+        });
+    });
+  });
+
+  it('set sandbox rules by property', () => {
+    cy.get('#init-button').click();
```
ndricimrr, line 30:
```suggestion
```

container/cypress/e2e/test-app/iframe/iframe-settings.cy.js
```diff
@@ -0,0 +1,54 @@
...
+      .get('iframe')
+      .then(elements => {
+        cy.get(elements.first())
+          .invoke('attr', 'sandbox')
+          .should('eq', 'allow-modals allow-popups');
+      });
+  });
+
+  it('set sandbox rules by attribute', () => {
+    cy.get('#init-button').click();
```
ndricimrr, line 44:
```suggestion
```

ndricimrr (commented):
im guessing just a typo.

container/cypress/e2e/test-app/iframe/iframe-settings.cy.js
walmazacn: Actually it's needed as I'm adding sandbox rules to this container - so it has to be initialized

container/cypress/e2e/test-app/iframe/iframe-settings.cy.js
walmazacn: In this case it's not needed - good catch :)

ndricimrr (approved):
Looks good

PR #3862, opened 2024-08-21
Changelog container

JohannesDoberer:
Make sure releasemeister docu is updated


JohannesDoberer:
lerna-changelog does not work. It does not support to run for a specific lib within a repo.
In addition lerna-changelog seems to be not actively maintained.
Decided to introduce a release-cli for container which updates the package.json and change the changelog.md file with the new changes.
Releasemeister docu is also updated.

JohannesDoberer:
UI5 support lib will be handled in a different issue #3867

container/CHANGELOG.md
```diff
@@ -0,0 +1,25 @@
+# Changelog
```
ndricimrr, line 1: This seems like a new separate file. I am guessing because we keep it separate for container. 

Would be nice to have a small comment here like in main [changelog](https://github.com/SAP/luigi/blob/main/CHANGELOG.md?plain=1#L3-L25)

container/CHANGELOG.md
```diff
@@ -0,0 +1,25 @@
...
+
```
ndricimrr, line 2:
```suggestion
[v1.2.0] (2024-07-25)
```

maybe we add this as well here

container/release-cli.js
```diff
@@ -0,0 +1,139 @@
...
+const github = new GitHub({
+  token: process.env.GITHUB_TOKEN
+});
+const repo = github.getRepo('SAP', 'luigi');
+
+const logWarning = str => console.log(color.yellow.bold(str));
+const logSuccess = str => console.log(color.green.bold(str));
+const logError = str => console.log(color.redBright.bold(str));
+
+async function getContainerReleases() {
```
ndricimrr, line 16: would be nice to have some small comment header on these functions. what the input and output looks like. easier to debug later 🤔

container/release-cli.js
```diff
@@ -0,0 +1,139 @@
...
+function updateVersionInPgkJson(version) {
+  packageJson.version = version;
+  fs.writeFileSync('./public/package.json', JSON.stringify(packageJson, null, 4));
+  logSuccess('Updated container/public/package.json');
+}
+
+/*
+  Update package.json and add changes to changelog
+*/
+async function prepareRelease() {
```
ndricimrr, line 60: could you split this function in sections. 2-3 sub functions or so if possible.

ndricimrr (commented):
small comments

container/CHANGELOG.md
JohannesDoberer: Indeed 👍 . Removed it in the last commit by accident.

ndricimrr (approved):
Tested and working well 👍

PR #3882, opened 2024-08-29
Add publishEvent documentation and tests

### LINKED ISSUE #3860: Document publishEvent functionality
We have a [publishEvent](https://github.com/SAP/luigi/blob/main/core/src/services/web-components.js#L70) functionality that works for WC based Mf (Core+Container), but it is not documented anywhere.

 Find out why its not documented, or if its docu is hidden somewhere, and how its intended to work, and document it (add tests if possible).

ndricimrr:
**Description**

- add documentation for publishEvent feature
- Add example for compound scenario
- add example for wc scenario

**Related issue(s)**

Fixes #3860

**Preview of docu section example:**

Screenshot 2024-09-30 at 12 14 59


client/luigi-element.d.ts
```diff
@@ -419,6 +419,69 @@ export interface LuigiClient {
...
+  /**
+   * Publish an event that can be listened to on the core side.
+   *
+   * Similar to LuigiClient.sendCustomMessage(...) but for WebComponent based microfrontends.
+   *
+   * @param event Custom Event to be published
+   * @returns
+   * @memberof LuigiClient
+   *
+   * @example
```
ndricimrr, line 431: need to have a discussion on where to put this example at.

ndricimrr: This would be clarified on #3903 and the publishEvent docu can then be put in LuigiClient API section with the proper label

JohannesDoberer (approved):
LGTM 👍

smahati (approved):
it's good as is, an alternative for publish event

client/luigi-client-wc-docu-mixin.js
```diff
@@ -0,0 +1,64 @@
+/**
+ * <!-- label-success: Web Component API -->
```
smahati, line 2: I don’t see syntax highlighting but that’s matching the rest of the code blocks. It is longer than most other examples but unless you could omit some declarations that are made in other parts of the guide, it is also fine. On capire the site generator has more plugins enabled which would be nice like highlighting only the relevant lines of code (for example: [link](https://cap.cloud.sap/docs/guides/using-services#mock-associations)) .

client/luigi-client-wc-docu-mixin.js
```diff
@@ -0,0 +1,64 @@
...
+ * // secondChild.js
+ * // Set the custom event name = sendInput
+ * this.LuigiClient.publishEvent(new CustomEvent('sendInput', { detail: 'My own message' }));
+ *
+ * // main.html
+ * myContainer.addEventListener('custom-message', event => {
+ *   console.log('My custom message from microfrontend', event.detail.data);
+ * }
+ *
+ * // *Note:* eventListeners.name must match CustomEvent name above
```
smahati, line 36: could we make this render as a note like in a >**Note** with the comment as the text and the config in a codeblock inside the Note?

client/luigi-client-wc-docu-mixin.js
```diff
@@ -0,0 +1,64 @@
...
+ * compoundConfig = {
+ *  ...
+ *  children: [
+ *   {
+ *     viewUrl: 'firstChild.js'
+ *     ...
+ *     eventListeners: [
+ *           {
+ *             source: 'input1',
+ *             name: 'sendInput',
```
smahati, line 47: We could highlight this line using a JS comment if it isn't a JSON file

smahati (changes requested):

client/luigi-client-wc-docu-mixin.js
ndricimrr: ah, I checked and I think the way its structured, it seems like all content in Examples: shows as a one-piece code example.

client/luigi-client-wc-docu-mixin.js
ndricimrr: I tried this but apparently the whole thing doesn't accept any extra styling.

smahati (approved):
yeah, I thought that would be bit of a stretch so thank you for trying :).

smahati:
A general Question, in the comments there are references to filenames Like wcWebcomponent.js, is it for illustration or is it a reference to another example/sample?

ndricimrr:
> A general Question, in the comments there are references to filenames Like wcWebcomponent.js, is it for illustration or is it a reference to another example/sample?

 wcComponent.js , firstChild.js , etc are just random file names I chose for the example as illustration.

docs/luigi-client-api.md
```diff
@@ -33,6 +33,72 @@ This document outlines the features provided by the Luigi Client API. It covers
 
 Use the functions and parameters to define the Lifecycle of listeners, navigation nodes, and Event data.
 
+#### publishEvent
+
+<!-- label-success: Web Component API -->
```
hardl, line 38: Shouldn't that be "Web Component API only"?

PR #3898, opened 2024-09-09
Configurable option to hide the X button in a modal

JohannesDoberer:


core/src/Modal.svelte
```diff
@@ -321,18 +321,20 @@
...
-          <div class="fd-bar__element">
-            <button
-              class="fd-button fd-button--transparent fd-button--compact"
-              on:click={() => dispatch('close', { activeDrawer: false })}
-              aria-label="close"
-              data-testid={settings.closebtn_data_testid && isModal ? settings.closebtn_data_testid : 'lui-modal-index-' + modalIndex}
-            >
-              <i class="sap-icon sap-icon--decline" />
-            </button>
+        {#if settings.hideXBtn !== true}
```
walmazacn, line 324: Shouldn't we add this option to the docs?

JohannesDoberer: I will do.

JohannesDoberer:
Need to be clarified with the technical guys from hyperspace portal.

PR #3902, opened 2024-09-10
Fixes wrong drawer position

### LINKED ISSUE #3899: Wrong position of drawer when btp-tool layout is active
Steps to reproduce:
Scenario 1:
1. Following fiddle-config:
```
Luigi.setConfig({
    navigation: {
        breadcrumbs: {
            clearBeforeRender: true,
      renderer: (el, items, clickHandler) => {
        el.classList.add('myBreadcrumb');
        let breadcrumbs = document.createElement('ol');
        breadcrumbs.setAttribute('style', 'top: 0;position: absolute;left: 0;');
        items.forEach((item, index) => {
          if (item.label) {
            let itemCmp = document.createElement('li');
            itemCmp.setAttribute('style', 'display:inline; margin: 0 10px;');
            itemCmp.setAttribute('data-testid', `breadcrumb_${item.label}_index${index}`);
            itemCmp.innerHTML = item.label;
            itemCmp._item = item;
            breadcrumbs.appendChild(itemCmp);
          }
        });
        breadcrumbs.addEventListener('click', event => {
          console.log('event detail', event);
          event.preventDefault();
          clickHandler(event.detail.item._item);
        });
        el.appendChild(breadcrumbs);
        return breadcrumbs;
      }
        },
            profile: {
                logout: {
                    label: "End session",
                    icon: "sys-cancel",
                },
            },
            validWebcomponentUrls:['.*?'],
            nodes: [
                { 
                pathSegment: 'home', 
                label: 'h', 
                hideFromNav: true, 
                children: [
                { 
                    pathSegment: 'overview', 
                    label: 'Overview', 
                    icon: 'home',
                    viewUrl: '/examples/microfrontends/multipurpose.html',
                    context: {
                        title: 'Welcome to Luigi Fiddle!',
                        content: 'Click on "Modify Config" at the bottom right and play around with your Luigi configuration'
                    }
                }
                ] 
            
            }],
        }, 
        routing: { 
            useHashRouting: true 
        }, 
        lifecycleHooks: {
            luigiAfterInit: () => {
                console.log('Luigi initialized.');
            }
        },
        settings: { 
            responsiveNavigation: 'semiCollapsible',
            btpToolLayout: true,
            experimental:{
                btpToolLayout: true,
            }
        },
    });
```
2. open a drawer e.g. in browser console run `Luigi.navigation().openAsDrawer('/home/overview')`
3. This leads to:

Scenario 2:
1. Same fiddle-config w/o breadcrumbs
2. open a drawer e.g. in browser console run `Luigi.navigation().openAsDrawer('/home/overview')`
3. This leads to:

Expected:
Scenario 1 align it with the content view
Scenario 2 align it with the content view

walmazacn:
**Description**

- fixes wrong drawer position

**Related issue(s)**

Resolves #3899


core/src/styles/_variables.scss
```diff
@@ -2,6 +2,7 @@ $mobileMinWidth: 375px;
 $desktopMinWidth: 600px;
 $desktopMaxWidth: 900px;
 $topNavHeight: var(--luigi__shellbar--height);
+$combinedLayoutGap: 1rem;
```
ndricimrr, line 5: Tested it and the drawer appears where it should be 👍 . 

I am bit confused why specifically **rem** here and why **1** . I am assuming as long as it puts the dialog bog in the right place should be okay, but thought maybe instead of a fixed number we could somehow calculate size/height of breadcrumb and account for it in there 🤔  I am not sure if [ --luigi__breadcrumb--height](https://github.com/SAP/luigi/blob/main/core/src/App.svelte#L2036) can be somehow used or not. 



(Very edge case and probably not used and we don't have to handle it necessarily, but: you can have **Breadcrumbs** **enabled** + **[showAsTabHeader](https://docs.luigi-project.io/docs/navigation-advanced?section=tabnav)= true** and the dialog I think would still need to be adjusted again. 🤔 )

walmazacn: It's '1rem' as it combines two layout gaps visible on the screen - it doesn't depend on the breadcrumbs at all:
- top padding value from the main wrapper ('0.5rem')
- gap between header and content elements ('0.5rem')

ndricimrr (approved):

PR #3920, opened 2024-09-18
Make clientAPI label distinguishable

### LINKED ISSUE #3903: Make Web Component API documentation label distinguishable
WebComponent LuigiClient API is usually in sync with Core. 

But not all functionality is fully in sync, such as **publishEvent** function which is only for WC.  This needs to be clear when looking at the documentation.

For example: **getActiveFeatureToggles** is available for both and it currently is labeled like this:
Screenshot 2024-09-10 at 16 11 18

Come up with a different naming for the "Web Component API" label to make it clear when its for both or only one of them. 
Maybe "Common" vs "Web Component only", or just put "Web Component only". Discuss the naming.

ndricimrr:
**Description**

**Changes proposed in this pull request:**
This PR adds a tooltip (title) to the labels that are added on the luigi client api documentation to give more clarification to the labels and avoid confusions

1. **Web App API only** --> This should represent all Luigi Client API functions (Classic ones) which you cant find for Luigi Container and WC-based Microfrontends in general. Please add suggestions if the phrasing is not right . 

Screenshot 2024-09-18 at 18 53 41

___________________________

2. **Web Component API ONLY** : --> This is aimed at all the LugiClient API functions that you would only find on Luigi Container clientAPI or WC-based microfrontends in general . Please add suggestions if the phrasing is not right . 
     
(it would look like this when #3882  is merged.)
Screenshot 2024-09-17 at 15 16 23

**Related issue(s)**

Fixes #3903

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -5,6 +5,7 @@ on:
 name: Backwards Compatibility Check
 env:
   CYPRESS_CACHE_FOLDER: ~/cypress-binary-cache
+  CYPRESS_DASHBOARD_RECORD_KEY: ${{ secrets.CYPRESS_DASHBOARD }}
   GH_TOKEN: ${{ github.token }}
 
 jobs:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,18 @@ The lerna-changelog tool detects changes based on PR labels and maps them to sec
 -->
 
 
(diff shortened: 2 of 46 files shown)


JohannesDoberer (commented):
Small comment.
Should we mark storageManager as well? As I can see it is not available for wc api.

ndricimrr:
@JohannesDoberer Added the changes

JohannesDoberer (approved):
LGTM 👍

PR #3922, opened 2024-09-19
Adds node property for hiding global search

### LINKED ISSUE #3921: Core: nav node property for hiding global search
Add a property to navigation node in configuration (proposal: 'hideGlobalSearch') that, when set to true, hides the global search in the shellbar if the perticular node is the current one (similar to 'hideSideNav' https://docs.luigi-project.io/docs/navigation-parameters-reference?section=hidesidenav).

walmazacn:
**Description**

- adds node property for hiding global search
- adjusts related documentation

**Related issue(s)**

Resolves #3921

### DIFF
--- a/core/src/services/routing.js
+++ b/core/src/services/routing.js
@@ -406,31 +406,15 @@ class RoutingClass {
       let cNode2 = currentNode;
       let hideGlobalSearchInherited = nodeObject.hideGlobalSearch;
       if (hideGlobalSearchInherited === undefined) {
-        while (cNode2) {
-          if (cNode2.tabNav && cNode2.hideGlobalSearch === true) {
-            hideGlobalSearchInherited = true;
-            break;
-          }
-          if (cNode2.hideGlobalSearch === false) {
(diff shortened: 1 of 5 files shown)


core/src/services/routing.js
```diff
@@ -404,6 +404,20 @@ class RoutingClass {
...
+            hideGlobalSearchInherited = true;
+            break;
+          }
+          if (cNode2.hideGlobalSearch === false) {
+            hideGlobalSearchInherited = false;
+            break;
+          }
+          cNode2 = NavigationHelpers.getParentNode(cNode2, pathData.navigationPath);
+        }
+      }
```
ndricimrr, line 420: Could we extract this logic in a function of its own and add some documentation to it with JSDoC. This function is super long and lets not have it become even longer 😄  . Also, if we plan to remove this feature in the future, it would be also much easier to remove, by just deleting the function thereafter.

ndricimrr (commented):
small comment

ndricimrr (approved):
Tested and working fine

PR #3923, opened 2024-09-19
HTML custom data support for Luigi(Compound)Container in VS Code

JohannesDoberer:


JohannesDoberer:
Currently it is not possible to automatically modify the users VS Code settings directly due to security reasons.
What we can do is a post-install script with a hint how to add this.

PR #3934, opened 2024-09-20
Adds node params to compound container typings

### LINKED ISSUE #3924: nodeParams are not documented in LuigiCompoundContainer.svelte.d.ts

walmazacn:
**Description**

- adds node params to compound container typings

**Related issue(s)**

Resolves #3924

### DIFF
--- a/container/typings/LuigiCompoundContainer.svelte.d.ts
+++ b/container/typings/LuigiCompoundContainer.svelte.d.ts
@@ -25,6 +25,12 @@ export default class LuigiCompoundContainer extends HTMLElement {
   deferInit: boolean;
 
   /**
+   * The parameters to be passed to the compound micro frontend.
+   * @since 1.4.0
+   */
+  nodeParams: Object;
+
+  /**
    * If set to true, the Luigi compound container webcomponent will not use the shadow DOM for rendering.
    * @since 1.2.0
    */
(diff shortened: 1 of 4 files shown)


container/typings/LuigiCompoundContainer.svelte.d.ts
```diff
@@ -24,6 +24,12 @@ export default class LuigiCompoundContainer extends HTMLElement {
    */
   deferInit: boolean;
 
+  /**
+   * The parameters to be passed to the compound micro frontend.
```
ndricimrr, line 28:
```suggestion
   * The parameters to be passed to the compound micro frontend. Will not be passed to the compound children.
```

Seems like this part is missing here. since we skip it for compound children. 

Also I saw that this sentence "Will not be passed to the compound children." is present on [LuigiContainer Typings : nodeParams ](https://github.com/SAP/luigi/blob/main/container/typings/LuigiContainer.svelte.d.ts#L81) . It should be removed from there and be kept here, since LuigiContainer doesn't deal with compound and that part there is wrong. You can do it in this PR maybe since its related

ndricimrr (commented):
Looks good. Small comments

ndricimrr (approved):
Looks good

Issue #3911, opened 2024-09-13
LeftNav in btpLayout: make 'expanded' the default state for nav item groups


hardl:
Categories with sub-categories or with property 'isGroup' set to true are rendered as nav groups. Unlike 'normal' categories, they should be expanded by default (if there is no explicit state set in localstorage). This might require the introduction of a dedicated localstorage key for them.

PR #3938, opened 2024-09-23
LeftNav in btpLayout: make 'expanded' the default state for nav item groups

amilewskaa:
**Description**

- make 'expanded' the default state for nav item groups (when they have sub-categories or `isGroup: true`
- store collapsed item keys in localStorage

**Related issue(s)**
#3911


core/src/utilities/helpers/navigation-helpers.js
```diff
@@ -322,6 +323,25 @@ class NavigationHelpersClass {
     return expandedList;
   }
 
+  storeCollapsedSuperCategoriesState(key, value) {
```
walmazacn, line 326: Can you add unit tests for these new methods? Please use existing 'navigation-helpers.spec.js' file for this purpose

amilewskaa: Added 😸

walmazacn (approved):
Looks good 👍

PR #3939, opened 2024-09-23
Adds skipInitCheck property to LuigiCompoundContainer

### LINKED ISSUE #3929: skipInitCheck property missing for LuigiCompoundContainer
We do not have a skipInitCheck property/attribute for LuigiCompoundContainer like for LuigiContainer

walmazacn:
**Description**

- adds skipInitCheck property to LuigiCompoundContainer
- adds related e2e test

**Related issue(s)**

Resolves #3929


container/src/services/webcomponents.service.ts
```diff
@@ -332,6 +332,9 @@ export class WebComponentService {
       getUserSettings: (): Object => {
         return this.thisComponent.userSettings || {};
       },
+      getSkipInitCheck: (): boolean => {
+        return !!this.thisComponent.skipInitCheck;
+      },
```
ndricimrr, line 337: I am not really sure if we need to have any `getSkipInitCheck` function. We don't have it for Container as well as Core WC. 
The `skip-init-check` is not a value that is set on the Container and we can fetch it from the inside. Its more a boolean that will affect how the component starts - with or without waiting for Init Check.

container/src/LuigiCompoundContainer.svelte
```diff
@@ -104,6 +106,7 @@
       nodeParams &&
       pathParams &&
       searchParams &&
+      skipInitCheck &&
```
ndricimrr, line 109: I think there needs to be some logic here to use the **skipInitCheck** set variable to "skip the init check" when **CompoundContainer** is first built. Similar to how **skipInitCheck** is used in **LuigiContainer**.

ndricimrr (commented):
Left some small comments

ndricimrr (approved):
Looks good

PR #3944, opened 2024-09-24
Adds nightly release task for container

### LINKED ISSUE #3932: Possibility to create nightly releases for container
See nightly release task for luigi core, client, plugins...etc

walmazacn:
**Description**

- adds nightly release task for container

**Related issue(s)**

Resolves #3932


scripts/tools/release-cli/release-cli.js
```diff
@@ -38,6 +38,7 @@ const pkgJsonPaths = {
   core: path.resolve(base, 'core', 'public', 'package.json'),
```
ndricimrr, line 38: I think there might be some issues with regards to the nextVersion generated for nightly Container.

So far nightlies have been done for the `core/client/etc` packages where we follow the `core/package.json` version as a basis + "-dev" + timestamp. 

Specficially [here](https://github.com/SAP/luigi/blob/d039ff2b364cc88ef25001763745bd5fc2cca00f/scripts/tools/release-cli/release-cli.js#L89-L91)
```
function getNextVersion() {
  return semver.inc(getVersion('core'), 'patch');
}
```

In this PR I think this would end up (due to [this line](https://github.com/SAP/luigi/blob/main/scripts/tools/release-cli/release-cli.js#L149)) publishing Container nightlies with 2.15.0-dev.2024.... It should instead look sth like `"1.3.0-dev.2024"`

ndricimrr: One solution might be 

https://github.com/SAP/luigi/blob/d039ff2b364cc88ef25001763745bd5fc2cca00f/scripts/tools/release-cli/release-cli.js#L180-L182

```
 for (const name of Object.keys(pkgJsonPaths)) {
   // check name or path

   if (name is related to container/package.json && Nightly = true) {
      const containerNightly = modify the version accordingly;
        writeVersion(pkgJsonPaths[name], <containerNightly here>);
    } else {
        writeVersion(pkgJsonPaths[name], input.version);
    }
  }
```

scripts/tools/release-cli/release-cli.js
```diff
@@ -90,6 +92,22 @@ function getNextVersion() {
   return semver.inc(getVersion('core'), 'patch');
 }
 
+function getVersionSuffix() {
```
ndricimrr, line 95: good idea extracting this into a function 👍 

Could you add some JSDoc on the function header. Would be helpful to know what it returns and what the returned value format looks like without having to read the whole code.

scripts/tools/release-cli/release-cli.js
```diff
@@ -178,7 +186,16 @@ function addToChangelog(versionText, changelog, lastline) {
    * PACKAGE VERSIONS
    */
   for (const name of Object.keys(pkgJsonPaths)) {
-    writeVersion(pkgJsonPaths[name], input.version);
+    let inputVersion = input.version;
+
+    if (name === 'container' && process.env.NIGHTLY === 'true') {
```
ndricimrr, line 191:
```suggestion
    // handle custom container version for nightly release 
    if (name === 'container' && process.env.NIGHTLY === 'true') {
```

ndricimrr, line 191: On second thought this works pretty well, but I think we also have to think that the same script **release-cli.js** is used for non-Nightly normal releases. 
So in that case **process.env.NIGHTLY** is not true, and that means 

the following will run 
` writeVersion(pkgJsonPaths[name], inputVersion);`
 
 with 
`writeVersion(pkgJsonPaths["container"], <Whatever Version Core has>);` and maybe end up publishing a Container release with a Core version. 

but Core Version !== Container version. 


So to alleviate this issue we can add on lines 41 and 67 above something like: 
```
// only add container to paths if nightly is present. Otherwise container path not needed
if (process.env.NIGHTLY is enabled) {
      pkgJsonPaths.container = path.resolve(base, 'container', 'public', 'package.json')
      installPaths.container = path.resolve(base, 'container'),
}
```

scripts/tools/release-cli/replaceInAllFiles.sh
```diff
@@ -24,6 +24,7 @@ SOURCES=(
   "core/src"
   "client/src"
   "client/luigi-client.d.ts"
+  "container/src"
```
ndricimrr, line 27: I think for container the replacement is done separately on `container/prepareNextRelease.js` [here](https://github.com/SAP/luigi/blob/main/container/prepareNextRelease.js). I am thinking this might not be needed 🤔

ndricimrr (commented):
Left some comments

ndricimrr (commented):
Seems like 

`process.env.NIGHTLY === true `

needs to be 

`process.env.NIGHTLY === 'true' `as env variables seem to be taken as strings instead. 

Two other old cases in this file where **NIGHTLY** is being checked as **boolean**, so you may change those as well. They didn't do any harm till now, but was outputting wrong logs.

scripts/tools/release-cli/release-cli.js
```diff
@@ -68,6 +67,11 @@ const installPaths = {
   testing_utilities: path.resolve(base, 'client-frameworks-support', 'testing-utilities')
 };
 
+if (process.env.NIGHTLY === 'true') {
```
ndricimrr, line 70: another (hopefully :D ) last edge case that I could think of is, if we run an [RC Release](https://github.com/SAP/luigi/blob/main/.github/workflows/rc-release.yml) for Core/Client/etc :

RC Releases are a modified version of a nightly, but released not nightly but [on request](https://github.com/SAP/luigi/actions/workflows/rc-release.yml), that we do only for Core/Client/etc, **not** Container.  

For RC Releases **process.env.NIGHTLY** needs to be '**true**' but also **process.env.NIGHTLY_VERSION** needs to be set with a (string) version , so only checking `NIGHTLY === 'true' ` might result in releasing an RC Release for Container without intending to do so. Container RC Releases are done manually on a separate process. 

So to fix that issue I think we can do:
`if (process.env.NIGHTLY === 'true' &&  process.env.NIGHTLY_VERSION is NOT set  )`

And I think we can add the same **if check** on **scripts/tools/publish-nightly.js**, just to be on the safe side, so we don't let it do RC release on Container. (For : publishPaths, packagePaths)

ndricimrr (approved):
I can't think of any other edge cases. Fine for me

scripts/tools/release-cli/release-cli.js
```diff
@@ -178,7 +192,17 @@ function addToChangelog(versionText, changelog, lastline) {
...
   for (const name of Object.keys(pkgJsonPaths)) {
-    writeVersion(pkgJsonPaths[name], input.version);
+    let inputVersion = input.version;
+
+    // handle custom container version for nightly release
+    if (name === 'container' && process.env.NIGHTLY === 'true') {
+      const containerNightlyVersion = getVersion('container');
+      const versionSuffix = getVersionSuffix();
+
+      inputVersion = containerNightlyVersion + versionSuffix;
```
ndricimrr, line 202:
```suggestion
      inputVersion = containerNightlyVersion + versionSuffix;
      logHeadline('\Container updated to v' + input.version + ':');
```

scripts/tools/publish-nightly.js
```diff
@@ -51,6 +51,11 @@ const publishPaths = {
   testing_utilities: ['client-frameworks-support', 'testing-utilities', 'dist']
 };
 
+if (process.env.NIGHTLY === 'true' && !process.env.NIGHTLY_VERSION) {
```
ndricimrr, line 54:
```suggestion
// Add container folder to publish in the nightly. Skip it for rc-release since container rc is done separately. 
if (!process.env.NIGHTLY_VERSION) {
  logHeadline('Added container to nightly release');
```
I think this should work better.

ndricimrr (commented):
last comments (hopefully 🤞 )

ndricimrr:
Container nightly release was [successful](https://www.npmjs.com/package/@luigi-project/container/v/1.3.0-dev.20241020912) from this branch. 

Run went well 
https://github.com/SAP/luigi/actions/runs/11141008661/job/30961168990

PR #3946, opened 2024-09-24
Adds option to disable luigiCookie

### LINKED ISSUE #3942: removal of luigiCookie
**Description**

I am getting chrome warnings:

```
Reading cookie in cross-site context may be impacted on Chrome 

Cookies with the SameSite=None; Secure and not Partitioned attributes that operate in cross-site contexts are third-party cookies. Chrome is moving towards a new experience that lets people make an informed choice with respect to third-party cookies.
```

and the affected cookie is your `luigiCookie`.

What are your plans with this issue?

Is it possible to remove this cookie and thus get rid of these problems?

**Reasons**

I am not sure how it would function in future chrome versions,
possibly chrome will show an accept cookie popup to users (?),
but it would be better to avoid this complication.

walmazacn:
**Description**

- adds option to disable luigiCookie

**Related issue(s)**

Resolves #3942


client/src/lifecycleManager.js
```diff
@@ -135,10 +136,13 @@ class LifecycleManager extends LuigiClientBase {
       },
       '*'
     );
-    this._tpcCheck();
   }
 
   _tpcCheck() {
+    if (this.currentContext?.internal?.thirdPartyCookieCheck?.disabled) {
```
ndricimrr, line 142: Do we need to send "luigi.tpcDisabled" message 🤔 . 

I am guessing this postMessage produces this warning on the other side:
`console.warn('Third party cookies are not supported! Silent token renewal might not work!');`

Maybe if they specifically set `disabled = true` they would know that tpc is disabled already and maybe warning would be redundant 🤔 . But lets discuss it tomorrow, not sure.

container/src/services/container.service.ts
```diff
@@ -106,7 +106,11 @@ export class ContainerService {
                   {
                     msg: LuigiInternalMessageID.SEND_CONTEXT_HANDSHAKE,
                     context: targetCnt.context || {},
-                    internal: {},
+                    internal: {
+                      thirdPartyCookieCheck: {
+                        disabled: !!(targetCnt.getAttribute && targetCnt.getAttribute('skip-cookie-check') === 'true')
```
ndricimrr, line 111: What about : ` disabled: !!targetCnt.skipCookieCheck`. I am guessing that might work as well, and shorter.

client/src/lifecycleManager.js
ndricimrr: I think another argument would be that when using an iframe based LuigiContainer, one would need to have a specific message listener for `'luigi.tpcDisabled'` on the host app. 

But before the message even gets to the Host App, the internal LuigiContainer ( somewhere in the getContainerManger ) would need to grab incoming iframe `luigi.tpcDisabled` messages and propagate them up to the host (parent), like we do for most other incoming iframe post messages.

ndricimrr (commented):
Left some comments, but mainly discussion points for daily.

container/src/services/container.service.ts
walmazacn: For some reason 'targetCnt.skipCookieCheck' always returns 'true' - not sure why

client/src/lifecycleManager.js
walmazacn: I can agree 'luigi.tpcDisabled' (same like 'luigi.third-party-cookie') message is used to handle errors configured in 'thirdPartyCookieCheck' settings - so most likely we can skip it here

container/src/services/container.service.ts
```diff
@@ -106,7 +106,11 @@ export class ContainerService {
                   {
                     msg: LuigiInternalMessageID.SEND_CONTEXT_HANDSHAKE,
                     context: targetCnt.context || {},
-                    internal: {},
+                    internal: {
+                      thirdPartyCookieCheck: {
+                        disabled: this.handleCookieCheck(targetCnt)
```
ndricimrr, line 111: I set this line to `disabled: targetCnt.skipCookieCheck` and bundled this and worked well. I think you can take a look at considering removing the need for the **handleCookieCheck** function introduced here. 

Printing out **targetCnt.skipCookieCheck** also results in **true** for me.

Did you make sure that here:
https://github.com/SAP/luigi/blob/bd332884f4d0362d0a800440c0ec517974b03d17/container/test-app/iframe/iframeContainer.html#L18-L23
you set `skip-cookie-check='true'` above in this file when testing, because it works for me. 

If `<luigi-container skip-cookie-check='true'></luigi-container>` doesn't set **skipCookieCheck** prop to **true**, then this needs to be analyzed properly, since [we rely on](https://github.com/SAP/luigi/blob/bd332884f4d0362d0a800440c0ec517974b03d17/container/src/LuigiContainer.svelte#L5-L27) the attribute to props conversion heavily all across the container code. 



Also, I noticed that our container iframe example is relying on some live Fiddle version and not the local **luigi-client** bundled lib: 

https://github.com/SAP/luigi/blob/bd332884f4d0362d0a800440c0ec517974b03d17/container/test-app/iframe/microfrontend.html#L7 

So that means your changes in `lifecycleManager.js` never get executed here. Maybe worth looking at, because I tried to manually copy the local built luigi-client and your code changes then work correctly.

container/test/services/container.service.spec.ts
```diff
@@ -68,7 +68,7 @@ describe('getContainerManager messageListener', () => {
     gtcSpy.mockRestore();
   });
 
-  it('test get context message', () => {    
+  it('test get context message', () => {
```
ndricimrr, line 71: The unit test I think is a good addition, but I think we can also add a small Cypress test
Our `iframe/iframeContainer.html` already produces a "cookies" alert automatically when you first open the example. apparently we left it there somehow:
Screenshot 2024-09-27 at 18 26 50

You can just check if that alert is present or not to make a test for it or whatever works easiest. 
I tested with your code and this alert doesnt appear when I set **skip-cookie-check** to **true** as expected, but I had to manually get the local script **src**= **"./luigi-client.js"** that I copied manually from **client/public**: 
https://github.com/SAP/luigi/blob/bd332884f4d0362d0a800440c0ec517974b03d17/container/test-app/iframe/microfrontend.html#L7

ndricimrr (commented):
Works as expected, but have some comments on Container code, that could be worth looking at.

docs/general-settings.md
```diff
@@ -298,15 +299,19 @@ This function is called with the following parameters:
 
 ## Third-party cookies support check
 
-You can check whether the user's browser supports third-party cookies by defining a **thirdPartyCookieCheck** object which expects a function called **thirdPartyCookieErrorHandling** and an optional **thirdPartyCookiesScriptLocation** parameter. When **thirdPartyCookiesScriptLocation** is set, the Luigi Core application checks third-party cookie support only once and not on every micro frontend call. If it is *not* set, the Luigi Core application checks third-party cookie support whenever a micro frontend is loaded.
+You can check whether the user's browser supports third-party cookies by defining a **thirdPartyCookieCheck** object which expects a function called **thirdPartyCookieErrorHandling** and optional **disabled** and **thirdPartyCookiesScriptLocation** parameters. When **thirdPartyCookiesScriptLocation** is set, the Luigi Core application checks third-party cookie support only once and not on every micro frontend call. If it is *not* set, the Luigi Core application checks third-party cookie support whenever a micro frontend is loaded.
```
ndricimrr, line 302:
```suggestion
You can check whether the user's browser supports third-party cookies by defining a **thirdPartyCookieCheck** object which expects a function called **thirdPartyCookieErrorHandling**, optional **disabled** and **thirdPartyCookiesScriptLocation** parameters. When **thirdPartyCookiesScriptLocation** is set, the Luigi Core application checks third-party cookie support only once and not on every micro frontend call. If it is *not* set, the Luigi Core application checks third-party cookie support whenever a micro frontend is loaded.
```

container/test-app/iframe/iframe-cookies.html
```diff
@@ -0,0 +1,39 @@
+<html>
+  <head>
+    <script type="module">
+      import '../bundle.js';
+    </script>
+  </head>
+  <body>
+    <h3>
+      This page is used to test Container Luigi CLient API functionalities for iframe
+      based microfrontend
```
ndricimrr, line 10:
```suggestion
      This page is used to test  **skip-cookie-check** feature for iFrame based LuigiContainer
```

ndricimrr (commented):
Looks good. Only some small comments

ndricimrr (approved):
Tested it and looks good.

PR #3948, opened 2024-09-25
release pkce plugin

JohannesDoberer:


walmazacn (approved):
Looks good 👍

Btw, after merging this you might create a follow-up task to uncomment the following line:
https://github.com/SAP/luigi/blob/main/website/fiddle/webpack.config.js#L58

PR #3956, opened 2024-09-27
Updates rollup version

### LINKED ISSUE #3953: Update rollup version in repo
where possible, provided it doesnt brake anything

#3955 
#3943
#3952 
and other cases.

walmazacn:
**Description**

- updates rollup version

**Related issue(s)**

Resolves #3953 | See also related issues:
- https://github.com/SAP/luigi/pull/3918
- https://github.com/SAP/luigi/pull/3943
- https://github.com/SAP/luigi/pull/3952
- https://github.com/SAP/luigi/pull/3955
- https://github.com/SAP/luigi/pull/3966


container/package.json
```diff
@@ -62,6 +62,9 @@
     "cli-color": "^2.0.4",
     "github-api": "^3.4.0"
   },
+  "optionalDependencies": {
+    "@rollup/rollup-linux-x64-gnu": "*"
```
ndricimrr, line 66: what is optionalDep used for 🤔  which other one needs it

ndricimrr (commented):
I think we can use this PR to update rollup wherever possible that includes the following:

- client-frameworks-support/client-support-angular
- core
- test/e2e-test-application
- website/docs
- core/examples/luigi-example-angular
- core/examples/luigi-example-vue
- core/examples/luigi-example-angular

I think **rollup** is used there but as a "child" dependency of another dependency. You can find these if you search "rollup" and it will show up in **package-lock.json**s  that you can use to find out which parent dependency has it as a child and update that. I think this issue is quite popular and many other packages might have already updated rollup internally. 

Example: `@angular-devkit/build-angular` depends on  `"rollup": "4.20.0",` on our `luigi-example-angular` and it was already bumped on their side [here](https://github.com/angular/angular-cli/pull/28480)


In case some pacakge that we use hasn't updated their rollup dep then you can note it down in the PR description. (Maybe we can also fix it with "overrides")

container/package.json
walmazacn: Without it bundle job for container was failing:
https://github.com/SAP/luigi/actions/runs/11071425700/job/30763262753

walmazacn:
I updated what's possible without breaking changes. The following stuff cannot be updated easily:
- 'core'
- 'website\docs'
- 'test/e2e-test-application'
- 'client-frameworks-support\client-support-angular'

Btw, it seems rollup in 'core/examples' is already safe enough (version 4+)

ndricimrr:
Looking at the reports following ranges are affected for rollup, so its not just older versions. v4 too : 
```
< 2.79.2           , patch = 2.79.2
>= 3.0.0, < 3.29.5 , patch = 3.29.5
>= 4.0.0, < 4.22.4 , patch = 4.22.5
```

So I think we should update the core examples as well if they are in the range and are not too much of a hassle to update.

container/package.json
```diff
@@ -48,11 +49,10 @@
     "eslint-plugin-svelte": "^2.33.0",
     "jest": "29.7.0",
     "jest-environment-jsdom": "29.7.0",
-    "rollup": "2.78.0",
-    "rollup-plugin-copy": "^3.4.0",
+    "rollup": "4.22.5",
```
ndricimrr, line 52: The bundle process on **container** seems to fail with the error shown in the end.

I am aware that we can somehow delete **npm_modules** folder and try again, but I think that is a bit invasive, since there is still an[ open issue](https://github.com/npm/cli/issues/4828) that is not resolved around this, so the current solution is more of a workaround . Since the container build is quite critical, I would propose to instead update to closest patch to v2, that fixes the warning mentioned and doesnt affect build process, and has no workarounds.

 We can than think later about migrating to a later version when they fix their own issues. 

```
Error when trying to bundle packages in container 
Stopping execution of the process due to error: Error: Command failed: cd container && npm run bundle
 /luigi/container/node_modules/rollup/dist/native.js:59
		throw new Error(
		^

Error: Cannot find module @rollup/rollup-darwin-x64. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
```

ndricimrr (commented):
One comment about the container

core/examples/luigi-example-angular/package.json
```diff
@@ -36,6 +36,7 @@
     "@types/node": "8.10.56",
     "glob-parent": "6.0.2",
     "lodash": "4.17.21",
+    "rollup": "4.22.5",
```
ndricimrr, line 39:
```suggestion
```
I don't think you need this. rollup here was a child dependency of `@angular-devkit/build-angular`, which you already update to latest version above. That contains the updated rollup inside.

core/package.json
```diff
@@ -19,6 +19,7 @@
     "lodash.template": "^4.5.0",
     "mock-local-storage": "^1.1.23",
     "node-fetch": "2.6.7",
+    "rollup": "^3.29.5",
```
ndricimrr, line 22: I think we can remove this rollup line and skip core update altogether. 

I would have to guess that, the Dependabot alert is **just outdated** for core regarding this rollup issue.

The warning most likely arised because down below here we have
  `"vite": "^4.0.0"` 
which at the time the warning was produced installed latest v4 vite = **4.5.3** due to the **^** sign

Currently `"vite": "^4.0.0"`  will install v4 vite = **4.5.5** as it was released recently as a patch to this issue. 


So the only thing to do here would be to have a **clean install** of depdencies and push the resulting **package-lock.json** 🤔

ndricimrr: To reproduce/test this you may try 
1. Fix vite in this file to   `"vite": "4.5.3"`
2. run **npm install**  (I usually delete node_modules and pacakge-lock before to make sure no cache is reused)
3. Check the output how many warnings it shows. (You may use **npm audit** to find the Dom Clobbering one)
4. Now change to `"vite": "^4.0.0"` and run a **clean npm install** again. You can see the dom clobering one is gone.

ndricimrr (approved):
Built and ran the packages locally to make sure. And it all looks good.

PR #3964, opened 2024-09-30
Clean Luigi store

JohannesDoberer:
We do not clear the subscriptions that we store in Luigi store. That leads to multiple initialization of route handling.

How to reproduce:
Browser console:
```
var cfg = Luigi.getConfig();
Luigi.unload();
Luigi.setConfig(cfg);
Luigi.unload();
Luigi.setConfig(cfg);
```

Result: 
=> navigate: luigi-client did not respond, using fallback by replacing iframe

Root cause:
How to check the multiple initialization of route handling:
1. add a console log [here](https://github.com/SAP/luigi/blob/main/core/src/App.svelte#L366) console.log('enableRouting')
2. start simpledev
3. browser console 
```
var cfg = Luigi.getConfig();
Luigi.unload();
Luigi.setConfig(cfg);
Luigi.unload();
Luigi.setConfig(cfg);

// when you do this more often you will see an increase of the console logs 'enableRouting'
Luigi.unload()
Luigi.setConfig(cfg);
Luigi.unload()
Luigi.setConfig(cfg);
....
```

### DIFF
--- a/.github/workflows/checkmarx.yml
+++ b/.github/workflows/checkmarx.yml
@@ -1,0 +1,17 @@
+name: Checkmarx One Scan
+on: workflow_dispatch
+jobs:
+  trigger_scan:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Checkmarx One CLI Action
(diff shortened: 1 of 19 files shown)


walmazacn (approved):
Clearing Luigi store works as expected - small unit test would be helpful

PR #3968, opened 2024-10-01
Fix fdToolLayout background-color

JohannesDoberer:
Wrong background-color in horizon-evening theme when you apply horizon evening (and don't use fundamental-ngx).


walmazacn (approved):
Looks good

PR #3971, opened 2024-10-02
Adds e2e tests for pathExists and showConfirmationModal

### LINKED ISSUE #3959: Add Container Cypress test for pathExists and showConfirmationModal
Check if there are any present Cypress tests for this already, and if not, add Cypress test for them.

You may check the implementation to find out how it works: 
[pathExists](https://github.com/SAP/luigi/blob/main/container/src/services/webcomponents.service.ts#L184-L200) and [showConfirmationModal](https://github.com/SAP/luigi/blob/main/container/src/services/webcomponents.service.ts#L226-L242) .

But it shouldn't be more complicated than having the test click on a button attached to the function call, and then checking the correct alert/log result is shown. 

Consider adding it for wc-based luigi container tests and luigi compound tests both **if** possible. 

--

This issue is part of https://github.com/SAP/luigi/issues/3757 epic

walmazacn:
**Description**

- adds e2e tests for pathExists and showConfirmationModal in container

**Related issue(s)**

Resolves #3959


ndricimrr (approved):
Tests look good 👍

PR #3983, opened 2024-10-07
Disable tpc via client

JohannesDoberer:
When LuigiClient functionality will be tested with testing-utilities lib the option to disable 3rd party cookie check with core has no effect. This PR introduces a possibility to disable 3rd party cookie check via LuigiClient.


walmazacn (approved):
Looks good

walmazacn (approved):
LGTM 👍

PR #3985, opened 2024-10-09
Add custom elements manifest for Luigi(Compound)Container

JohannesDoberer:


JohannesDoberer:
We decided to skip the attribute section because it's not relevant for the ui5 wrapper

walmazacn (approved):
I checked the manifest against related [schema](https://github.com/webcomponents/custom-elements-manifest/blob/main/schema.d.ts) and it seems to be correct
Anyway, we should think of adding an option to generate this file automatically in the future

PR #4008, opened 2024-10-31
Adds document attribute for disable tcpcheck

### LINKED ISSUE #4006: Client: use document.head attribute as single-source-of-truth for disable tcpCheck
instead of setting the member variable this.disableTpcCheck here https://github.com/SAP/luigi/blob/ed8eae559b78c46d8c59c3ff8a6ccd207c2d411d/client/src/lifecycleManager.js#L234C10-L234C26, set an attribute at document.head and read its value here https://github.com/SAP/luigi/blob/ed8eae559b78c46d8c59c3ff8a6ccd207c2d411d/client/src/lifecycleManager.js#L143

This allows disabling tcpCheck not only via luigi client but also declarative in mfe html.

walmazacn:
**Description**

- adds document attribute for disabling 3rd party cookie check
- removes `disableTpcCheck` prop to keep one source of truth

**Related issue(s)**

Resolves #4006


JohannesDoberer (approved):
LGTM 👍

client/luigi-client.d.ts
```diff
@@ -594,14 +594,10 @@ export declare interface StorageManager {
 /**
  * Registers a listener called with the context object and the Luigi Core domain as soon as Luigi is instantiated. Defer your application bootstrap if you depend on authentication data coming from Luigi.
  * @param {Lifecycle~initListenerCallback} initFn the function that is called once Luigi is initialized, receives current context and origin as parameters
- * @param {boolean} disableTpcCheck if set to `true` third party cookie check will be disabled via LuigiClient.
  * @memberof Lifecycle
  */
-export function addInitListener(initFn: (context: Context, origin?: string) => void, disableTpcCheck?: boolean): number;
-export type addInitListener = (
-  initFn: (context: Context, origin?: string) => void,
-  disableTpcCheck?: boolean
```
hardl, line 603: please keep the disableTcpCheck argument, it should still be possible to disable via js, e.g. in a test environment.

client/src/lifecycleManager.js
```diff
@@ -225,13 +237,11 @@ class LifecycleManager extends LuigiClientBase {
...
    * Registers a listener called with the context object and the Luigi Core domain as soon as Luigi is instantiated. Defer your application bootstrap if you depend on authentication data coming from Luigi.
    * @param {Lifecycle~initListenerCallback} initFn the function that is called once Luigi is initialized, receives current context and origin as parameters
-   * @param {boolean} disableTpcCheck if set to `true` third party cookie check will be disabled via LuigiClient.
    * @memberof Lifecycle
    * @example
    * const initListenerId = LuigiClient.addInitListener((context) => storeContextToMF(context))
    */
-  addInitListener(initFn, disableTpcCheck) {
-    this.disableTpcCheck = disableTpcCheck;
+  addInitListener(initFn) {
```
hardl, line 244: please keep the disableTcpCheck argument, it should still be possible to disable via js, e.g. in a test environment.
if it is true, just call document.head.setAttribute('disable-tpc-check') here instead of this.disableTpcCheck=true

client/src/luigi-client.js
```diff
@@ -26,8 +26,8 @@ class LuigiClient {
     }
   }
 
-  addInitListener(initFn, disableTpcCheck) {
-    return lifecycleManager.addInitListener(initFn, disableTpcCheck);
+  addInitListener(initFn) {
```
hardl, line 29: see above

hardl (changes requested):

client/src/lifecycleManager.js
```diff
@@ -140,17 +148,21 @@ class LifecycleManager extends LuigiClientBase {
   }
 
   _tpcCheck() {
-    if (this.currentContext?.internal?.thirdPartyCookieCheck?.disabled || this.disableTpcCheck) {
+    const tpcCheckDisabled =
+      this._isTpcCheckDisabled() || this.currentContext?.internal?.thirdPartyCookieCheck?.disabled;
```
hardl, line 152: moving the check
`this.currentContext?.internal?.thirdPartyCookieCheck?.disabled;`
into `_isTpcCheckDisabled()` could save one LOC ;)

hardl (approved):

PR #4028, opened 2024-11-19
Auth exp check reset

### LINKED ISSUE #4007: Option to reset expiration(-soon) check interval
in oauth2 plugin, when the onAuthExpireSoon event is fired, the check interval is cleared without an easy option to set it again, because the initial assumption was that a full auth redirect flow will eventually take place and thus the whole app state is reset anyway. There are now customer use cases where a new token can be obtained while staying at the page, which makes it necessary to establish new check intervals.

hardl:
fixes #4007


walmazacn (approved):
LGTM (also unit tests) 👍

JohannesDoberer (approved):

PR #4057, opened 2024-12-09
Adds test that checks that Core and Container WC clientAPI are the same

### LINKED ISSUE #3663: Add test that checks that Core and Container WC clientAPI are the same
Follow up from #3643 

To be done after Container Cypress Integration tests are restructured properly in #3562 

### Possible solution: 
This can work as a Cypress Integration test. 

Have two apps that use WC microfrontend with the same exact UI, but one of them is built as a Core based App, the other as a Container based only WC microfrontend. Maybe just put these under two different routes if possible.
 
At the end have one singular Cypress test file/s for both apps. 

The same tests should be ran twice but on different app routes. If test passes twice it means both Core and Client have the same clientAPI functionality. 

(Keep in mind that there might be exceptions where some Core WC clientAPI functionality might not make sense in Container scenario)

walmazacn:
**Description**

- adds test that checks that Core and Container WC clientAPI are the same

**Related issue(s)**

Resolves #3663


test/e2e-client-api-test-app/cypress/e2e/container-wc.cy.js
```diff
@@ -0,0 +1,505 @@
...
+    it('getCurrentLocale, getActiveFeatureToggles, getCurrentTheme', () => {
+      const alertMessages = [
+        'LuigiClient.getCurrentLocale()=en',
+        'LuigiClient.getActiveFeatureToggles()=,0',
+        'LuigiClient.uxManager().getCurrentTheme()=false'
+      ];
+
+      cy.get(containerSelector)
+        .shadow()
+        .contains('Click me')
```
JohannesDoberer, line 20:
```suggestion
        .contains('getCurrentLocale(), getActiveFeatureToggles, getCurrentTheme()')
```

test/e2e-client-api-test-app/cypress/e2e/container-wc.cy.js
```diff
@@ -0,0 +1,505 @@
...
+      cy.visit('http://127.0.0.1:3000/views/container/wc.html');
+      stub = cy.stub();
+    });
+
+    it('getCurrentLocale, getActiveFeatureToggles, getCurrentTheme', () => {
+      cy.on('window:alert', stub);
+
+      cy.get(containerSelector)
+        .shadow()
+        .contains('Click me')
```
JohannesDoberer, line 282:
```suggestion
        .contains('getCurrentLocale(), getActiveFeatureToggles, getCurrentTheme()')
```

JohannesDoberer (approved):
LGTM 👍

PR #4058, opened 2024-12-11
Wrong order of productswitcher item

JohannesDoberer:
Elements inside product switcher item have wrong order when `addNavHrefs:true` in navigation settings + space was not correct due to missing fd class


walmazacn (approved):
LGTM 👍

PR #4065, opened 2024-12-13
BtpLayout: Backdrop fix

JohannesDoberer:


walmazacn (approved):
I can confirm both 'addBackdrop' and 'removeBackdrop' methods work as expected 👍

PR #4106, opened 2025-01-09
dismissKey optional in closeAlert

JohannesDoberer:


container/src/services/container.service.ts
```diff
@@ -24,7 +24,7 @@ export class ContainerService {
   sendCustomMessageToIframe(iframeHandle: IframeHandle, msg: object, msgName?: string) {
     const messageName = msgName || 'custom';
 
-    if (iframeHandle.iframe.contentWindow) {
+    if (iframeHandle?.iframe?.contentWindow) {
```
walmazacn, line 27: Good catch!

PR #4115, opened 2025-01-16
showAlert promise for webcomponents

JohannesDoberer:


walmazacn (approved):
Both feature and tests work as expected

container/src/LuigiContainer.svelte
```diff
@@ -145,11 +145,14 @@
...
       };
 
-      thisComponent.closeAlert = (id: string, dismissKey: string) => {
-        if(webcomponent){
-          webcomponentService.resolveAlert(id, dismissKey);
-        }else{
-          ContainerAPI.closeAlert(id, dismissKey, iframeHandle);
+      thisComponent.closeAlert = (id: string, dismissKey?: string) => {
+        // check if thisComponent is in dom
+        if (thisComponent.isConnected) {
```
walmazacn, line 150: Good idea - is this check already included in tests?

walmazacn: Thanks - PR approved

PR #4128, opened 2025-01-24
add navheader data to navheader renderer

hardl:
... so data from titleResolver can be used in custom renderer

### DIFF
--- a/core/src/navigation/LeftNav.svelte
+++ b/core/src/navigation/LeftNav.svelte
@@ -253,7 +253,7 @@
         navHeaderContainer.innerHTML = '';
       }
 
-      navHeader.renderer(navHeaderContainer, navParentNode, clickHandler);
+      navHeader.renderer(navHeaderContainer, navParentNode, clickHandler, navHeader);
     }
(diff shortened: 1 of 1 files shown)


core/src/navigation/LeftNav.svelte
```diff
@@ -253,7 +253,7 @@
         navHeaderContainer.innerHTML = '';
       }
 
-      navHeader.renderer(navHeaderContainer, navParentNode, clickHandler);
+      navHeader.renderer(navHeaderContainer, navParentNode, clickHandler, navHeader);
```
walmazacn, line 256: Do we have to pass entire 'navHeader' object, so as to get a chunk related to 'titleResolver'? Perhaps it could be optimized a bit?

hardl: Good point, but in this particular case access to the title resolver and its outcome was the main driver for this PR. We can still think about optimizing when move it out of experimental.

walmazacn (approved):
LGTM 👍

PR #4137, opened 2025-02-11
Appswitcher enhancements

hardl:
Adds an option for appswitcher items to be resolved as selected for other routes than their corresponding links. Additionally it can be matched against context values of the current node.


JohannesDoberer (approved):
LGTM

PR #4159, opened 2025-02-19
Left nav small enhancements

hardl:
* calls blur() on mouse up in btp left nav to avoid weird overlap
* adds current nodes context to navheader for renderer


walmazacn (approved):
Left nav item loses focus when mouseup method is triggered - if 'btpToolLayout' is enabled

PR #4148, opened 2025-02-18
Fix: Wrong backgroundColor for profile initials

JohannesDoberer:
Currently with the following settings the Initials has no backgroundColor:
Bildschirmfoto 2025-02-18 um 12 12 28

Can be reproduced by the following steps:

1. index.html
  -> replace `<link rel='stylesheet' href='/public/luigi.css'>` with `<link rel='stylesheet' href='/public/luigi_horizon.css'>`
2. Luigi config:
```
settings: {
  ....
   responsiveNavigation: 'Fiori3',
   profileType: 'Fiori3',
   experimental: {
      profileMenuFiori3: true
   }
},
navigation: {
  ...nodes,
  profile: {
          logout: {
            label: 'Sign Out',
            icon: "sys-cancel"
          },
          staticUserInfoFn: () => {
            return new Promise(resolve => {
              resolve({
                name: 'Static User',
                initials: 'LU',
                email: 'other.luigi.user@example.com',
                description: 'Luigi Developer'
              });
            });
          }
        }
}
```

PR #4163, opened 2025-02-24
Fix user menu popover initials bgColor

JohannesDoberer:
Same config like #4148 but in user menu popover


walmazacn (approved):
LGTM 👍

walmazacn (approved):
Everything looks good

PR #4185, opened 2025-03-07
Fix appswitcher spacing and nav

hardl:
fixes styling and routing on click in case of "single-entry" in appswitcher (btpLayout)


walmazacn (approved):
LGTM 👍
(tested in default mode and with btpLayout enabled)

PR #4194, opened 2025-03-14
leftnav-btp: fix observer issue

hardl:


walmazacn (approved):
I tested it with 'btpToolLayout' enabled and everything works as expected

PR #4195, opened 2025-03-17
luigi-example-openui5: Use horizon theme in luigi

flovogt:
Use the latest recommended luigi settings.


JohannesDoberer (approved):
👍 
Thx for contribution

PR #4196, opened 2025-03-17
Fix cypress record key is null

JohannesDoberer:


test/e2e-js.sh
```diff
@@ -15,7 +15,7 @@ WS_FID_PID=$!
 NG_EXAMPLE="$BASE_DIR/../test/e2e-test-application"
 
 cd $NG_EXAMPLE
-if [ "$USE_CYPRESS_DASHBOARD" == "true" ]; then
+if [ "true" == "true" ] && [ -n "$CYPRESS_DASHBOARD_RECORD_KEY" ] ; then
```
walmazacn, line 18: It seems the first condition (`"true" == "true"`) is not needed as it's truthy all the time?

JohannesDoberer: Thx 👍 
Leftover from testing

walmazacn (approved):
LGTM 👍

PR #4247, opened 2025-04-11
Extend luigi event w callback

hardl:


container/src/constants/communication.ts
```diff
@@ -177,11 +178,19 @@ export namespace Events {
...
-  constructor(type: string, data: unknown, payload?: unknown) {
+  constructor(type: string, data: unknown, payload?: unknown, callback?: Function) {
     super(type);
     this.detail = data;
     this.payload = payload || data || {};
+    this.callbackFn = callback;
+  }
+
+  callback(data: unknown): void {
+    if (this.callbackFn) {
```
walmazacn, line 191: Would be safer to check 'callbackFn' also with `GenericHelperFunctions.isFunction`

hardl: i guess that is what typescript is for ;) 
Note that this callbackFn is only defined by us internally, not by the a customer dev

container/typings/constants/events.d.ts
```diff
@@ -177,11 +178,19 @@ export namespace Events {
...
-  constructor(type: string, data: unknown, payload?: unknown) {
+  constructor(type: string, data: unknown, payload?: unknown, callback?: Function) {
     super(type);
     this.detail = data;
     this.payload = payload || data || {};
+    this.callbackFn = callback;
+  }
+
+  callback(data: unknown): void {
+    if (this.callbackFn) {
```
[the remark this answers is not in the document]
hardl: see other reply

walmazacn (approved):
LGTM 👍

PR #4250, opened 2025-04-11
Adds update job for workflows

### LINKED ISSUE #4240: Github actions: create a new "update" job
Since we have to add following lines to test.yml
line 114
```
- run: sudo apt-get update
- run: sudo apt-get -y install libgtk2.0-0t64 libgtk-3-0t64 libgbm-dev libnotify-dev libnss3 libxss1 libasound2t64 libxtst6 xauth xvfb
```
and to mend.yml to get our pipelines running we can create a new "update" job that runs before this jobs and use the reference in both files.

walmazacn:
**Description**

- adds update job for workflows

**Related issue(s)**

Resolves #4240


.github/workflows/mend.yml
```diff
@@ -21,8 +23,9 @@ jobs:
         with:
           path: ~/.cache
           key: whitesource_cache-${{ github.ref }}
-      - run: sudo apt-get update
-      - run: sudo apt-get -y install build-essential python-is-python3 libgtk2.0-0t64 libgtk-3-0t64 libgbm-dev libnotify-dev libnss3 libxss1 libasound2t64 libxtst6 xauth xvfb
+      - uses: ./.github/workflows/update-deps
+        with:
+          list-of-deps: 'build-essential python-is-python3 libgtk2.0-0t64 libgtk-3-0t64 libgbm-dev libnotify-dev libnss3 libxss1 libasound2t64 libxtst6 xauth xvfb'
```
ndricimrr, line 28: This PR reveals we have:

 `libgtk2.0-0t64 libgtk-3-0t64 libgbm-dev libnotify-dev libnss3 libxss1 libasound2t64 libxtst6 xauth xvfb`
 
  dependencies in `mend.yml` almost identical with **test.yml** for e2e Cypress tests. 
  
  **Mend Scans** and **Cypress** have little in common, so I would assume this was just "copy-pasted" in the very early days of making these workflows 🤔 . Maybe worth removing them, if we don't need them all. 
  
...AI says Mend doesn't need any of these libs and you can remove all of them - **but I wouldn't trust it completely** :D , maybe worth trying first. 

Screenshot 2025-04-15 at 14 00 51

walmazacn: It seems you're right - Mend workflow is fully functional without this step

PR #4257, opened 2025-04-15
Adds missing callback functions in container

### LINKED ISSUE #4249: Container: add missing callback functions in container service
e.g. for show_alert_request, etc.

walmazacn:
**Description**

- adds missing callback functions in container

**Related issue(s)**

Resolves #4249

### DIFF
--- a/container/src/services/container.service.ts
+++ b/container/src/services/container.service.ts
@@ -1,7 +1,6 @@
 import { Events, LuigiEvent } from '../constants/communication';
 import type { IframeHandle, ContainerElement } from '../constants/container.model';
 import { LuigiInternalMessageID } from '../constants/internal-communication';
-import { GenericHelperFunctions } from '../utilities/helpers';
 
 export class ContainerService {
   /**
@@ -52,7 +51,6 @@ export class ContainerService {
    * @param {ContainerElement} targetCnt the targeted HTML element onto which the event is dispatched
    * @param {Object} data custom data added to the event to be dispatched
    * @param {Function} callback
-   * @param {string} callbackName
    */
   dispatch(msg: string, targetCnt: ContainerElement, data: object, callback?: (arg?) => void, payload?: object): void {
     const customEvent = new LuigiEvent(msg, data, payload, callback);
@@ -133,7 +131,15 @@ export class ContainerService {
                 this.dispatch(Events.NAVIGATION_REQUEST, targetCnt, event.data.params);
                 break;
               case LuigiInternalMessageID.ALERT_REQUEST:
-                this.dispatchWithPayload(Events.ALERT_REQUEST, targetCnt, event, event.data?.data?.settings);
+                this.dispatchWithPayload(
+                  Events.ALERT_REQUEST,
+                  targetCnt,
+                  event,
+                  event.data?.data?.settings,
+                  (id: string, dismissKey?: boolean | string) => {
+                    targetCnt.notifyAlertClosed(id, dismissKey);
+                  }
+                );
                 break;
               case LuigiInternalMessageID.INITIALIZED:
                 this.dispatch(Events.INITIALIZED, targetCnt, event.data.params);
@@ -155,7 +161,10 @@ export class ContainerService {
                   Events.SHOW_CONFIRMATION_MODAL_REQUEST,
                   targetCnt,
                   event.data.data,
-                  event.data.data?.settings
+                  event.data.data?.settings,
+                  (modalResult: boolean) => {
+                    targetCnt.notifyConfirmationModalClosed(modalResult);
+                  }
                 );
                 break;
               case LuigiInternalMessageID.SHOW_LOADING_INDICATOR_REQUEST:
--- a/container/src/services/webcomponents.service.ts
+++ b/container/src/services/webcomponents.service.ts
@@ -78,7 +78,6 @@ export class WebComponentService {
    * @param msg the message to be delivered
    * @param data the data to be sent
    * @param callback the callback function to be called
-   * @param callbackName name of the callback function
    */
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   dispatchLuigiEvent(msg: string, data: object, callback?: (arg?: any) => void) {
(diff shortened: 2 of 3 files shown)


hardl (changes requested):
there are still some missing callbacks, e.g. pathExist, getCurrentRoute, ...

JohannesDoberer (approved):
LGTM

JohannesDoberer (changes requested):

JohannesDoberer (commented):
In addition I think it would make sense that we extend the current function https://github.com/SAP/luigi/blob/main/container/test-app/iframe/microfrontend.html#L210
using the promise and add a corresponding e2e test.
Same for pathExists

JohannesDoberer (approved):
LGTM

container/test-app/iframe/microfrontend.html
```diff
@@ -275,6 +296,13 @@ <h1 id="title">iFrame based Container</h1>
           });
       }
 
+      function testAlertCallback() {
+        LuigiClient.uxManager().showAlert({
+          text: 'Callback Test',
+          type: 'info',
+        });
```
hardl, line 303: Would be great to add '.then(...)' to test promise e2e

container/test-app/iframe/microfrontend.html
```diff
@@ -149,6 +152,16 @@ <h1 id="title">iFrame based Container</h1>
...
       }
 
+      function testConfirmationModalCallback() {
+        LuigiClient.uxManager().showConfirmationModal({
+          type: 'confirmation',
+          header: 'Callback Test',
+          body: 'Are you sure you want to do this?',
+          buttonConfirm: 'Yes',
+          buttonDismiss: 'No',
+        });
```
hardl, line 162: Would be great to add '.then(...)' to test promise e2e

container/test-app/iframe/iframeContainer.html
```diff
@@ -126,7 +126,14 @@ <h3>
         alert('show-alert-request message received: ' + JSON.stringify(event.detail));
         console.log(event.detail.data.data);
         alertSettings = event.detail.data.data;
-        showAnAlert(alertSettings);
+
+        if (alertSettings?.settings?.text !== 'Callback Test') {
+          showAnAlert(alertSettings);
+        } else {
+          if (event.callbackFn && typeof event.callbackFn === 'function') {
```
hardl, line 133: callbackFn is private and therefore should not be used, people (and AI) might look into this test app and think this is how it is supposed to be used.
Use event.callback() instead.

container/test-app/iframe/iframeContainer.html
```diff
@@ -148,7 +155,14 @@ <h3>
         const data = event.detail;
         const val = confirm(data.body);
         alert('show-confirmation-modal message received: ' + JSON.stringify(data));
-        luigiContainer.notifyConfirmationModalClosed(val);
+
+        if (data?.data?.data?.settings?.header !== 'Callback Test') {
+          luigiContainer.notifyConfirmationModalClosed(val);
+        } else {
+          if (event.callbackFn && typeof event.callbackFn === 'function') {
```
hardl, line 162: callbackFn is private and therefore should not be used, people (and AI) might look into this test app and think this is how it is supposed to be used.
Use event.callback() instead.

container/test-app/iframe/iframeContainer.html
```diff
@@ -228,6 +242,13 @@ <h3>
           Events.GET_CURRENT_ROUTE_REQUEST,
           'message received: ' + JSON.stringify(event.detail.data),
         );
+
+        if (event.callbackFn && typeof event.callbackFn === 'function') {
+          event.callbackFn({
```
hardl, line 247: callbackFn is private and therefore should not be used, people (and AI) might look into this test app and think this is how it is supposed to be used.
Use event.callback() instead.

container/test-app/iframe/iframeContainer.html
```diff
@@ -254,6 +275,13 @@ <h3>
           Events.CHECK_PATH_EXISTS_REQUEST,
           'message received: ' + JSON.stringify(event.detail.data),
         );
+
+        if (event.callbackFn && typeof event.callbackFn === 'function') {
+          event.callbackFn({
```
hardl, line 280: callbackFn is private and therefore should not be used, people (and AI) might look into this test app and think this is how it is supposed to be used.
Use event.callback() instead.

container/cypress/e2e/test-app/iframe/iframe-container.cy.js
hardl, line 1: i'd rather not use spy in e2e tests but rely on actual dom changes induced by the functionality. It's also easier to debug the tests.

container/src/services/container.service.ts
```diff
@@ -194,7 +203,21 @@ export class ContainerService {
                 }
                 break;
               case LuigiInternalMessageID.GET_CURRENT_ROUTE_REQUEST:
-                this.dispatchWithPayload(Events.GET_CURRENT_ROUTE_REQUEST, targetCnt, event, event.data.data);
+                this.dispatchWithPayload(
+                  Events.GET_CURRENT_ROUTE_REQUEST,
+                  targetCnt,
+                  event,
+                  event.data.data,
+                  (data: object) => {
```
hardl, line 211: there is no need to expose the correlation id to the customer developer and educate him where to grab it from the payload and to put it as part of the callback data.
We know where it is in the internal payload, so extract it and set it fixed in the data object of the postmessage in callback

container/src/services/container.service.ts
```diff
@@ -209,7 +232,21 @@ export class ContainerService {
...
+                this.dispatchWithPayload(
+                  Events.CHECK_PATH_EXISTS_REQUEST,
+                  targetCnt,
+                  event,
+                  event.data.data,
+                  (data: object) => {
+                    target.postMessage(
+                      {
+                        msg: LuigiInternalMessageID.SEND_PATH_EXISTS_ANSWER,
+                        data
```
hardl, line 244: there is no need to expose the correlation id to the customer developer and educate him where to grab it from the payload and to put it as part of the callback data.
We know where it is in the internal payload, so extract it and set it fixed in the data object of the postmessage in callback

hardl (changes requested):

hardl (approved):
LGTM

PR #4266, opened 2025-04-18
Preserve non client viewgroup iframes

### LINKED ISSUE #4258: Make it possible to preserve non-client viewgroup iframes
by adding an option for a viewGroup in viewGroupSettings, e.g. **noClientCheck** or sth., that prevents a nav check when reactivating the iframe, which currently leads to a reload because there is no **luigi.navigation.ok** response.

walmazacn:
**Description**

- adds option to preserve non client viewgroup iframes

**Related issue(s)**

Resolves #4258


core/src/services/iframe.js
```diff
@@ -113,6 +113,18 @@ class IframeClass {
...
+    const noClientCheck = viewGroupSettings && viewGroup ? viewGroupSettings[viewGroup]?.noClientCheck : false;
+
+    /**
+     * check for `noClientCheck` attribute
+     * when set to `true`, it prevents a navigation check when reactivating the iframe
+     */
+    if (noClientCheck) {
+      return;
+    }
+
```
hardl, line 127: Please never do sth. like iframe..contentWindow.xyz as it will fail for cross-origin cases
```suggestion
    if (config?.iframe?.vg && NavigationHelpers.getViewGroupSettings(config.iframe.vg)?.noClientCheck) {
      return;
    }
```

hardl: same check must be done in `initHandshakeFailed` after first config.iframe.luigi check

hardl (changes requested):

hardl (approved):
lgtm

PR #4275, opened 2025-04-28
Keyboard events in Fiddle

### LINKED ISSUE #4272: Override 'save' cmd for fiddle
It should be possible to apply the changed luigi config by key event 'save'.

walmazacn:
**Description**

- adds keyboard events in Fiddle

**Related issue(s)**

Resolves #4272

### DIFF
--- a/website/fiddle/src/App.svelte
+++ b/website/fiddle/src/App.svelte
@@ -1,11 +1,11 @@
 <script>
+  import { onDestroy, onMount } from 'svelte';
+
   import luigiCorePkgInfo from '../node_modules/@luigi-project/core/package.json';
   import defaultConfig from './defaultConfig.js';
-  import { onMount } from 'svelte';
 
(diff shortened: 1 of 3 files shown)


website/fiddle/src/App.svelte
```diff
@@ -101,10 +105,30 @@
     }
   }
 
+  function initKeyboardEvents() {
```
JohannesDoberer, line 108: I think the function can be shortened. The shortened version works on OSX. Can't test it on win.
Why do we need the metaKeyEnabled variable?

```
function initKeyboardEvents() {
    window.addEventListener('keydown', (event) => {
      if ((event.ctrlKey || event.metaKey) && event.key === 's'){
        event.preventDefault();
        saveConfig();
      }
    });
  }
```

JohannesDoberer (approved):

PR #4286, opened 2025-05-14
fd v0.40 update

JohannesDoberer:
fundamental styles v0.40


walmazacn (approved):
It seems everything is fine - including left navigation

Issue #4281, opened 2025-05-06
User Menu


JohannesDoberer:

JohannesDoberer:
The new user menu is described here https://sap.github.io/fundamental-styles/?path=/docs/sap-fiori-components-user-menu--docs

- Determine and define a proper name for the new profileType.
- Ensure the new profileType is gated behind an experimental flag.
- Support only the simple version of the feature, without any popovers.

PR #4294, opened 2025-05-22
add Vega User Menu

VincentUllal:
**Description**

- add new User Menu 

**Related issue(s)**

See #4281


core/src/AuthorizationVegaProfileMenu.svelte
```diff
@@ -0,0 +1,315 @@
+<script>
```
walmazacn, line 1: It seems the logic here is the same like in 'Authorization.svelte' component - perhaps it would be better to create two dummy components (just html and styles) for both profile types (one for fiori and one for vega)? Then they can be injected into 'Authorization.svelte' component

Thanks to it we don't have to take care of same logic in two different places

walmazacn (changes requested):
New profile menu looks good, but I'd suggest to look into my comments

walmazacn (approved):
LGTM 👍

PR #4316, opened 2025-06-23
App switcher

JohannesDoberer:


core-modular/dev-tools/templates/simple/luigi-config.js
```diff
@@ -1,6 +1,46 @@
...
+          {
+            title: 'experimental2',
+            link: '/experimental/exp2',
+            subTitle: 'experimental2'
+          }
+        ],
+        itemRenderer_: (item, slot, appSwitcherApiObj) => {
+          let a = document.createElement('a');
+          a.setAttribute('style', 'background-color: red; color:green;');
+          console.log('Hello from itemRenderer');
```
walmazacn, line 35: Is it needed?

core-modular/src/services/navigation.service.ts
```diff
@@ -1,13 +1,32 @@
...
 export interface TopNavData {
   appTitle: string;
   logo: string;
   topNodes: [any];
   productSwitcher?: ProductSwitcher;
   profile?: ProfileSettings;
+  appSwitcher?: AppSwitcher;
 }
 
+export interface AppSwitcher {
```
walmazacn, line 13: Would be great to keep all interfaces/types in a separate file - then it's easier to import them in other places. In my opinion there's no need to mix it with some service logic

core-modular/src/services/navigation.service.ts
```diff
@@ -277,6 +302,23 @@ export class NavigationService {
     return undefined;
   }
 
+  getAppSwitcherData(appSwitcherData: AppSwitcher, headerSettings: any) {
```
walmazacn, line 305: Would be nice to have return type for each method that actually returns some value. Sth like:
`getAppSwitcherData(appSwitcherData: AppSwitcher, headerSettings: object): AppSwitcher {...}`

core-modular/src/services/navigation.service.ts
JohannesDoberer: Will probably be implemented with another issue.

PR #4349, opened 2025-07-02
Fix undefined bug appSwitcherData

ndricimrr:
When appSwitcherData is undefined, this produces a runtime error, due to access to `appSwitcherData.items`


ndricimrr:
bad branching

PR #4359, opened 2025-07-07
Fix drawer top position

JohannesDoberer:
There are a few config options to test:
- drawer + backdrop true/false
- drawer + breadcrumbs + backdrop true/false
- drawer + btpToolLayout + backdrop true/false
- drawer + btpToolLayout  + breadcrumbs + backdrop true/false
- drawer + vega  + breadcrumbs + backdrop true/false


walmazacn (approved):
LGTM 👍

Issue #4324, opened 2025-06-26
wc-starter-poc


VincentUllal:
Add a script that opens webcomponent based microfrontends, without setup, for testing/developement

--> See https://github.tools.sap/luigi-project/wc-starter-poc

PR #4361, opened 2025-07-08
4324 wc starter

VincentUllal:
add a script that takes a path to a webcomponent, inserts it into a container and starts a simple webserver with the wc and the communication panel from XTest

Usage:
go to the container folder and run the script `npm run start-wc <pathToWc>`

**Related issue(s)**
See #4324


VincentUllal:
moved to github enterprise: https://github.tools.sap/luigi-project/wc-starter-poc

PR #4366, opened 2025-07-11
Fixes border radius issue in usermenu

### LINKED ISSUE #4364: Vega: investigate border radius in usermenu/profile
in the current state we have a border radius (top @ firstchild, bottom @ lastchild):

which should not be there according to fd styles https://sap.github.io/fundamental-styles/?path=/docs/sap-fiori-components-user-menu--docs

walmazacn:
**Description**

- fixes border radius issue in usermenu

**Related issue(s)**

Fixes #4364

### DIFF
--- a/core/src/AuthorizationVegaProfileMenu.svelte
+++ b/core/src/AuthorizationVegaProfileMenu.svelte
@@ -271,6 +271,7 @@
     vertical-align: middle;
     margin-right: 5px;
   }
+
   .nav-icon {
     height: 2em;
   }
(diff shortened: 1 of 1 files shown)


JohannesDoberer (approved):

walmazacn:
Root cause of reported issue comes from the following set of styles in core:
https://github.com/luigi-project/luigi/blob/main/core/src/App.svelte#L2622

To be more specific - this is how 'luigi-menu__link--border-radius' mixin works:
https://github.com/luigi-project/luigi/blob/main/core/src/styles/_mixins.scss#L13

PR #4378, opened 2025-07-18
Browserlist cleanup

### LINKED ISSUE #4374: Investigate if .browserlistrc is still needed
Investigation of `.browserlistrc` necessity.
It was introduced with this PR 
With this issue https://github.com/luigi-project/luigi/pull/601/files#diff-43302d1852d8010ee012aab4dda2d01abb664e11f73617fe625f2ba112a958ed 

If we don't need it anymore just delete it.

walmazacn:
**Description**

- browserlist cleanup

**Related issue(s)**

Resolves #4374

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,1 +1,0 @@
-last 1 version, not dead, > 0.2%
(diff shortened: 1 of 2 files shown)


walmazacn:
After investigation it seems we don't need this file anymore

PR #4412, opened 2025-07-31
NodeParams in core-headless

JohannesDoberer:
- addNodeParams in core api
- getNodeParams in hashChanged function
- missing: send nodeParams via context to LuigiClient


core-modular/src/modules/routing-module.ts
```diff
@@ -11,6 +12,15 @@ export const RoutingModule = {
       window.addEventListener('hashchange', (ev) => {
         console.log('HashChange', location.hash);
         const path = NavigationHelpers.normalizePath(location.hash);
+        
+        const [, query] = path.split("?");
```
walmazacn, line 16: I think such pattern is not very readable - if you want to skip first variable, then at least mark it as sth optional:
`const [_, query] = path.split("?");`

PR #4447, opened 2025-08-19
Client permission for search params in core-modular

JohannesDoberer:


core-modular/src/services/navigation.service.ts
```diff
@@ -215,7 +218,12 @@ export class NavigationService {
     if (path == '') {
       // poor mans implementation, full path resolution TBD
       return pathData.rootNodes[0].pathSegment;
-    } else if (pathData.selectedNode && !pathData.selectedNode.viewUrl && pathData.selectedNode.children?.length > 0) {
+    } else if (
+      pathData.selectedNode &&
+      !pathData.selectedNode.viewUrl &&
+      Array.isArray(pathData.selectedNode.children) &&
```
hardl, line 224: pathData.selectedNode.children?.length

PR #4463, opened 2025-08-25
Docu for subcategories

JohannesDoberer:
Decided not to add the edge case to the docu because it is not the recommended way.
In addition it is documented in the issue ;)


docs/navigation-configuration.md
```diff
@@ -190,3 +190,81 @@ To define all subsequent nodes, use the category id:
...
+        sideNav: {
+            style: 'vega'
+        }
+    }
+});
+```
+
+You can define a subcategory by using the syntax `category: { id: 'mfe::subCat', ... }`. The double colon `::` is used to define a subcategory.
+The double colon is the default delimiter and can be changed with the property [settings.sideNav.subCategoryDelimiter](https://docs.luigi-project.io/docs/general-settings?section=sidenavsubcategorydelimiter).
+Additionally, the [settings.sideNav.style](https://docs.luigi-project.io/docs/general-settings?section=sidenavstyle) property must be set to `vega`. The `vega` theme is also available starting from Luigi version 2.23.0.
```
hardl, line 270: `'vega' theme` could be misleading, as it is not a theme in our current terminology. Replace by e.g. 'layout'.

hardl (changes requested):

hardl (approved):

PR #4480, opened 2025-09-02
Localization functions in core modular

### LINKED ISSUE #4476: Localization related functions in core modular
Implement localization related functions in core modular

walmazacn:
**Description**

- adds localization functions in core modular

**Related issue(s)**

Resolves #4476

### DIFF
--- a/core-modular/test/utilities/helpers/generic-helpers.spec.ts
+++ b/core-modular/test/utilities/helpers/generic-helpers.spec.ts
@@ -10,7 +10,7 @@ describe('Generic-helpers', () => {
     windowLocationImplementation = window.location;
     delete window.location;
     window.location = {
-      search: function() {
+      search: function () {
         return '';
       }
     } as any;


core-modular/src/utilities/defaultLuigiTranslationTable.ts
```diff
@@ -0,0 +1,23 @@
+const defaultLuigiInternalTranslationTable: Record<string, any> = {
```
JohannesDoberer, line 1: This should be used as well. I talked with Philipp and the head lib should get the translated values.

JohannesDoberer (changes requested):
Some comments.
Please also provide an example something like in e2e-test-application that we can test everything:
```
  customTranslationImplementation = () => {
    return {
      getTranslation: (key, interpolations, locale) => {
        return '*' + key + '* ' + (locale || Luigi.i18n().getCurrentLocale());
      }
    };
  };
  ```

**Comment by walmazacn** [comment]
TODO - discuss when and where to use 'addCurrentLocaleChangeListener' method from 'i18nService':
https://github.com/luigi-project/luigi/blob/main/core/src/App.svelte#L1179

Without updating the context the following method won't return correct result:
`LuigiClient.uxManager().getCurrentLocale()`

**Review by JohannesDoberer — APPROVED** [review]

File: core-modular/src/services/navigation.service.ts [hunk 2]
```diff
@@ -169,7 +169,7 @@ export class NavigationService {
     if (pathSegments?.length > 0 && pathSegments[0] === '') {
       pathSegments = pathSegments.slice(1);
     }
-    const rootNodes = cfg.navigation?.nodes || [];
+    const rootNodes = JSON.parse(JSON.stringify(cfg.navigation?.nodes)) || [];
```
JohannesDoberer: Why is this necessary?

walmazacn: It's for making original config data immutable

core-modular/src/services/i18n.service.ts
```diff
@@ -138,13 +138,12 @@ export class i18nService {
    * Sets locale to all Luigi containers
    */
   private broadcastLocaleToAllContainers(locale: string): void {
-    const containerWrapper = this.luigi.getEngine()._connector?.getContainerWrapper();
+    const containers = document.querySelectorAll('luigi-container');
```
JohannesDoberer: Discuss in daily
const containers = Array.from(document.querySelectorAll('[class^="luigi-"]'))
  .filter(el => Array.from(el.classList)
    .some(c => c.endsWith('-container'))
  );

PR #4481, opened 2025-09-04
4479 web component helpers in core modular

### LINKED ISSUE #4479: move 'web-component-helpers.js' to 'core-modular/src/utilities/helpers'

VincentUllal:

move 'web-component-helpers' to 'core-modular/src/utilities/helpers'
move tests to 'core-modular/test/utilities/helpers'

**Related issue(s)**
Resolves #4479

### DIFF
--- a/core-modular/src/utilities/helpers/web-component-helpers.ts
+++ b/core-modular/src/utilities/helpers/web-component-helpers.ts
@@ -1,0 +1,216 @@
+/**
+ * Default compound renderer.
+ */
+class DefaultCompoundRendererClass {
+  rendererObject: any;
+  config: any;
+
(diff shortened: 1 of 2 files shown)


core-modular/src/utilities/helpers/web-component-helpers.ts
```diff
@@ -0,0 +1,216 @@
+/**
+ * Default compound renderer.
+ */
+class DefaultCompoundRendererClass {
```
walmazacn, line 4: Please export const directly instead of creating new class - like here:
https://github.com/luigi-project/luigi/blob/main/core-modular/src/utilities/helpers/state-helpers.ts

It also applies to other classes in this file

PR #4493, opened 2025-09-15
Fix vega category subitem space

JohannesDoberer:


walmazacn (approved):
LGTM 👍

JohannesDoberer:
After research and discussion it should be no icon support at all for category sub items

PR #4554, opened 2025-10-28
JSON parse resilience

hardl:
Catching JSON.parse errors at several places to avoid follow-up errors that are only fixable by deleting invalid localStorage keys


walmazacn (approved):
Looks good - no issues while bundling core

PR #4572, opened 2025-11-11
headless - path routing 

JohannesDoberer:


JohannesDoberer:
follow-up #4579 and #4577

PR #4645, opened 2025-12-04
Container close current modal

JohannesDoberer:

PR #4651, opened 2025-12-05
headless - close current modal

JohannesDoberer:
implementation in headless for #4645


JohannesDoberer:
fixed in a better way #4653

PR #4653, opened 2025-12-05
Headless close modal internal

JohannesDoberer:


core-modular/src/services/modal.service.ts
```diff
@@ -0,0 +1,96 @@
+import type { Luigi } from '../core-api/luigi';
+import type { ModalSettings } from './navigation.service';
+
+export interface ModalPromiseObject {
+  closePromise?: Promise<void>;
+  resolveFn?: () => void;
+  onCloseRequestHandler?: () => void;
+  onInternalClose?: Function;
```
walmazacn: Be consistent and use `() => void` instead of `Function`

core-modular/src/types/connector.ts
```diff
@@ -16,7 +16,12 @@ export interface LuigiConnector {
 
   getContainerWrapper(): HTMLElement;
 
-  renderModal(content: HTMLElement, modalSettings: ModalSettings, onCloseCallback?: Function): any;
+  renderModal(
+    content: HTMLElement,
+    modalSettings: ModalSettings,
+    onCloseCallback?: Function,
```
walmazacn: Please use `() => void` instead of `Function`

PR #4681, opened 2025-12-23
Context as signal in client support angular

### LINKED ISSUE #4672: provide context as signal in client-support-angular
**Description**

right now in LuigiContextService there are 3 options, Context, Context as observable and Context as promise

I propose to add a forth one and that it's Context as a signal

also I noticed there is a dependency on zone https://github.com/luigi-project/luigi/blob/main/client-frameworks-support/client-support-angular/projects/client-support-angular/src/lib/service/luigi-context.service.impl.ts#L14
I think we need to remove it so that we can move to zoneless

**Reasons**

 this will ensure that the context is easily integrated with the new Angular signals

walmazacn:
**Description**

- adds  context as signal in client support angular

**Related issue(s)**

Resolves #4672

### DIFF
--- a/client-frameworks-support/client-support-angular/projects/client-support-angular/README.md
+++ b/client-frameworks-support/client-support-angular/projects/client-support-angular/README.md
@@ -3,7 +3,7 @@
 This library provides several features which make it easier to run your Angular application inside the Luigi micro frontend framework.
 If you want to know more about Luigi, please have a look at the [Luigi homepage](https://luigi-project.io/).
 
-> **NOTE:** Starting from v20 and on this package will only be released alongside Angular versions. So, `@luigi-project/client-support-angular@v20.0.0` will support Angular `v20.0.0` and so on. 
+> **NOTE:** Starting with version 20, this package will be aligned with Angular's major versioning. For example, '@luigi-project/client-support-angular@^20' will support '@angular/core@^20', and so on. 
 
 ## How to use the library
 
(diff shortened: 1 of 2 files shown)


client-frameworks-support/client-support-angular/projects/client-support-angular/src/lib/service/luigi-auto-routing.service.ts
```diff
@@ -1,42 +1,32 @@
...
 
   constructor(
     private router: Router,
     private luigiContextService: LuigiContextService
   ) {
-    this.subscription.add(
-      this.router.events.pipe(this.doFilter() as any).subscribe(this.doSubscription.bind(this) as () => void)
-    );
+    this.router.events
+      .pipe(this.doFilter(), takeUntilDestroyed(this.destroyRef))
```
robertIsaac: AFAIK you don't need to pass destroyRef if you are using it from the constructor

robertIsaac:
@walmazacn when do you plan to merge and release this?

walmazacn:
> @walmazacn when do you plan to merge and release this?

This PR has to be reviewed internally - if no issues are reported then it can be merged and published. It may happen this week, so track the latest changelog

JohannesDoberer:
Hey @robertIsaac,
we released @luigi-project/client-support-angular@v21.1.0

Cheers,
Johannes

PR #4689, opened 2026-01-07
openViewInNewTab in core modular

### LINKED ISSUE #4580: headless - open view in new tab
Open view in new tab - [see](https://github.com/luigi-project/luigi/blob/main/core/src/App.svelte#L1093) 
Also - https://docs.luigi-project.io/docs/luigi-client-api?section=newtab

walmazacn:
**Description**

- adds openViewInNewTab in core modular

**Related issue(s)**

Resolves #4580

### DIFF
--- a/core/examples/luigi-example-next/package-lock.json
+++ b/core/examples/luigi-example-next/package-lock.json
@@ -10,12 +10,12 @@
       "dependencies": {
         "@luigi-project/client": "^2.26.0",
         "@luigi-project/core": "^2.26.0",
-        "next": "^15.5.9",
-        "react": "^19.2.0",
-        "react-dom": "^19.2.0"
+        "next": "^15.5.11",
+        "react": "^19.2.4",
+        "react-dom": "^19.2.4"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


core-modular/src/modules/communicaton-module.ts
```diff
@@ -15,7 +15,9 @@ export const CommunicationModule = {
       UXModule.luigi?.ux().hideLoadingIndicator(containerElement.parentNode);
     });
     containerElement.addEventListener(Events.NAVIGATION_REQUEST, (event: any) => {
-      luigi.navigation().navigate(event.detail.link, event.detail.preserveView, event.detail.modal, event.callbackFn);
+      const { link, preserveView, modal, newTab } = event.detail;
+
+      luigi.navigation().navigate(link, preserveView, modal, newTab, event.callbackFn);
```
JohannesDoberer: We cannot change the signature of the navigate fn from core-api (backwards compatibility).
With this [PR](https://github.com/luigi-project/luigi/pull/4733) we have a new function where all this options like newTab, withoutSync etc can be handled.

PR #4750, opened 2026-01-27
node data-management

JohannesDoberer:


core-modular/src/services/navigation.service.ts
```diff
@@ -756,7 +801,7 @@ export class NavigationService {
   }
 
   private prepareRootNodes(navNodes: any[], context: Record<string, any>): Node[] {
-    const rootNodes = cloneDeep(navNodes) || [];
```
walmazacn: You can remove 'cloneDeep' import if it's not needed anymore

core-modular/src/services/node-data-management.service.ts
```diff
@@ -50,7 +50,7 @@ export class NodeDataManagementService {
    * Returns the root node
    * @returns {Node} root node
```
walmazacn: Change it to 'RootNode' as well

walmazacn (changes requested):
Please apply minor suggestions from me - the rest looks good

core-modular/src/services/navigation.service.ts
```diff
@@ -498,13 +534,23 @@ export class NavigationService {
...
+      activeNode = undefined;
+    } else if (selectedNode.tabNav) {
+      parentNode = lastElement;
     } else {
-      navItems = this.buildNavItems([...pathToLeftNavParent].pop()?.children || [], selectedNode, pathData);
+      parentNode = [...pathToLeftNavParent].pop();
     }
+
+    let children = (await this.getChildren(parentNode, parentNode?.context || {})) || [];
+    navItems = this.buildNavItems(children, activeNode, pathData);
```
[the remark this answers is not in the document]
walmazacn: It works as expected

PR #4797, opened 2026-02-12
Temp ng support esm client

hardl:


JohannesDoberer:
fixed in https://github.com/luigi-project/luigi/pull/4747

PR #4853, opened 2026-03-04
From context from closest context

JohannesDoberer:


core-modular/src/utilities/helpers/routing-helpers.ts
```diff
@@ -498,20 +498,28 @@ export const RoutingHelpers = {
    * @param params - Optional query parameters to append to the final route.
    * @returns a string representing the full route from the root to the given node, including query parameters if provided.
    */
-  buildRoute(node: Node, path?: string, params?: string): string {
+  getNodePath(node: Node, params?: string): string {
```
walmazacn: Now it seems 'getNodePath' has JSdoc description related to 'buildRoute' - add new description or at least apply the correct order

core-modular/src/utilities/helpers/generic-helpers.ts
```diff
@@ -245,5 +245,27 @@ export const GenericHelpers = {
...
   escapeRegExp(string: string): string {
     return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+  },
+
+  /**
+   * Checks if a given input string begins a hash with slash
+   * @param {string} path
+   * @returns {boolean}
+   */
+  hasHash: (path: string): boolean => {
```
walmazacn: Would be nice to have some unit tests for these new methods

PR #4960, opened 2026-03-30
Fd update 41

hardl:


core/src/navigation/LogoTitle.svelte
```diff
@@ -350,13 +350,9 @@
 {/if}
 
 <style lang="scss">
-  // Force height because of base64 img src
-  .lui-customlogo img {
-    height: 24px;
```
walmazacn, line 355: I tested it in test application and without 'height' property logo image is not visible

Image

walmazacn (changes requested):

walmazacn (approved):
LGTM 👍

PR #4995, opened 2026-04-07
Jest location in core modular

walmazacn:
Fixes unit test issues in core modular - mostly related to mocking window location


JohannesDoberer (approved):
LGTM 👍

walmazacn:
There's a better solution to solve this problem

PR #5006, opened 2026-04-10
get microfrontends

JohannesDoberer:


core-modular/src/core-api/dom-elements.ts
```diff
@@ -45,4 +48,73 @@ export class Elements {
   getNavFooterContainer(): HTMLElement | null {
     return this.luigi.getEngine()._connector?.getCoreAPISupportedElements().getNavFooterContainer() || null;
   }
+
+  /**
+   * Returns a list of all available micro frontends.
+   * @returns {Array<{id: string, active: boolean, container: HTMLElement, type: 'main'|'split-view'|'modal'}>} list of objects defining all micro frontends from the DOM
```
walmazacn: You could use 'MicrofrontendInDom' type which is already defined (but not exported) - then it can also be used as return type in 'getMicrofrontends' method

walmazacn (approved):
LGTM 👍

PR #5010, opened 2026-04-13
Breadcrumbs in core modular

walmazacn:
**Description**

- adds breadcrumbs in core modular

**Related issue(s)**

Resolves #4906


core-modular/dev-tools/templates/simple/luigi-config.js
```diff
@@ -10,6 +10,11 @@ window.onload = () => {
           }
         }
       },
+      breadcrumbs: {
```
JohannesDoberer, line 13: We need `clearBeforeRender` and `renderer` properties as well.
First there will be other heads not only ui5 headlib and second backwards compatibility.

core-modular/dev-tools/templates/simple/luigi-ui/lui-ui5wc.js
```diff
@@ -649,6 +650,37 @@ const connector = {
...
+    const selectedNode = breadcrumbData.selectedNode;
+
+    breadcrumbData.items.forEach((item) => {
+      const el = document.createElement('ui5-breadcrumbs-item');
+
+      if (item.node) {
+        el.textContent = `${item.label}`;
+
+        if (item.node.label !== selectedNode.label && item.node.pathSegment !== selectedNode.pathSegment) {
+          el.setAttribute('href', item.route);
```
JohannesDoberer: IMO we should handle routing in core-modular like other routes in TopNav, LeftNav and TopNav (navClick)

JohannesDoberer (commented):
Please see my comments

JohannesDoberer (approved):
LGTM

core-modular/src/services/navigation.service.ts
```diff
@@ -577,6 +579,84 @@ export class NavigationService {
...
+
+    const hashRouting = this.luigi.getConfigValue('routing.useHashRouting');
+    const currentPath = RoutingHelpers.getCurrentPath(hashRouting);
+    const start = breadcrumbConfig.omitRoot ? 2 : 1;
+
+    for (let i = start; i < nodesInPath.length; i++) {
+      const node = nodesInPath[i];
+      const route = RoutingHelpers.mapPathToNode(currentPath.path, node);
+
+      if (route && previousBreadcrumbs[route]) {
```
JohannesDoberer, line 635: The caching mechanism does not work as expected. This condition is never `true` because previousBreadCrumbs[route] will be always undefined because of L. 587

JohannesDoberer (changes requested):

core-modular/src/services/navigation.service.ts
```diff
@@ -27,15 +27,28 @@ import { NavigationHelpers } from '../utilities/helpers/navigation-helpers';
...
+import { get, writable } from '../utilities/store';
+import type { LuigiStore } from '../utilities/store';
 import { AuthLayerSvc } from './auth-layer.service';
 import { serviceRegistry } from './service-registry';
 import { ModalService } from './modal.service';
 
 export class NavigationService {
   modalService?: ModalService;
   nodeDataManagementService?: NodeDataManagementService;
+  private _breadcrumbStore: LuigiStore;
```
JohannesDoberer: The _breadcrumbStore uses a reactive writable() store, but it's never subscribed to anywhere — the reactivity is unused. A simple class property (navigation.service.ts is a singleton) `private previousBreadcrumbs: Record<string, BreadcrumbItem> = {}` would suffice, avoids the get() overhead, and is more consistent with the core.

core-modular/src/services/navigation.service.ts
```diff
@@ -577,6 +580,86 @@ export class NavigationService {
...
+  async getBreadcrumbData(path: string, pData?: PathData): Promise<BreadcrumbData> {
+    const breadcrumbConfig = this.luigi.getConfigValue('navigation.breadcrumbs');
+    const pathData = pData ?? (await this.getPathData(path));
+    const nodesInPath = pathData?.nodesInPath || [];
+    const navItems: BreadcrumbItem[] = [];
+    let showBreadcrumb;
+    let basePath = '';
+
+    if (!breadcrumbConfig || (path === '' && nodesInPath?.[0].viewUrl)) {
+      return {};
```
JohannesDoberer, line 593:
```suggestion
      this.previousBreadcrumbs = {};
      return {};
```
IMO we need to clear the cache here. For example when configChanged() will be called and breadcrumb will be disabled there is no need to hold the cache.

JohannesDoberer (commented):
One small comment rest looks good for me 👍

JohannesDoberer (approved):

PR #5011, opened 2026-04-13
switch from jsdom to happy-dom

hardl:


core-modular/test/utilities/helpers/generic-helpers.spec.ts
```diff
@@ -2,20 +2,27 @@ import { GenericHelpers } from '../../../src/utilities/helpers/generic-helpers';
...
   let locationSearchString: string;
+  let originalLocation: Location;
 
   beforeAll(() => {
-    jest.spyOn(window, 'window', 'get').mockImplementation(() => {
-      return {
-        location: {
-          search: locationSearchString
-        },
-        crypto: globalThis.crypto
```
walmazacn, line 12: Just wonder - is 'window.crypto' not needed to be mocked anymore?

core/test/services/web-components.spec.js
```diff
@@ -150,7 +150,8 @@ describe('WebComponentService', function () {
           return myEl;
         });
       sb.stub(container, 'replaceChild').callsFake(() => {});
-      sb.stub(window, 'location').value({ origin: 'http://localhost' });
+      // Don't stub entire window.location in happy-dom as it breaks document.baseURI
+      // sb.stub(window, 'location').value({ origin: 'http://localhost' });
```
walmazacn: It seems this line might be removed?

core-modular/test/utilities/helpers/generic-helpers.spec.ts
hardl: it never was, it was just made available at the mocked window object. Since now only the location of the window is mocked, crypto is simply there ;)

walmazacn (approved):
LGTM 👍

PR #5021, opened 2026-04-14
replace vite in core with rollup

hardl:
potential fix for https://github.com/luigi-project/luigi/security/dependabot/2457


JohannesDoberer (approved):
LGTM 👍

PR #5032, opened 2026-04-15
Replace template virtual tree path

JohannesDoberer:
Adds a placeholder resolution variable virtualTreePath to the viewUrl processing , which will be replaced by the sub-route, if defined. If not defined, sub-route will be appended as before.


walmazacn (approved):
LGTM 👍

PR #5036, opened 2026-04-16
bump node

hardl:
bumps node version in workflow files, see https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/


walmazacn (approved):
LGTM 👍

PR #5042, opened 2026-04-21
Loading indicator in modal and drawer

walmazacn:
**Description**

- adds loading indicator for modal and drawer in core modular

**Related issue(s)**

Partially resolves #4939


JohannesDoberer (commented):
As far as I can see in core the loadingIndicator will not disappear.
How to reproduce:
Change e.g. the viewURL of empty page to an html file which does not exist and open this node as a modal and you can see the loadingIndicator stays.

In addition for the contentarea the LI (loadingIndicator) will be opened and closed via core-modular (connector api) (and when the handshake is done) and for modals/dialog the LI will be opened and closed manually.

JohannesDoberer (commented):
Tested it and this is still valid for me. (..but it is also not that important in core-modular simpledev but for head we need the same functionality like in core)

-> Change e.g. the viewURL of empty page (in fiddle) to an html file which does not exist and open this node as a modal and you can see the loadingIndicator stays.

JohannesDoberer (dismissed):
LGTM

JohannesDoberer (approved):

PR #5064, opened 2026-04-30
Prevent custom message dispatch from triggering internal event l…

JohannesDoberer:
- Add `isCustomMessage` flag to `addEventListener` to distinguish custom message listeners from internal event listeners
- Filter custom message dispatch to only invoke listeners explicitly registered via `addCustomMessageListener`
- Prevents internal event listeners with coincidentally matching names from being triggered by custom messages


walmazacn (approved):
LGTM 👍

PR #5067, opened 2026-05-05
postcss update

hardl:
potential fix for https://github.com/luigi-project/luigi/security/dependabot/2506 and friends


core/examples/luigi-example-next/package-lock.json
walmazacn, line 1: Should this file be completely removed?

hardl: yes

walmazacn (approved):
LGTM 👍

PR #5076, opened 2026-05-06
Support for intent mapping

walmazacn:
**Description**

- adds support for intent mapping

**Related issue(s)**

Resolves #4917


JohannesDoberer (approved):

JohannesDoberer:
following IntentMapping (it's the one from e2e-test-application)
```
intentMapping: [
        {
          semanticObject: 'Sales',
          action: 'settings',
          pathSegment: '/home2/c1/asdf/1'
        },
        {
          semanticObject: 'Component',
          action: 'settings',
          pathSegment: '/home2/c4/detail'
        },
        {
          semanticObject: 'External',
          action: 'view',
          externalLink: { url: 'https://www.sap.com/', openInNewTab: true }
        }
      ],
```
When I run this in **core**:
```
LuigiClient.linkManager().navigateToIntent('Sales-settings', {dynamic: 'pr1'})
LuigiClient.linkManager().navigate('/#?intent=Sales-settings?dynamic=pr1')
```
-> resulturl `http://localhost:4100/#/home2/c1/asdf/1?~dynamic=pr1`
+ leftnav node "dynchild" is selected and the correct content in contentarea

In **core-modular** when I run the same commands the nodeParams in the url are not present 
+ the leftNav node "dynchild" is not selected + wrong mf in contentarea + alert "luigi.notExactTargetNode"
-> result url `http://localhost:4400/#/home2/c1/asdf`

The same with component
`LuigiClient.linkManager().navigate('/#?intent=Component-settings?project=abc')`
-> `http://localhost:4400/#/home2/c4 instead http://localhost:4400/#/home2/c4/detail?~project=abc`
+ alert "luigi.notExactTargetNode"

External works as expected (edited)

core-modular/src/utilities/helpers/routing-helpers.ts
JohannesDoberer: The function signature says this function returns an object `{ path: string; query: string }`. Here it returns undefined. Needs to be checked if this could lead to an runtime-error in handleRouteChange when .path will be called.

core-modular/src/services/navigation.service.ts
```diff
@@ -879,7 +880,21 @@ export class NavigationService {
...
-    const computedPath = await this.buildPath(path, options || {});
+    const hashRouting = this.luigi.getConfig().routing?.useHashRouting;
+    let computedPath = await this.buildPath(path, options || {});
+
+    if (intent) {
+      const intentPath = RoutingHelpers.getIntentPath(path.replace('#', ''), this.luigi);
+
+      if (typeof intentPath === 'string') {
+        computedPath = intentPath;
+      } else {
```
JohannesDoberer: When I run `LuigiClient.linkManager().navigateToIntent('External-view');`
I'm getting a url like: `http://localhost:4400/#/home2/c1#?intent=External-view``
In core the url stays.

JohannesDoberer (changes requested):
Please see comments

core-modular/src/utilities/helpers/routing-helpers.ts
walmazacn: It's not true - even if new tab is opened code execution doesn't stop here. For me an object `{ path: string; query: string }` is returned in each case

core-modular/src/services/navigation.service.ts
```diff
@@ -879,7 +880,23 @@ export class NavigationService {
...
+        const locationPath = hashRouting ? location.hash : location.pathname;
+
+        RoutingHelpers.handleExternalIntentPath(intentPath as Record<string, any>);
+
+        computedPath = locationPath;
+      }
```
JohannesDoberer:
```suggestion
} else if (intentPath && typeof intentPath === 'object') {
    RoutingHelpers.handleExternalIntentPath(intentPath);
    return;
  } else {
    return;
  }
```
When we have a call to an external link we can stop the handleNavigationRequest here, no need to go ahead in this function. We do the same in core (routing.js L.345)
Also suggest to check if intentPath is an object, because if intentPath is false it runs into handleExternalIntentPath, which also is checked there, but in my opinion it's a little bit cleaner, but also matter of taste.. up to you.

core-modular/src/utilities/helpers/routing-helpers.ts
```diff
@@ -331,32 +378,45 @@ export const RoutingHelpers = {
...
-          history.replaceState((window as any).state, '', intentPath);
-        }
+          if (isReplaceRouteActivated) {
+            history.replaceState((window as any).state, '', intentPath);
+          }
 
-        return { path: intentPath, query: location.search };
+          return { path: intentPath, query: location.search };
+        } else {
+          RoutingHelpers.handleExternalIntentPath(intentPath as Record<string, any>);
```
JohannesDoberer, line 397:
```suggestion
          RoutingHelpers.handleExternalIntentPath(intentPath as Record<string, any>);
          return { path: '', query: '' };
```
+ in handleRouteChange:
```
async handleRouteChange(
  routeInfo: { path: string; query: string },
  withoutSync = false,
  preventContextUpdate = false
): Promise<void> {
  const path = routeInfo.path;
  if (!path) return;
```
For example: When you are in  /home2/c1/details and you add the hash `location.hash = ?intent=External-view` in browser console the browser doesn't navigate to /home2 in core here it does.. but in both cases the url is wrong (maybe it needs to clarified with Philipp)
Happens only if `useHashRouting=false`

JohannesDoberer (commented):
Please see my comments.
In addition I realised we don't have a core-api implementation
````
Luigi.navigation().navigateToIntent(...)
```
Maybe you can also added to this branch.

JohannesDoberer (approved):
LGTM 👍

PR #5101, opened 2026-05-11
handleViewUrlMisconfigured

JohannesDoberer:


core-modular/src/services/routing.service.ts
```diff
@@ -554,4 +560,45 @@ export class RoutingService {
...
+   */
+  async handleViewUrlMisconfigured(
+    node: Node,
+    viewUrl: string,
+    previousPathData: PathData,
+    pathUrlRaw: string
+  ): Promise<boolean> {
+    const { children, intendToHaveEmptyViewUrl, compound } = node;
+
+    const hasChildrenNode = (children && Array.isArray(children) && children.length > 0) || children || false;
```
walmazacn: I'm not sure why do we need the second step in this check. You already check if 'children' exist in step one
Let's say 'children' is an empty table and in such case 'hasChildrenNode' equals empty table as well - I suppose it's not an expected result?

walmazacn (dismissed):
LGTM 👍

walmazacn (dismissed):
✅

walmazacn (approved):
Fixes look good

PR #5107, opened 2026-05-13
add commit-msg hook to check DCO

hardl:
commit-msg hook, fails if commit message has duplicate or no "Signed-off-by" line.


walmazacn (approved):
LGTM 👍

PR #5116, opened 2026-05-15
Context switcher in test app

### LINKED ISSUE #5082: Context switcher in test app
Implement context switcher in test app relying on data provided by core modular

walmazacn:
**Description**

- adds context switcher in test app

**Related issue(s)**

Resolves #5082

### DIFF
--- a/core-modular/dev-tools/templates/simple/luigi-ui/lui-ui5wc.js
+++ b/core-modular/dev-tools/templates/simple/luigi-ui/lui-ui5wc.js
@@ -452,9 +452,9 @@ const connector = {
       shellbar.innerHTML = html;
 
       if (topNavData.contextSwitcher) {
-        const data = {...topNavData.contextSwitcher};
+        const data = { ...topNavData.contextSwitcher };
         const switcher = document.createElement('ui5-combobox');
-        const createSwitcherOption = function(item, type) {
+        const createSwitcherOption = function (item, type) {
           const element = data.config.customOptionsRenderer ? 'div' : 'ui5-cb-item';
           const option = document.createElement(element);
           let label = item.label;
@@ -495,7 +495,7 @@ const connector = {
           switcher.setAttribute('placeholder', data.config.defaultLabel);
         }
 
-        switcher.addEventListener('selection-change', function(event) {
+        switcher.addEventListener('selection-change', function (event) {
           const selectedType = event.detail?.item?.attributes?.type?.textContent;
           const selectedValue = event.detail?.item?.attributes?.value?.textContent;
 
@@ -625,7 +625,7 @@ const connector = {
         console.log('shellbar._lastTopNavData', shellbar._lastTopNavData);
       }
       if (topNavData.contextSwitcher) {
-        const data = {...topNavData.contextSwitcher};
+        const data = { ...topNavData.contextSwitcher };
         const switcher = shellbar.querySelector('ui5-combobox');
         if (!switcher) {
           return;
--- a/core-modular/dev-tools/templates/simple/luigi-ui/lui-ui5wc.js
+++ b/core-modular/dev-tools/templates/simple/luigi-ui/lui-ui5wc.js
@@ -451,6 +451,114 @@ const connector = {
 
       shellbar.innerHTML = html;
 
+      if (topNavData.contextSwitcher) {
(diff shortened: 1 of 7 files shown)


core-modular/dev-tools/templates/simple/luigi-ui/lui-ui5wc.js
```diff
@@ -516,6 +641,20 @@ const connector = {
...
       }
+      if (topNavData.contextSwitcher) {
+        const data = { ...topNavData.contextSwitcher };
+        const switcher = shellbar.querySelector('ui5-combobox');
+        if (!switcher) {
+          return;
+        }
+        if (data.selectedOption) {
+          switcher.setAttribute('selected-value', data.selectedOption.link);
+          switcher.setAttribute('value', data.selectedOption.label);
```
JohannesDoberer: When I remove the `label` in options and click on one of the entries in the context switcher e.g. Env4 I see `undefined`

JohannesDoberer (commented):
Looks good so far. Just a minor change needed. Rest works as expected

core-modular/dev-tools/templates/simple/luigi-ui/lui-ui5wc.js
```diff
@@ -451,6 +451,131 @@ const connector = {
...
+              const result = node.clickHandler(node);
+
+              if (!result) {
+                return;
+              }
+            }
+
+            setTimeout(() => globalThis.Luigi.navigation().navigate(selectedValue));
+          } else {
+            globalThis.Luigi.navigation().navigate(selectedValue);
```
JohannesDoberer: core-modular should do the navigation.
In leftNav and topNav have a navClickFn e.g. L48
We should not use `Luigi.navigation().navigate()` in head impl.

core-modular/src/services/navigation.service.ts
JohannesDoberer:
```suggestion
    const selectedNodePath = ContextSwitcherHelpers.getSelectedNode(currentPath?.path, options, parentNodePath);
```
`getSelectedNode` function is not async

core-modular/src/services/navigation.service.ts
```diff
@@ -1162,7 +1162,7 @@ export class NavigationService {
   }
 
   private async buildContextSwitcher(): Promise<ContextSwitcher> {
-    const actions = await this.luigi.getConfigValueAsync('navigation.contextSwitcher.actions');
+    const actions: ContextSwitcherItem[] = await this.luigi.getConfigValueAsync('navigation.contextSwitcher.actions');
     const config = this.luigi.getConfigValue('navigation.contextSwitcher');
     const options: ContextSwitcherItem[] = await ContextSwitcherHelpers.fetchOptions(undefined, this.luigi);
```
JohannesDoberer: Only here we call fetchOptions with `undefined`. Do we need this parameter in general? When is `existingOptions` needed?
The early return in fetchOptions will never be called in context-switcher-helpers.

core-modular/src/services/navigation.service.ts
```diff
@@ -1182,6 +1182,30 @@ export class NavigationService {
...
+
+        if (node?.clickHandler) {
+          const result = node.clickHandler(node);
+
+          if (!result) {
+            return;
+          }
+        }
+
+        setTimeout(() => this.luigi.navigation().navigate(selectedValue));
```
JohannesDoberer: Why is the `setTimeout` needed? Maybe a comment would help

core-modular/src/services/navigation.service.ts
```diff
@@ -1182,6 +1182,30 @@ export class NavigationService {
       options,
       parentNodePath
     );
+    const handleChangeEvent = (event: any) => {
+      const selectedType = event.detail ? event.detail?.item?.attributes?.type?.textContent : undefined;
+      const selectedValue = event.detail ? event.detail?.item?.attributes?.value?.textContent : event.target.value;
```
JohannesDoberer: This is tightly coupled to UI5 web component internals (item.attributes.type.textContent). If the UI layer changes, this breaks silently. Maybe we can make it more general?

JohannesDoberer (commented):
Please see comments

JohannesDoberer (approved):
LGTM 👍

PR #5146, opened 2026-05-26
remove --sapThemeMetaData-Base-baseLib prop from css

### LINKED ISSUE #5118: `core/luigi.css` contains invalid CSS syntax breaking Vite 6/7/8 builds
**Description**

The distributed production CSS file `@luigi-project/core/luigi.css` contains invalid/malformed CSS syntax. While older, more forgiving bundlers (like Vite 5 / esbuild) treated these syntax issues as non-blocking warnings, modern strict parsers—specifically **Lightning CSS**, which is the default CSS transformer and minifier in **Vite 6, 7, and 8**—treat them as fatal errors and abort the build.

This affects any modern frontend application trying to compile or update to current Vite versions while using Luigi Core.

**Expected result**

The distributed `luigi.css` asset should conform strictly to modern W3C CSS specifications so it can be safely processed, transformed, and minified by strict, high-performance tooling (like Lightning CSS, Rolldown, or Oxc) without crashing consumer build pipelines.

**Actual result**

Running a production build results in a fatal compilation failure during the CSS post-processing/minification stage:

```text
Error: [lightningcss minify] Unexpected end of input
1 | @charset "UTF-8";.luigi-alert--overlay.svelte-120xz9q{position:absolute;...
2 | /*$vite$:1*/
3 | ...ource": "11.32.2"}}')}:root --sapThemeMetaData-Base-baseLib:{"Path": "Base.baseLib.sap_fiori_3.css_variables",...
  |                                                                ^
4 | .fd-shell+(--fixed)
5 |     .fd-shell__header+(--fixed)

```

There are two root causes in the CSS output:

1. **Unescaped JSON Injections:** Raw JSON metadata blocks are attached to custom properties (e.g., `--sapThemeMetaData-...:{"Path":...}`) without being wrapped safely inside valid string literals. The strict parser interprets `{` as the opening of an unclosed CSS declaration block.
2. **Unresolved Preprocessor Code:** The selectors `.fd-shell+(--fixed)` and `.fd-shell__header+(--fixed)` appear to be uncompiled Sass/PostCSS mixins or macros that were shipped raw.

**Steps to reproduce**

1. Initialize a clean Vite 6 or Vite 8 project (e.g., `npm create vite@latest`).
2. Install the Luigi core package: `npm install @luigi-project/core`.
3. Import the CSS file in your main entrypoint (e.g., `main.js` or `style.css`):
```javascript
import '@luigi-project/core/luigi.css';

```

4. Execute a production build: `npm run build`.
5. The build will fail with the `[lightningcss minify] Unexpected end of input` exception.

**Troubleshooting**

The issue can be temporarily bypassed by consumers by explicitly downgrading Vite's internal CSS architecture to use PostCSS and the legacy `esbuild` minifier in `vite.config.js`:

```javascript
css: {
  transformer: 'postcss',
},
build: {
  cssMinify: 'esbuild',
}

```

This confirmation proves that the failure is purely down to modern tooling rejecting the non-standard syntax present in the shipped `luigi.css` bundle. A permanent fix requires updating the upstream build pipeline to wrap JSON theme metadata safely in strings and fully resolve CSS preprocessor macros before npm publication.

_Issue report created with help of AI tools_

hardl:
should fix https://github.com/luigi-project/luigi/issues/5118

### DIFF
--- a/core/rollup.config.mjs
+++ b/core/rollup.config.mjs
@@ -10,6 +10,15 @@ import * as fs from 'fs';
 
 const production = !process.env.ROLLUP_WATCH;
 
+// Removes the `--sapThemeMetaData-Base-baseLib:{...json...};` custom property
+// declaration. Its value is a JSON object literal whose nested braces are not
+// valid CSS custom-property syntax and break some downstream parsers. The
+// declaration is always followed by more declarations in the same `:root`
+// block, so the value's terminator is the declaration-separator `;`. We can't
+// anchor on `}` because the JSON contains nested `}` followed by `}` which
+// would match the inner brace and prematurely close the rule.
+const SAP_THEME_METADATA_RE = /--sapThemeMetaData-Base-baseLib:\{[\s\S]*?\};/;
+
 // Suppress warnings
 const onwarn = (warning, warn) => {
   // Suppress circular dependency warnings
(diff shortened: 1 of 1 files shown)


walmazacn (approved):
LGTM 👍

PR #5175, opened 2026-06-08
category selection indicator

hardl:
This PR fixes a visual issue where collapsed navigation categories in the left nav did not show any selection highlight when one of their child nodes was the currently active route.


walmazacn (approved):
LGTM 👍

PR #5181, opened 2026-06-11
API usage in auth OIDC pkce plugin

### LINKED ISSUE #5165: Replace private _client.* API usage in auth-oidc-pkce plugin
Summary:
  The @luigi-project/plugin-auth-oidc-pkce plugin reaches into a private property (_client) of the oidc-client-ts UserManager for its signout flow. Replace these calls with the public UserManager API so the plugin's contract with its dependency is stable across oidc-client-ts updates. 

Background:
  When the plugin was migrated from oidc-client to oidc-client-ts, two methods that were public on the old library's UserManager were no longer exposed publicly. The migration was  completed by accessing them via the internal _client property instead of switching to the public replacements. This means any patch release of oidc-client-ts could rename or remove  _client and silently break logout for all consumers.

Acceptance criteria:
  - [ ] No occurrences of _client remain in plugins/auth/src/auth-oidc-pkce/index.js.
  - [ ] logout() uses UserManager.signoutRedirect().
  - [ ] _processLogoutResponse() uses UserManager.signoutRedirectCallback().
  - [ ] signoutData.state is renamed to url_state and the encodeURI wrapper is removed.
  - [ ] _processLogoutResponse() returns early after handing off to the callback so the trailing resolve(true) is not reachable on the logout path.
  - [ ] Manual verification against oidc-mockserver.

walmazacn:
**Description**

- API usage in auth OIDC pkce plugin

**Related issue(s)**

Resolves #5165


plugins/auth/src/auth-oidc-pkce/index.js
JohannesDoberer, line 101:
```
logout(authData, authOnLogoutFn) {
  const signoutData = {
    id_token_hint: authData && authData.idToken,
    url_state: window.location.href
  };

  authOnLogoutFn();

  return this.client
    .signoutRedirect(signoutData)
    .catch(function (err) {
      console.error('[OIDC] logout() Error', err);
    });
}
```
signoutRedirect() internally navigates the browser via window.location.replace(), which means its promise never resolves and any .then() callback is dead code. Moving authOnLogoutFn() before the call ensures the onLogout lifecycle hook fires reliably before the page unloads.

JohannesDoberer (commented):
Please see comment.
In addition please double check. You can easily put a console log to the onLogout event in the luigi-config. With the current implementation I don't see the console log, with the suggestion I can see it.

JohannesDoberer (approved):
LGTM

PR #5221, opened 2026-06-16
Security: Open Redirect Vulnerability in Logout Handler

tomaioo:
## Summary

Security: Open Redirect Vulnerability in Logout Handler

## Problem

**Severity**: `High` | **File**: `core-modular/dev-tools/templates/simple/auth/idpmock/logout.html:L10`

The logout.html file contains an open redirect vulnerability. The `post_logout_redirect_uri` parameter is extracted from the URL and used for redirection after logout. While there is a check to ensure the redirect starts with `window.location.origin`, the `decodeChars` function uses a flawed regex replacement that only escapes special characters rather than properly validating the URL. An attacker could potentially bypass the origin check using URL encoding tricks, subdomain takeover, or by exploiting the loose `startsWith` check with techniques like `https://attacker.com?origin.com`. This could lead to phishing attacks by redirecting users to malicious sites after logout.

## Solution

Implement strict URL validation using the URL API to parse and validate the redirect URI. Use a whitelist of allowed redirect domains instead of relying on `startsWith`. Consider: `const url = new URL(redirectTo); if (url.origin !== window.location.origin) { redirectTo = '/auth/logout.html'; }`

## Changes

- `core-modular/dev-tools/templates/simple/auth/idpmock/logout.html` (modified)


core-modular/dev-tools/templates/simple/auth/idpmock/logout.html
```diff
@@ -9,11 +9,16 @@
     <script>
       window.onload = function () {
         let href = encodeURI(window.location.href);
-        const decodeChars = (string) => string.replace(/[<>'*+?^${}()|[\]\\]/g, '\\$&');
+        const decodeChars = (string) => string.replace(/[<>''*+?^${}()|[\]\\]/g, "\\\\$&");
```
JohannesDoberer, line 12: What is the idea behind the double backslash?
The url will be modified using double backslashes
```
  http://localhost:4400/auth/idpmock/logout.html\\?post_logout_redirect_uri=http%3A%2F%2Flocalhost%3A440...
```

| Input | Before | Afterwards |
|--------|--------|------------|
| `?` | `\?` | `\\?` |
| `+` | `\+` | `\\+` |
| `(` | `\(` | `\\(` |

JohannesDoberer (commented):
Hello @tomaioo,
thank you for your contribution.
Please see comment.
The origin check looks good to me 👍 .
In addition we are using [dco](https://github.com/apps/dco).
Could you please **Signed-off-by** to your commits?

PR #5247, opened 2026-06-19
fix: invalidate node cache before renderTopNav in configChanged

JohannesDoberer:
The NodeDataManagementService cache was only cleared after renderTopNav, causing getTopNavData to use stale cached rootNodes. Dynamically added navigation nodes via push+configChanged were not reflected in the shellbar.

Reproduce:
Browser console:
```
Luigi.getConfig().navigation.nodes.push({pathSegment: 'tets', label:'tets', viewUrl: '/microfrontend.html', icon:'home'})
Luigi.configChanged()
```

With this fix you should see the new node in top nav, before you didn't see it.


core-modular/src/modules/ui-module.ts
```diff
@@ -174,6 +174,15 @@ export const UIModule = {
         settings.header
     */
 
+    if (
+      noScopes ||
+      scopes.includes('navigation') ||
```
walmazacn, line 179: Should it be truthy even if 'navigation.nodes' or/and 'navigation.viewgroupdata' is falsy? If yes then checking for 'nodes' and 'viewgroupdata' is not needed in this statement

JohannesDoberer: The sub-scopes navigation.nodes and navigation.viewgroupdata can be passed independently without navigation (e.g. Luigi.configChanged('navigation.nodes')). In that case scopes.includes('navigation') is false, but the cache still needs to be invalidated since the node tree changed. This matches the scope checks in the blocks below (renderLeftNav, renderTabNav, etc.).

walmazacn (approved):
LGTM 👍

PR #5250, opened 2026-06-22
dependency overrides in ng support

hardl:
Overrides several dependencies.

Analysis Conclusion: the override-driven bumps (piscina 5.1.4→5.2.0, @babel/core 7.29.0→7.29.7, esbuild 0.27.7→0.28.1 in vite's nested tree) produce bit-identical published artifacts. Consumers cannot tell the difference.


walmazacn (approved):
LGTM 👍

PR #5257, opened 2026-06-23
updateTopNavigation in core modular

### LINKED ISSUE #5213: core-modular: Implement updateTopNavigation() for badge counter updates
## Summary

The public API method `Luigi.navigation().updateTopNavigation()` is missing in core-modular. This method is used in the original core to allow the core app to trigger badge counter refreshes.

Additionally, the aggregation logic for badge counters is missing — in the original core, badges from child nodes are summed up and displayed on their parent category node. This needs to be implemented in core-modular as well.

## Background

In the original `core`, the badge update flow works as follows:
1. The core app calls `Luigi.navigation().updateTopNavigation()`
2. Core sends a `postMessage` with `'luigi.navigation.update-badge-counters'`
3. `TopNav.svelte` and `LeftNav.svelte` listen for the message and re-render badges

## Acceptance Criteria

- [ ] `Luigi.navigation().updateTopNavigation()` is available in core-modular's Navigation API
- [ ] Calling the method triggers a re-evaluation of all `badgeCounter.count()` functions
- [ ] Both TopNav and LeftNav receive the badge data initially and after an update (known issue: ui5 webcomponent side-nav does not support badges)
- [ ] The UI layer (separate concern) is notified via the existing reactive store/update mechanism (e.g. `configChanged('navigation')` or a dedicated scope)
- [ ] Badge aggregation for category nodes is handled appropriately in core-modular
- [ ] Client compatibility: micro-frontends using the existing Client API continue to work without changes

## Technical Notes

- The type definition for `badgeCounter` is already in place (`core-modular/src/types/navigation.ts`)
- Consider whether aggregation logic (summing child badges for parent nodes) should live in core-modular
- The reactive update path should use the existing `_ui.update(scopes)` pattern rather than introducing a message-bus

walmazacn:
**Description**

- adds 'updateTopNavigation' in core modular

**Related issue(s)**

Resolves #5213


JohannesDoberer (commented):
~~Superseded by the review below (with core comparison).~~

JohannesDoberer (commented):
This is an AI review. Please check the listed points (if they are valid)

## Badge Counter Review — Comparison with Core

I compared the `buildNavItems` badge logic in core-modular with the original implementation in `core/src/utilities/helpers/navigation-helpers.js`. Several key behaviors from core are missing or incorrectly ported.

---

### 1. Category badge accumulation is missing

**Core (correct)** — accumulates badge count as each node joins the category:
```javascript
if (hasBadge && !cats[catLabel].badgeCounter) {
  cats[catLabel].badgeCounter = { label: '', count: () => badgeCount };
} else if (hasBadge) {
  const updatedCount = cats[catLabel].badgeCounter.count() + badgeCount;
  cats[catLabel].badgeCounter.count = () => updatedCount;
}
```

**Core-modular (broken)** — sets category badge only on first node, never updates:
```typescript
if (!catNode) {
  catNode = { badgeCounter: hasBadge ? { count: () => Number(badgeCount), ... } : undefined, ... };
}
// catNode.badgeCounter is NEVER updated when subsequent nodes join the category
```

**Result:** Category with nodes having counts 1, 2, 3 → core shows 6, core-modular shows 1.

---

### 2. Badge count is doubled for category child nodes

**Core** — calls `count()` once, uses `badgeCount` directly:
```javascript
badgeCount = await node.badgeCounter.count();
```

**Core-modular** — calls `count()` twice and sums both results:
```typescript
badgeCount = (node.badgeCounter.count as () => number)();          // first call → e.g. 5
nodeBadgeCount = (node.badgeCounter.count as () => number)();      // second call → e.g. 5
badgeCounter = { count: () => Number(nodeBadgeCount + badgeCount) }; // → 10 instead of 5
```

**Result:** Every category child's badge displays 2× its actual value.

---

### 3. `async forEach` instead of `for...of`

**Core (correct)** — uses `for...of` in an async function, so `await` works:
```javascript
for (const node of rawChildren) {
  badgeCount = await node.badgeCounter.count();  // properly awaited
}
```

**Core-modular (broken pattern)** — uses `async` callback in `forEach`, which does NOT await:
```typescript
nodes?.forEach(async (node) => {
  badgeCount = await (node.badgeCounter.count as () => Promise<number>)();
});
// code here runs BEFORE any async callback completes
```

Currently not an active bug because the `isPromise` check (see point 4) prevents the `await` branch from ever being reached. But if async counters are ever used, the function would return before resolution completes.

---

### 4. Unnecessary `isPromise` check — core just uses `await`

**Core** — simply awaits the result (works for both sync and async):
```javascript
badgeCount = await node.badgeCounter.count();
```

**Core-modular** — adds a broken `isPromise` guard on the function reference:
```typescript
if (GenericHelpers.isPromise(node.badgeCounter.count)) {  // checks if FUNCTION has .then → always false
  badgeCount = await (node.badgeCounter.count as () => Promise<number>)();
} else {
  badgeCount = (node.badgeCounter.count as () => number)();
}
```

`isPromise` checks for `.then` on the value — but `count` is a function, not a Promise. This check is always false. The simpler `await count()` pattern from core handles both cases correctly.

---

### 5. Dead code: unreachable branch

```typescript
if (hasBadge) {           // hasBadge = !!node.badgeCounter
  if (!node.badgeCounter) { // ← always false when hasBadge is true
    ...
  }
}
```

Lines 264-268 can never execute.

---

### Suggested fix

Align with core's approach:
1. Use `for...of` with `await` (make `buildNavItems` async)
2. Simply `await node.badgeCounter.count()` — no `isPromise` check needed
3. Accumulate category badge on each iteration (port core's `updatedCount` logic)
4. Remove the double-call / dead code

core-modular/src/services/navigation.service.ts
```diff
@@ -636,11 +686,12 @@ export class NavigationService {
     const pathDataTruncatedChildren = parentNode
       ? this.getTruncatedChildren(parentNode.children ?? [])
       : this.getTruncatedChildren(selectedNode.children ?? []);
+    const navData = await this.buildNavItems(pathDataTruncatedChildren, selectedNode, pathData);
```
JohannesDoberer, line 705:
```suggestion
    const navData = await this.buildNavItems(pathDataTruncatedChildren, selectedNode, pathData, true);
```
I would suggest to calculate the badge count also for leftnav, although it is not supported for ui5 head lib but for may further heads.

core-modular/src/services/navigation.service.ts
```diff
@@ -430,16 +474,20 @@ export class NavigationService {
       parentNode = [...pathToLeftNavParent].pop();
     }
 
-    let children = (await this.getChildren(parentNode, parentNode?.context || {})) || [];
-    navItems = this.buildNavItems(children, activeNode, pathData);
+    const children = (await this.getChildren(parentNode, parentNode?.context || {})) || [];
+    const navData = await this.buildNavItems(children, activeNode, pathData);
```
JohannesDoberer, line 475:
```suggestion
    const navData = await this.buildNavItems(children, activeNode, pathData, true);
```
I would suggest to calculate the badge count also for leftnav, although it is not supported for ui5 head lib but for may further heads.

JohannesDoberer (approved):
LGTM

PR #5258, opened 2026-06-23
Fix/nav item click link nodes

JohannesDoberer:
## Summary
Fixes navigation for nodes that use the link property instead of pathSegment
Clicking such nodes in the side navigation previously did nothing because getNodePath(node) requires pathSegment
Absolute links (e.g. /home/settings) now navigate directly
Relative links (e.g. dps/dps1) are resolved against the parent node's path

## Problem
Navigation nodes can be configured with a link property to redirect to another path:

{
  link: '/home/settings',
  label: 'Go to absolute path',
  icon: 'switch-views'
}
The getRouteLink() helper in routing-helpers.ts already handles node.link correctly for building `<a href>` attributes. However, navItemClick() — which handles the actual click event — only called getNodePath(node), which traverses pathSegment up the node tree. For link-nodes (which have no pathSegment), this produced an empty or undefined path, causing silent navigation failure.

## Fix
Added a check for node.link in navItemClick() before falling through to getNodePath():

Absolute links (/home/settings): navigate directly
Relative links (dps/dps1): resolve against the parent node's path via getNodePath(node.parent)

### DIFF
--- a/core-modular/src/services/navigation.service.ts
+++ b/core-modular/src/services/navigation.service.ts
@@ -451,6 +451,16 @@ export class NavigationService {
       return;
     }
 
+    if (node.link) {
+      const isAbsolute = node.link.startsWith('/');
+      if (isAbsolute) {
+        return this.luigi.navigation().navigate(node.link);
+      }
+      const parentPath = node.parent ? RoutingHelpers.getNodePath(node.parent) : '';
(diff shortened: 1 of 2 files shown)


JohannesDoberer (commented):
ai commented wrong pr

walmazacn (approved):
LGTM 👍

PR #5264, opened 2026-06-26
Storage manager unit tests

### LINKED ISSUE #5259: storageManager unit tests

walmazacn:
**Description**

- adds storage manager unit tests in client

**Related issue(s)**

Resolves #5259


JohannesDoberer:
This is a review of an AI tool.


## Review: `client/test/storageManager.spec.js`

### What works well

- All 6 public API methods (`setItem`, `getItem`, `removeItem`, `clear`, `has`, `getAllKeys`) are covered
- Good use of `it.each` for the `has` true/false cases
- Clean teardown with `jest.restoreAllMocks()` in `afterEach`

### Issues & suggestions

**1. Tests primarily verify mock behavior, not the actual implementation**

The real `execute` method is completely replaced in each test. This means the actual flow (`execute` → `createPendingOperation` → `sendMessage` → event-based resolution via `processEvent`) is never exercised.

The `toHaveBeenCalledWith` assertions are valuable — they confirm correct parameter delegation. But the mock implementations with the Map logic go beyond that and effectively test themselves.

**Suggestion:** Instead of mocking `execute`, mock `helpers.getRandomId` and `helpers.sendPostMessageToLuigiCore`, then trigger `processEvent` with a matching event to resolve the Promise. This tests the real flow end-to-end.

**2. The test-level `pendingOperation` Map (line 4) has no relation to the module's internal state**

This Map is a standalone variable only used inside mock implementations. Assertions like `expect(pendingOperation.size).toEqual(1)` verify self-managed mock state, not the actual internal state of `storageManager`. This creates false confidence.

**3. `setItem` test: misleading key naming**

`storageKey = 'test'` (line 29) suggests a relationship to the actual storage key `'keyExample'` passed to `setItem` (line 40), but they're unrelated. The condition `!pendingOperation.has(storageKey)` in the mock is always `true` because the Map starts empty — it's unnecessary logic that adds confusion without value.

**4. No reject/error paths tested**

None of the tests exercise the `reject` callback. The real implementation rejects when `data.status === 'ERROR'` in `processEvent` — this path should be covered.

**5. Constructor test is trivial**

The `beforeEach` replaces `storageEventProcessor` with a mock object, so `expect(storageManager.storageEventProcessor).toBeTruthy()` only verifies that the mock assignment worked — not the real constructor behavior.

---

### Summary

The tests confirm correct parameter delegation to `storageEventProcessor.execute`, which has value. To strengthen them:

1. Test the real `execute`/`processEvent` flow by mocking only at system boundaries (`helpers`)
2. Add error/reject scenarios
3. Remove the test-level `pendingOperation` Map — it doesn't reflect real module state

JohannesDoberer:
## Review: updated `client/test/storageManager.spec.js`

### Greatly improved since last iteration 👍

The tests now exercise the real flow (`execute` → `createPendingOperation` → `sendMessage` → `processEvent`) with mocks only at system boundaries (`helpers`). Error paths are covered. The fake `pendingOperation` Map is gone. Nice work.

### One remaining issue: reject tests can silently pass without asserting

All reject tests use a `try/catch` pattern without a fail-guard:

```js
try {
  await storageManager.setItem(resultMock.key, resultMock.value);
} catch (error) {
  expect(executeSpy).toHaveBeenCalledWith(...);
  expect(error).toEqual(resultMock);
}
```

If the Promise unexpectedly **resolves** instead of rejecting, the `catch` block is never entered → 0 assertions run → Jest considers the test passed. This means a broken reject path would still show green.

**Fix:** Add `expect.assertions(2)` as the first line in every reject test:

```js
it('should reject when storage operation returns an error', async () => {
  expect.assertions(2);
  // ... rest stays the same
});
```

Affected tests:
- `setItem` – "if promise is rejected" (line 53)
- `getItem` – "if promise is rejected" (line 98)
- `removeItem` – "if promise is rejected" (line 143)
- `clear` – "if promise is rejected" (line 188)
- `has` – "if promise is rejected" (line 229)
- `getAllKeys` – "if promise is rejected" (line 270)

### Minor: test descriptions for reject cases

"should store an item for a specific key if promise is rejected" reads contradictory — it doesn't store anything in the error case. Consider something like "should reject when setItem operation fails" to better describe what's actually being tested.

JohannesDoberer (approved):

PR #5267, opened 2026-06-29
Fix/open node in modal side nav

JohannesDoberer:
### Summary
- Nodes with `openNodeInModal` now correctly open their `viewUrl` in a modal without navigating the main content area
- Same approach as core: `href` is set for accessibility/Ctrl+click, but `preventDefault()` on the click event stops native anchor navigation

### Problem
Clicking a navigation node with `openNodeInModal: true` would:
1. Open the modal correctly (via `navItemClick`)
2. **Also** navigate the main content area to that node's path

The root cause was that the UI5 SideNavigation component renders an internal `<a href="...">` tag in its shadow DOM. On click, the browser's native link navigation fires before any `selection-change` event, changing `location.hash` and triggering `handleRouteChange` via the `hashchange` listener.

### Changes
- **`navigation.service.ts`** (`navItemClick`): Open modal and return early when `openNodeInModal` is set (supports both `true` and `ModalSettings` object)
- **`navigation.ts`**: Extend `openNodeInModal` type to `boolean | ModalSettings`
- **`lui-ui5wc.js`** (simple-app + template): Add capture-phase click listener that calls `event.preventDefault()` on trusted user clicks without meta keys — same pattern as core's `handleNavAnchorClickedWithoutMetaKey`. This prevents the shadow DOM anchor from navigating while keeping `href` intact for accessibility and Ctrl/Cmd+click (open in new tab).
- **`navigation.service.spec.ts`**: Unit tests for `navItemClick` with `openNodeInModal`

### Differences from previous approach
- `href` is **not** suppressed in `buildNavItems` — it remains set on all nodes (same as core)
- Navigation is prevented at the UI layer via `preventDefault()`, not at the data layer

### Test plan
- [x] Navigate to a node with `openNodeInModal: true` → modal opens, URL does not change
- [x] Navigate to a node with `openNodeInModal: { size: 'm' }` → modal opens with correct size
- [x] Ctrl/Cmd+click on any nav item → opens href in new tab
- [x] Close modal → previous content still visible
- [x] Normal navigation nodes still work as before
- [x] Unit tests pass

### DIFF
--- a/core-modular/dev-tools/templates/simple/luigi-ui/lui-ui5wc.js
+++ b/core-modular/dev-tools/templates/simple/luigi-ui/lui-ui5wc.js
@@ -339,7 +339,7 @@ function renderNodeOrCategory(item, leftNavData) {
       el.setAttribute('target', item.externalLink.sameWindow ? '_self' : '_blank');
     } else {
       el.setAttribute('luigi-route', leftNavData.basePath + '/' + item.node.pathSegment);
-      if (item.href) el.setAttribute('href', item.href);
+      if (item.href && !item.node?.openNodeInModal) el.setAttribute('href', item.href);
     }
     el._luigiItem = item;
     if (item.selected) el.setAttribute('selected', '');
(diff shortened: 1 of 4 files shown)


walmazacn (approved):
LGTM 👍

PR #5286, opened 2026-07-07
OIDC pkce logout url

hardl:
Legacy plugin-auth-oidc supported a distinct logoutUrl for internal error redirects (onAuthExpired, onAuthError, silent-renew failures, login-callback failures), separate from the
  OIDC-spec post_logout_redirect_uri used as the IdP end-session callback. The PKCE rewrite collapsed both into post_logout_redirect_uri, forcing users whose IdP endpoint differs from
  their app's error landing page to fork or monkey-patch.

  This PR:
  - Adds logoutUrl back as a first-class setting; the four error-redirect sites now route through it.
  - Falls back to post_logout_redirect_uri when logoutUrl is unset, so auth errors are always routed somewhere.


walmazacn (approved):
LGTM 👍

PR #5294, opened 2026-07-09
header - disabled

JohannesDoberer:
## Summary

When `settings.header.disabled` is set to `true` in the Luigi config, the shellbar (top navigation) should be hidden. Previously this was only handled by the legacy Luigi Core — core-modular had no support for it.

- `getTopNavData()` now returns `null` when `settings.header.disabled` is `true`, signaling to the connector that the header should not be rendered
- Updated `LuigiConnector.renderTopNav` type signature to accept `TopNavData | null`
- Updated the simple-app template connector (`lui-ui5wc.js`) to handle `null` in `renderTopNav` (hides shellbar) and decoupled `renderLeftNav` from the burger button so the side navigation still renders when the shellbar is absent

## Changes

- **`core-modular/src/services/navigation.service.ts`** — early return `null` from `getTopNavData` when header is disabled
- **`core-modular/src/types/connector.ts`** — `renderTopNav(data: TopNavData | null)`
- **`core-modular/dev-tools/templates/simple/luigi-ui/lui-ui5wc.js`** — null guard in `renderTopNav`, `renderLeftNav` no longer requires burger button to exist

## Test plan

- Start simple-app with `settings.header.disabled: true` in `luigi-config.js`
- Verify shellbar is not rendered
- Verify left navigation still renders correctly with nodes
- Verify `hideSideNav` still works on nodes when shellbar is present

### DIFF
--- a/client/luigi-client.d.ts
+++ b/client/luigi-client.d.ts
@@ -805,8 +805,16 @@ export type getCoreSearchParams = () => CoreSearchParams;
  * LuigiClient.addCoreSearchParams({luigi:'rocks'});
  * LuigiClient.addCoreSearchParams({luigi:'rocks', false});
  */
-export function addCoreSearchParams(searchParams: CoreSearchParams, keepBrowserHistory: Boolean, preventLuigiConfigUpdate: Boolean): void;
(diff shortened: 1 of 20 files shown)


walmazacn (dismissed):
LGTM 👍

walmazacn (dismissed):
✅

hardl (approved):

PR #5308, opened 2026-07-15
nodeparams with openAsModal

### LINKED ISSUE #5304: linkManager().withParams().openAsModal() breaks routing — node params appended to path
## Description

When calling `LuigiClient.linkManager().withParams({test: true}).openAsModal('/home/multipurpose', {...})` from within an iframe MFE, the route resolution fails with:

> Could not find the requested route /home/multipurpose?~test=true.

The node params (`~test=true`) are being appended to the path as a query string before route resolution, which causes the path matcher to fail.

## Expected behavior

Node params should be handled separately from path resolution. The modal should open at `/home/multipurpose` and the node params should be passed to the target MFE context.

## Steps to reproduce

1. Configure a node with `viewUrl` pointing to a page that uses LuigiClient
2. From that page, call: `LuigiClient.linkManager().withParams({test: true}).openAsModal('/home/multipurpose', {title: 'Test', size: 'm'})`
3. Observe error alert: "Could not find the requested route /home/multipurpose?~test=true."

JohannesDoberer:
### Summary

Fixes #5304

When calling `LuigiClient.linkManager().withParams({test: true}).openAsModal('/home/multipurpose')` from an iframe MFE, the node params were incorrectly appended to the path **before** route resolution, causing a "Could not find the requested route /home/multipurpose?~test=true" error.

### Root Cause

In `NavigationService.buildPath()`, node params are appended as query parameters (e.g. `?~key=value`) to the navigation path. For regular navigation this is fine — but for modals and drawers, this combined path is then passed to `pathExists()` / `getPathData()` which try to match it as a route segment and fail.

### Changes

- **`navigation.service.ts`** — For modal/drawer navigations, exclude `nodeParams` from `buildPath()` to prevent them from being appended to the route path. Instead, pass them separately via the settings object.
- **`navigation.ts`** (`openAsModal`) — After successful route resolution, append nodeParams to the modal URL path with `~` prefix (matching core behavior: `?modal=/path?~key=value`). Pass nodeParams separately to `createContainer` via `luigiParams` so the MFE receives them.
- **`navigation.ts`** (`openAsDrawer`) — Pass nodeParams to `createContainer` via `luigiParams`. Drawers don't persist data to the URL (consistent with core).
- **`ui-module.ts`** — Accept optional `luigiParams` in `openModal` and `openDrawer`, forwarding them to `createContainer`.
- **`types/navigation.ts`** — Add optional `nodeParams` field to `ModalSettings` and `DrawerSettings` interfaces.

### URL behavior (now matches core)

Before (broken): Route resolution fails because path includes params:

`/home2/c2?~test=true`

After (fixed): nodeParams correctly appended to modal path in URL, not blocking route resolution:

`http://localhost:4400/#/home2/c1?modal=%2Fhome2%2Fc2%3F~tets%3Dtest&modalParams=%7B%22title%22%3A%22MFE2%22%7D`

### Test Plan

- [x] Added unit test verifying nodeParams are appended to the modal path in the URL (not serialized in `modalParams`)
- [x] Updated existing test assertions for new `luigiParams` parameter in `openModal`/`openDrawer`
- [x] All 755 tests passing
- [x] Manual test: `LuigiClient.linkManager().withParams({test: 'true'}).openAsModal('/home2/c2')` opens modal without error
- [x] Manual test: URL contains `?modal=%2Fhome2%2Fc2%3F~test%3Dtrue` (nodeParams in modal path)
- [x] Manual test: MFE inside modal receives nodeParams via `LuigiClient.getNodeParams()`


walmazacn (dismissed):
LGTM 👍

walmazacn (approved):
✅

PR #5313, opened 2026-07-16
Fix/5298 get current route wc

### LINKED ISSUE #5298: `linkManager().getCurrentRoute()` not handled for WC/iframe micro frontends
## Description
The `getCurrentRoute()` API on `linkManager()` (including `fromClosestContext()`, `fromContext()`, `fromParent()`, and `fromVirtualTreeRoot()` variants) does not work in the UI5 library.

## Root Cause
The Luigi Container dispatches a `GET_CURRENT_ROUTE_REQUEST` event on the container element when a micro frontend calls `LuigiClient.linkManager().getCurrentRoute()`. However, no `addEventListener` for this event is registered in the library's container setup code (unlike other events like `NAVIGATION_REQUEST`, `ALERT_REQUEST`, etc. which are all handled).

The Promise returned by `getCurrentRoute()` never resolves.

## Expected Behavior
The library should handle the `GET_CURRENT_ROUTE_REQUEST` event and respond with the current route, similar to how it's handled in the original Luigi framework.

## Steps to Reproduce
1. Configure a WC node with `webcomponent: true` and a `navigationContext`
2. Inside the WC, call `this.LuigiClient.linkManager().fromClosestContext().getCurrentRoute()`
3. The returned Promise never resolves

JohannesDoberer:
## Fix: Handle `getCurrentRoute()` for WC/iframe micro frontends

Closes #5298

### Problem

`linkManager().getCurrentRoute()` and its variants (`fromClosestContext()`, `fromContext()`, `fromParent()`, `fromVirtualTreeRoot()`) never resolved when called from Web Component or iframe-based micro frontends. The `CommunicationModule` was missing a listener for the `GET_CURRENT_ROUTE_REQUEST` event dispatched by the Luigi Container.

### Root Cause

The Luigi Container emits a `GET_CURRENT_ROUTE_REQUEST` event when micro frontends invoke `getCurrentRoute()`. While other events (e.g. `NAVIGATION_REQUEST`, `ALERT_REQUEST`) were properly handled in `CommunicationModule.addListeners()`, this one was missing — causing the Promise to hang indefinitely.

An additional subtlety: for iframe-based micro frontends, the Container's `dispatchWithPayload` places the navigation options (`fromParent`, `fromClosestContext`, etc.) on `event.payload` rather than `event.detail`. The handler uses `event.payload` which works correctly for both WC and iframe sources.

### Solution

1. **Added `getCurrentRoutePath()` method** to `NavigationService` that computes the current route path relative to a context node. It reuses the existing `getPathData()` and `RoutingHelpers.getSubPath()` infrastructure but applies subtraction semantics (current node path minus context node path) rather than the concatenation semantics of `buildPath()`.

2. **Added event listener** for `Events.GET_CURRENT_ROUTE_REQUEST` in `CommunicationModule` that delegates to the new method and responds via `event.callback()`.

### Changed Files

- `core-modular/src/modules/communicaton-module.ts` — added `GET_CURRENT_ROUTE_REQUEST` listener
- `core-modular/src/services/navigation.service.ts` — added `getCurrentRoutePath()` method
- `core-modular/test/services/navigation.service.spec.ts` — 9 unit tests covering all variants

### How to Test

1. `cd core-modular && npm run build && npm run simpledev`
2. Navigate to `#/home2/wc`
3. Click the `getCurrentRouteFromParent` button → should display `/wc`
4. Click `getCurrentRouteFromClosestContext` → should display `/wc` (requires `navigationContext` on parent node)


walmazacn (dismissed):
LGTM 👍

walmazacn (approved):
✅

PR #5344, opened 2026-07-22
fix: sanitize child_process call in bootstrap.js...

anupamme:
## Summary
Code hygiene improvement to \`bootstrap.js\`: replace \`exec()\` with \`execFile()\`.

## Background
Semgrep rule \`javascript.lang.security.detect-child-process.detect-child-process\` flagged this as a potential command injection. After manual review, this is a **false positive**:

- \`folder\` comes only from the hardcoded literal array at lines 8–24 of \`bootstrap.js\`; no external input flows into the call.
- \`bootstrap.js\` is a dev/CI helper script at the repo root, not included in any published npm package.
- Semgrep fired because \`folder\` is a function parameter — the rule cannot see across the call site that the caller always passes a hardcoded string from the same file. This is the classic false-positive shape for this taint rule (it labels itself "needs manual review" for exactly this reason).

The change from `` exec(`cd ${folder} && npm install`) `` to \`execFile('npm', ['install'], { cwd: folder })\` is accepted as a **code hygiene improvement** — avoiding shell string interpolation even for trusted input.

## Changes
- \`bootstrap.js\`: use \`execFile\` with \`cwd\` option instead of \`exec\` with a shell string

## Verification
- [x] Build passes
- [x] \`npm run bootstrap\` works correctly with the new implementation

### DIFF
--- a/bootstrap.js
+++ b/bootstrap.js
@@ -2,7 +2,7 @@
  * This file is used to run (npm install) in all of the folders which have it as a prerequisite in the pipeline and during local development
  */
 
-const { exec } = require('child_process');
+const { execFile } = require('child_process');
 
 // Array of folder names
 const folders = [
@@ -29,9 +29,7 @@ let isVerbose = verboseFlagIndex !== -1;
 // Function to install npm packages in given folder
 function installPackages(folder, index, totalFolders) {
   return new Promise((resolve, reject) => {
-    const command = `cd ${folder} && npm install`;
-
-    exec(command, (error, stdout, stderr) => {
+    execFile('npm', ['install'], { cwd: folder }, (error, stdout, stderr) => {
       if (error) {
         console.error(`\x1b[31mError installing npm packages in ${folder} \x1b[0m`);
         reject(error);


hardl:
Thanks for the report, but this looks like a Semgrep false positive rather than a real vulnerability. A few things:

  1. folder is not user-controllable.

  Look at lines 8–24 of bootstrap.js: the array is a hardcoded list of literal strings ('./', 'core', 'core-modular', …) written directly in the source file. The folder parameter is
  populated only by iterating that array. There is no external input path — no argv (aside from the --verbose flag which never touches exec), no env var, no file read, no network. To
  influence what folder contains, an attacker would need write access to bootstrap.js itself, at which point command injection is not the interesting attack.

  2. bootstrap.js is not shipped to consumers.

  This script lives at the repo root and is invoked via npm run bootstrap for local development and CI. It is not part of any published npm package — @luigi-project/core,
  @luigi-project/client, @luigi-project/container, etc. each have their own package.json under their respective subdirs, and none include this file in their published contents. The
  "Node.js library, affects downstream consumers" framing in the PR description doesn't apply here.

  3. Semgrep flagged this at LOW confidence with "needs manual review".

  The rule uses taint tracking and fires whenever a function parameter flows into exec(), even if the caller's argument is a hardcoded string from the same file. That's exactly why the
  rule labels itself as needing manual review — the rule can't see across the call site to determine that the values are literals from an in-file array. This is the classic false-positive
  shape for this rule.

  Happy to reconsider if there's a concrete attack scenario I'm missing — but as it stands, this looks like a scanner-metric fix rather than a real security issue.

We will consider to apply it for code hygiene reasons though...

anupamme:
I've updated the PR description. As for a concrete attack scenario, I don't have it yet.

JohannesDoberer (approved):
LGTM 👍
Thank you for your contribution!! 👍

PR #5350, opened 2026-07-23
Npm audit fix script

hardl:
introduces a script that runs through all relevant packages and calls 'npm audit fix'


walmazacn (approved):
LGTM 👍

PR #5352, opened 2026-07-23
Check pinned script

hardl:
script that checks if all external dependencies in package.json files in the monorepo are pinned to exect version


walmazacn (approved):
LGTM 👍

PR #5353, opened 2026-07-23
Feature/iframe creation interceptor core modular

JohannesDoberer:
container version is just for testing. Will be replaced when container iframeCreationInterceptor is released


core-modular/src/modules/ui-module.ts
```diff
@@ -15,7 +15,7 @@ import type { LuigiParams } from '../types/routing';
 import { GenericHelpers } from '../utilities/helpers/generic-helpers';
 import { AuthHelpers } from '../utilities/helpers/auth-helpers';
 
-const createContainer = async (node: Node, luigi: Luigi, luigiParams?: LuigiParams): Promise<HTMLElement> => {
+const createContainer = async (node: Node, luigi: Luigi, luigiParams?: LuigiParams, microFrontendType?: string): Promise<HTMLElement> => {
```
walmazacn: You could create a special literal type for this case - it can limit values to given scope:
`type MFType = 'main' | 'modal' | 'drawer' | 'usersettings'`

walmazacn (dismissed):
LGTM 👍

walmazacn (dismissed):
LGTM 👍

hardl (approved):

PR #5355, opened 2026-07-24
Anchor handling in core modular

### LINKED ISSUE #5307: `setAnchor` not handled for Web Component micro frontends in core-modular
## Description

When a Web Component micro frontend calls `LuigiClient.setAnchor('value')`, the `SET_ANCHOR_LINK_REQUEST` event is dispatched by the WC client API but no `addEventListener` is registered in `core-modular`'s `addListeners()` function to handle it and update the URL hash fragment.

## Expected behavior

Calling `LuigiClient.setAnchor('LuigiRocks')` from a WC should append `#LuigiRocks` to the URL (same as it works for iframe-based MFEs in the classic Luigi framework).

## Actual behavior

The event is silently dropped — the URL is not updated.

## Root cause

In `@luigi-project/core-modular`, the `addListeners(element, luigi)` function (which registers event handlers on the luigi-container element) handles `ADD_NODE_PARAMS_REQUEST`, `ADD_SEARCH_PARAMS_REQUEST`, etc. but does NOT register a listener for `SET_ANCHOR_LINK_REQUEST`.

## Origin

Discovered during e2e test migration from `luigirepo/test/e2e-test-application/cypress/e2e/tests/1-angular/luigi-client-lifecycle-manager-features.cy.js` → `test-lifecycle-manager.cy.js`.

walmazacn:
**Description**

- adds anchor handling in core modular

**Related issue(s)**

Resolves #5307


JohannesDoberer (changes requested):
It does not work for me.
I have a node 
```
{
  pathSegment: 'wc',
   label: 'wc',
   viewUrl: 'http://localhost:4400/helloWorldWC.js',
   webcomponent: true
},
```
**The url is localhost:4400/#/home2/wc**

The web component has two buttons calling `LuigiClient.setAnchor('value')` and `LuigiClient.getAnchor()`.
When setAnchor is called I'm getting an alert displaying `*luigi.notExactTargetNode* en` + redirected to localhost:4400/#/home2 when **no** `pageNotFoundHandler` is defined.
If pageNotFoundHandler is defined the app does what is configured in pageNotFoundHandler.

When I'm using the core-api impl it is the same.
No anchor will be added to url.

Does it work for you?

JohannesDoberer (changes requested):
When a anchor is set either via core-api or client (both cases works) I can't get the anchor via LuigiClient but via core-api.
LuigiClient.getAnchor() returns '';
Mabye also a test LuigiClient would be good

JohannesDoberer:
Setting and getting an anchor works but when I'm going in history back and forth via browser I'm getting an alert *luigi.notExactTargetNode* en.
Could you investigate if it is an issue regarding this issue or a general problem. If general I would fix it in a separate issue.

JohannesDoberer (dismissed):
LGTM

JohannesDoberer (approved):
LGTM

PR #5356, opened 2026-07-24
fix: resolve getCurrentRoute promise for empty route paths in web components

JohannesDoberer:
## Summery
- Fix getCurrentRoute() in the web component client API to always resolve the promise, regardless of the route value

## Problem
When a web component calls LuigiClient.linkManager().fromVirtualTreeRoot().getCurrentRoute() and the current position is at the virtualTree root, the computed route is '' (empty string — a valid result meaning "at root").

However, the promise callback used a truthy check:
```
(route) => {
  if (route) {
    resolve(route);
  }
}
```
Since '' is falsy, the promise was never resolved nor rejected — it hung indefinitely. This caused await getCurrentRoute() to never return, making the feature unusable for virtualTree root positions.

In contrast, the classic core implementation always resolves synchronously via a Proxy on window.Luigi.navigation(), so the same scenario works there.

## Fix
Remove the conditional check and always resolve with the route value as-is:
```
(route) => {
  resolve(route);
}
```
This passes through whatever getCurrentRoutePath returns — including '' or undefined — letting the caller decide how to handle it.


walmazacn (approved):
LGTM 👍

hardl (approved):

PR #5357, opened 2026-07-24
Fixes 'onNodeActivation' hook in core modular

### LINKED ISSUE #5256: onNodeActivation hook not checked during client-initiated navigation in core-modular
## Description

The `onNodeActivation` hook configured on navigation nodes is not being evaluated during client-initiated navigation (e.g. via `LuigiClient.linkManager().navigate()`). Navigation proceeds even when `onNodeActivation` returns `false`.

## Root Cause

In `core-modular/src/services/navigation.service.ts`, the `shouldPreventNavigation` / `shouldPreventNavigationForPath` method is only called inside `openViewInNewTab` (line 861), but **not** inside `handleNavigationRequest` which handles the main navigation flow.

In the old Luigi Core, `shouldPreventNavigation` was checked at 4 different points:
- `core/src/App.svelte` (3 occurrences, lines 883, 972, 1094)
- `core/src/services/routing.js` (line 58)

## Expected Behavior

When a node has `onNodeActivation` returning `false`, navigation to that node should be prevented regardless of the navigation source (side nav click, API call, or client linkManager).

## Steps to Reproduce

```javascript
// Luigi config
{
  pathSegment: 'not-navigated',
  label: 'Prevent Navigation',
  viewUrl: '/some-page.html',
  onNodeActivation: node => {
    Luigi.showAlert({ text: 'Navigation prevented', type: 'info' });
    return false;
  },
}

// From within MFE:
LuigiClient.linkManager().navigate('not-navigated');
// Expected: navigation is prevented, alert is shown
// Actual: navigation proceeds to /not-navigated
```

## Suggested Fix

Add `shouldPreventNavigationForPath` check in `handleNavigationRequest` before executing the navigation:

```typescript
// In handleNavigationRequest, after building the path:
if (await this.shouldPreventNavigationForPath(W)) {
  return;
}
```

## Labels
- e2e-migration
- bug

walmazacn:
**Description**

- fixes 'onNodeActivation' hook in core modular

**Related issue(s)**

Resolves #5256

### DIFF
--- a/core-modular/src/core-api/navigation.ts
+++ b/core-modular/src/core-api/navigation.ts
@@ -104,6 +104,10 @@ export class Navigation {
       return Promise.reject(new Error('Navigation with an absolute path prevented.'));
     }
 
+    if (await this.navService.shouldPreventNavigationForPath(path)) {
+      return;
+    }
+
     if (!modalSettings?.keepPrevious) {
       const closed = await this.modalService.closeModalsWithDirtyCheck();
       if (!closed) return;
@@ -141,6 +145,10 @@ export class Navigation {
     if (path === '/') {
       console.warn('Navigation with an absolute path prevented.');
       return Promise.reject(new Error('Navigation with an absolute path prevented.'));
+    }
+
+    if (await this.navService.shouldPreventNavigationForPath(path)) {
+      return;
     }
 
     const normalizedPath = path.replace(/\/\/+/g, '/');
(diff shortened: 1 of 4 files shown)


JohannesDoberer (changes requested):
When I have a onNodeActivation hook and try to open the specific node via `Luigi.navigation().openAsModal()`, the hook won't be called. Same for openAsDrawer

JohannesDoberer (approved):
LGTM

PR #5358, opened 2026-07-24
Feature/confirmation modal button confirm false

JohannesDoberer:
### Summary
Adds the ability to hide the confirm button in the ConfirmationModal by setting `buttonConfirm: false` in the settings — matching the existing behavior in the core.
Extends the ConfirmationModalSettings.buttonConfirm type from string to `string | false`.
Fixes the translation logic in core-api/ux.ts and ux-module.ts to preserve the false value instead of overriding it with the default translation key.


walmazacn (approved):
LGTM 👍

PR #5378, opened 2026-07-30
Race condition in container

### LINKED ISSUE #5376: Race condition: Loading indicator never hides for cached web components opened in modal
## Summary

When opening a web component micro frontend in a modal via `Luigi.navigation().openAsModal()`, the loading indicator stays active infinitely **if the web component has already been loaded/registered previously** (e.g., navigated to in the main content area first).

## Root Cause

In `container/src/LuigiContainer.svelte`, the `wc_ready` event listener is registered **after** `renderWebComponent()` is called. When the web component is already registered in `window.customElements` (cached from a previous load), `renderWebComponent()` → `attachWC()` executes **synchronously**, dispatching the `wc_ready` event immediately — before the listener exists. As a result, the `INITIALIZED` event is never dispatched and `hideLoadingIndicator` is never called.

### Code path (container/src/LuigiContainer.svelte):

```javascript
// 1. renderWebComponent() is called FIRST
webcomponentService.renderWebComponent(viewurl, ...);

// 2. wc_ready listener is registered AFTER
(thisComponent.getNoShadow() ? thisComponent : mainComponent).addEventListener('wc_ready', () => {
  // This never fires because wc_ready already happened synchronously
  thisComponent.initialized = true;
  webcomponentService.dispatchLuigiEvent(Events.INITIALIZED, {});
});
```

### Why it works on page reload with modal URL:

On a fresh page load, the web component is **not yet in the customElements registry**. The `registerWCFromUrl()` uses a dynamic `import()` which is asynchronous, giving the `wc_ready` listener time to be registered before `attachWC()` fires the event.

### Why it works in the main content area:

When navigating to the node for the first time, the WC is not cached yet → async import → listener registers in time. On subsequent navigations in the main area, the container may be reused from a view group cache rather than being recreated, so the race condition doesn't manifest.

## Steps to Reproduce

1. Start the `core-modular/dev-tools/simple-app` dev server
2. Navigate to any page (e.g., `/home2/c1`) — **do NOT** navigate to the WC node first
3. Open the browser console and run:
   ```javascript
   Luigi.navigation().openAsModal('/home2/wc')
   ```
4. **Expected:** Modal opens, loading indicator disappears, WC content is visible
5. **Actual:** Modal opens but the loading indicator (busy indicator) stays active forever, hiding the WC content

### Key detail — to trigger the bug reliably:

Navigate to `/home2/wc` in the main area **first** (this registers the custom element), then navigate away, then call `openAsModal('/home2/wc')`. The WC is now cached in `customElements`, causing the synchronous code path.

## Proposed Fix

Register the `wc_ready` event listener **before** calling `renderWebComponent()`:

```javascript
// Register listener FIRST
if (!skipInitCheck) {
  (thisComponent.getNoShadow() ? thisComponent : mainComponent).addEventListener('wc_ready', () => {
    if (
      !(thisComponent.getNoShadow() ? thisComponent : mainComponent)._luigi_mfe_webcomponent
        ?.deferLuigiClientWCInit
    ) {
      thisComponent.initialized = true;
      webcomponentService.dispatchLuigiEvent(Events.INITIALIZED, {});
    }
  });
}

// THEN call renderWebComponent
const webComponentValue = GenericHelperFunctions.checkWebcomponentValue(webcomponent);
webcomponentService.renderWebComponent(viewurl, ...);
```

This ensures the listener is in place regardless of whether `attachWC()` executes synchronously (cached WC) or asynchronously (fresh import).

## Affected File

- `container/src/LuigiContainer.svelte` (around line 260-290 in the `initialize` function)

## Environment

- Luigi core-modular
- Any connector (UI5, custom)
- Any web component micro frontend that doesn't explicitly call `LuigiClient.luigiClientInit()` (relying on the automatic `wc_ready` mechanism)

walmazacn:
**Description**

- fixes race condition in container

**Related issue(s)**

Resolves #5376


JohannesDoberer (approved):
LGTM

PR #5387, opened 2026-08-03
withoutSync issue in core modular

### LINKED ISSUE #5282: withoutSync() navigation still replaces iframe content in core-modular
## Description

`LuigiClient.linkManager().withoutSync().navigate(path)` should change the URL and update the side navigation selection without replacing the iframe content. However, in core-modular the route change handler still swaps the iframe when the target node has a different `viewUrl`.

## Steps to reproduce

1. Navigate to a node (e.g. `/projects/pr2`) which loads `withoutsync.html` in the iframe
2. From the microfrontend, call `LuigiClient.linkManager().withoutSync().navigate('/projects/pr2/virtual-tree')`
3. The `virtual-tree` node has `viewUrl: '/multipurpose.html'` (different from parent)

## Expected behavior

- URL changes to `/projects/pr2/virtual-tree` ✓
- Side navigation highlights the target node ✓  
- Iframe content remains unchanged (still shows `withoutsync.html`) ✗

## Actual behavior

The iframe is replaced with the target node's viewUrl (`/multipurpose.html`), losing the original microfrontend content.

## Analysis

In core-modular's bundled source, the `hashchange` listener extracts the `withoutSync` flag:
```js
let n = !!e?.detail?.withoutSync;
this.handleRouteChange(getCurrentPath(...), n, t);
```

But `handleRouteChange` appears to still trigger the iframe replacement when it resolves the new route to a node with a different viewUrl.

## Impact

2 e2e tests from `navigation-4.cy.js` are blocked:
- "withoutSync -> it should remain on the current page and highlight side nav item"
- "withSync -> it should change page" (depends on withoutSync working to validate the contrast)

walmazacn:
**Description**

- fixes 'withoutSync' issue in core modular

**Related issue(s)**

Resolves #5282


JohannesDoberer (changes requested):
It does not work for me.
Follwing config:
```
{
          pathSegment: 'home2',
          icon: 'home',
          showBreadcrumbs: false,
          viewUrl: 'https://fiddle.luigi-project.io/examples/microfrontends/multipurpose.html',
          children: [
...
{
              pathSegment: 'c2',
              label: 'MFE2',
              icon: 'calendar',
              viewUrl: 'http://localhost:4400/developers.html#developers',
              children: [
                {
                  pathSegment: 'virtuel-tree',
                  label: 'VT',
                  icon: 'calendar',
                  viewUrl: 'http://localhost:4400/microfrontend.html#vt',
                  virtuelTree: true
                },
              ]
            },
```
I'm on `home2/c2`
When I run 
```LuigiClient.linkManager().withoutSync().navigate('/home2/c2/virtuel-tree')```
I don't have any container in the dom. Also the expected behaviour, which means developers.html is still unchanged in the content area does not work.

JohannesDoberer (changes requested):
The `showLoadingIndicator` call sits outside the `if (!withoutSync)` guard, so it fires unconditionally — even when no new container is created. Since no new container means no `INITIALIZED` event, the indicator is never hidden again.
Rest works as expected 👍

JohannesDoberer (approved):
LGTM

PR #5404, opened 2026-08-05
Fix/category without label

JohannesDoberer:
### Summary
- Fix nodes with invalid category objects (no `label` and no `id`) being rendered as broken category groups instead of flat navigation items
- Fix iframe content not updating when navigating between nodes that share the same base URL but differ only in the hash fragment
- Nodes with an invalid category are placed at the beginning of the left nav items list (matching old core behavior)

### Changes

**`navigation.service.ts`**
- A category is now only treated as valid if it is a string or an object with `id` or `label`. Otherwise the node renders as a standalone item.
- Nodes with an invalid category object are unshifted to the front of the items array.

**`ui-module.ts`**
- When reusing an existing container (same base URL, determined by `isSameUrl`), the previous `viewUrl` is now compared against the resolved one. If the URL changed (e.g. different hash), `updateViewUrl()` is called to notify the iframe of the new route. Previously only `updateContext()` was called, which does not propagate the `viewUrl` to the microfrontend.

### Test plan
- [ ] Configure a node with `category: { icon: 'group' }` (no label/id) → should render as a flat item at the top of the left nav
- [ ] Configure two sibling nodes with same base URL but different hashes (e.g. `/mfe.html#child3` and `/mfe.html#child4`) without `viewGroup` → clicking between them should update the iframe content

### DIFF
--- a/core-modular/src/modules/ui-module.ts
+++ b/core-modular/src/modules/ui-module.ts
@@ -358,6 +358,8 @@ export const UIModule = {
       }
 
       if (viewGroupContainer) {
+        const previousViewUrl = viewGroupContainer.viewurl;
+
         if (!withoutSync) {
           viewGroupContainer.style.display = 'block';
           viewGroupContainer.viewurl = resolvedViewUrl;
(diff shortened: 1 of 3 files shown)


walmazacn (dismissed):
LGTM 👍

hardl (approved):

PR #5446, opened 2026-09-01
Fix context switcher keyboard navigation and profile aria-expanded

c5346163:
**Description**

- Add keyboard support for the context switcher (Arrow Up/Down, Home/End, Escape, Enter/Space) so users can move between options without Tab
- Keep `aria-expanded` in sync on the context switcher trigger, including the `addNavHrefs` link trigger
- Remove the Enter/Space `keyup` handler that closed the menu again right after it opened
- Bind Fiori3 and Vega profile `aria-expanded` to the actual popover state instead of hardcoding `true`

This was developed with AI assistance (Cursor). The change is original to this codebase and follows the SAP AI contribution guideline.

**Related issue(s)**

No existing issue. Found while using the context switcher with keyboard: Down Arrow did not move to the next option, and Escape could interact badly with a profile control that stayed `aria-expanded="true"` while closed.

## Test plan
- [ ] Open the context switcher with Enter or click; first option is focused
- [ ] Arrow Down/Up move between options; Home/End jump to first/last
- [ ] Enter on the trigger does not immediately close the menu
- [ ] Escape closes the menu and returns focus to the trigger
- [ ] Profile control `aria-expanded` is `false` when closed and `true` when open

### DIFF
--- a/core/src/navigation/ContextSwitcher.svelte
+++ b/core/src/navigation/ContextSwitcher.svelte
@@ -1,5 +1,5 @@
 <script>
-  import { createEventDispatcher, onMount, getContext, beforeUpdate } from 'svelte';
+  import { createEventDispatcher, onMount, getContext, beforeUpdate, tick } from 'svelte';
   import { ContextSwitcherHelpers } from './services/context-switcher';
   import ContextSwitcherNav from './ContextSwitcherNav.svelte';
   import { LuigiConfig } from '../core-api';
@@ -213,14 +213,17 @@
     DropdownKeyboardHelpers.applyRovingTabindex(items, index);
   }
 
-  function closeAndFocusTrigger() {
+  async function closeAndFocusTrigger() {
     if (isDropdownOpen()) {
       toggleDropdownState();
     }
-    const trigger = document.querySelector('[data-testid="luigi-contextswitcher-button"]');
-    if (trigger) {
-      trigger.focus();
-    }
+    await tick();
+    setTimeout(() => {
+      const trigger = document.querySelector('[data-testid="luigi-contextswitcher-button"]');
+      if (trigger) {
+        trigger.focus();
+      }
+    }, 0);
   }
 
   function onTriggerMouseDown(event) {
--- a/core/src/navigation/ContextSwitcher.svelte
+++ b/core/src/navigation/ContextSwitcher.svelte
@@ -248,13 +248,14 @@
     DropdownKeyboardHelpers.handleTriggerKeydown(event, {
       isOpen: isDropdownOpen(),
       isDisabled: !renderAsDropdown || event.currentTarget.getAttribute('aria-disabled') === 'true',
+      isAnchor: event.currentTarget.tagName === 'A',
       onToggle: (focus) => {
         focusMenuOnOpen = focus || 'first';
-        if (DropdownKeyboardHelpers.isActivationKey(event)) {
+        if (event.currentTarget.tagName === 'A' && DropdownKeyboardHelpers.isActivationKey(event)) {
           skipNextTriggerClick = true;
           setTimeout(() => {
             skipNextTriggerClick = false;
-          }, 50);
+          }, 500);
         }
         toggleDropdownState();
       },
@@ -277,10 +278,11 @@
   <!-- DESKTOP VERSION (popover): -->
   {#if !isMobile}
     <div class="fd-shellbar__action fd-shellbar__action--desktop">
-      <div class="fd-popover fd-popover--right">
-        <!-- svelte-ignore a11y-click-events-have-key-events -->
+      <!-- svelte-ignore a11y-click-events-have-key-events -->
+      <!-- svelte-ignore a11y-no-static-element-interactions -->
+      <div class="fd-popover fd-popover--right" on:click|stopPropagation={() => {}}>
         <div class="fd-popover__control" role="presentation" on:click|stopPropagation={() => {}}>
-          {#if addNavHrefForAnchor && selectedOption !== config.defaultLabel}
+          {#if addNavHrefForAnchor && selectedOption}
             <a
               href={selectedOption ? getRouteLink(selectedOption) : undefined}
               class="fd-button fd-button--transparent fd-shellbar__button fd-button--menu fd-shellbar__button--menu lui-ctx-switch-menu"
--- a/core/src/navigation/ContextSwitcher.svelte
+++ b/core/src/navigation/ContextSwitcher.svelte
@@ -1,5 +1,5 @@
 <script>
-  import { createEventDispatcher, onMount, getContext, beforeUpdate } from 'svelte';
+  import { createEventDispatcher, onMount, getContext, beforeUpdate, afterUpdate } from 'svelte';
   import { ContextSwitcherHelpers } from './services/context-switcher';
   import ContextSwitcherNav from './ContextSwitcherNav.svelte';
   import { LuigiConfig } from '../core-api';
@@ -10,7 +10,8 @@
     StateHelpers,
     NavigationHelpers,
     GenericHelpers,
-    EventListenerHelpers
+    EventListenerHelpers,
+    DropdownKeyboardHelpers
   } from '../utilities/helpers';
 
   const dispatch = createEventDispatcher();
@@ -40,6 +41,9 @@
   let selectedNodePath;
   export let addNavHrefForAnchor;
   let isContextSwitcherDropdownShown;
+  let popoverEl;
+  let wasDropdownShown = false;
+  let focusMenuOnOpen = 'first';
 
   onMount(async () => {
     StateHelpers.doOnStoreChange(store, async () => {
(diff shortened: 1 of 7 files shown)


JohannesDoberer (commented):
Hello Konrad,

thank you for this PR — it addresses a valid and worthwhile feature request.

After reviewing the changes, I noticed that switching to arrow key navigation and removing Tab navigation within the menu constitutes a breaking change for existing users.

The recommended approach would be to support both navigation patterns simultaneously: Tab and Shift+Tab would continue to move focus between items (preserving backward compatibility), while arrow keys provide the new ARIA-compliant navigation.

Let me know if you have any questions — happy to help.

Best regards,

Johannes

core/src/navigation/ContextSwitcher.svelte
JohannesDoberer, line 206:
```suggestion
        if (isDropdownOpen()) {
```
Is needed to close the dropDown automatically

core/src/navigation/ContextSwitcher.svelte
```diff
@@ -195,30 +219,136 @@
...
+  }
+
+  $: if (!isMobile) {
+    setContentInert(Boolean(dropDownStates && dropDownStates.contextSwitcherPopover));
+  }
+
+  function isDropdownOpen() {
+    return Boolean(dropDownStates && dropDownStates.contextSwitcherPopover);
+  }
+
```
JohannesDoberer, line 243:
```suggestion
 function keyboardNavEnabled() {
    return !customOptionsRenderer;
  }
```

core/src/navigation/ContextSwitcher.svelte
```diff
@@ -195,30 +219,136 @@
...
+    const items = DropdownKeyboardHelpers.getMenuItems(root);
+    if (!items.length) {
+      return;
+    }
+    const index = which === 'last' ? items.length - 1 : 0;
+    DropdownKeyboardHelpers.applyRovingTabindex(items, index);
+  }
+
+  function scheduleMenuFocus() {
+    if (isMobile) {
```
JohannesDoberer, line 255:
```suggestion
    if (isMobile || !keyboardNavEnabled()) {
```

core/src/navigation/ContextSwitcher.svelte
```diff
@@ -195,30 +219,136 @@
...
+    if (skipNextTriggerClick) {
+      skipNextTriggerClick = false;
+      return;
+    }
+    focusMenuOnOpen = 'first';
+    toggleDropdownState();
+    scheduleMenuFocus();
+  }
+
+  function onTriggerKeydown(event) {
```
JohannesDoberer, line 303:
```suggestion
  function onTriggerKeydown(event) {
    if (!keyboardNavEnabled()) {
      const isDisabled = !renderAsDropdown || event.currentTarget.getAttribute('aria-disabled') === 'true';
      if (isDisabled || event.repeat) {
        if (DropdownKeyboardHelpers.isActivationKey(event)) {
          event.preventDefault();
        }
        return;
      }
      if (DropdownKeyboardHelpers.eventKey(event) === 'Escape') {
        if (isDropdownOpen()) {
          event.preventDefault();
          closeAndFocusTrigger();
        }
        return;
      }
      if (DropdownKeyboardHelpers.isActivationKey(event)) {
        event.preventDefault();
        event.stopPropagation();
        skipNextTriggerClick = true;
        setTimeout(() => {
          skipNextTriggerClick = false;
        }, 1000);
        toggleDropdownState();
      }
      return;
    }
```

core/src/navigation/ContextSwitcher.svelte
```diff
@@ -195,30 +219,136 @@
...
+        toggleDropdownState();
+        scheduleMenuFocus();
+      },
+      onFocusFirst: () => focusMenuItem('first'),
+      onFocusLast: () => focusMenuItem('last'),
+      onClose: closeAndFocusTrigger
+    });
+  }
+
+  function onPopoverKeydown(event) {
```
JohannesDoberer, line 324:
```suggestion
  function onPopoverKeydown(event) {
    if (!keyboardNavEnabled()) {
      if (DropdownKeyboardHelpers.eventKey(event) === 'Escape') {
        event.preventDefault();
        closeAndFocusTrigger();
      }
      return;
    }
```

core/src/navigation/ContextSwitcher.svelte
```diff
@@ -230,21 +360,17 @@
               {/if}
             </a>
           {:else}
-            <button
+            <!-- svelte-ignore a11y-click-events-have-key-events -->
+            <div
```
JohannesDoberer, line 364: Is there a reason why you had to change it from `button` to `div`?
I run the e2e tests locally and had no issues. Discussed also with the architect and decided to leave it how it is, because you can also consider it as "breaking change" and if it is technically not necessary, we do not change it.

core/src/navigation/ContextSwitcherNav.svelte
```diff
@@ -22,11 +26,89 @@
...
     dispatch('goToOption', { option, selectedOption });
   }
+
+  function clearFocusTimeout() {
+    if (focusTimeout) {
+      clearTimeout(focusTimeout);
+      focusTimeout = undefined;
+    }
+  }
+
```
JohannesDoberer, line 36:
```suggestion
  
  function keyboardNavEnabled() {
    return !customOptionsRenderer;
  }
```

core/src/navigation/ContextSwitcherNav.svelte
```diff
@@ -22,11 +26,89 @@
...
+  function focusOpenItem() {
+    // Two ContextSwitcher instances share dropDownStates. The mobile copy
+    // must not steal focus from the desktop popover that the e2e asserts on.
+    if (!isContextSwitcherDropdownShown || !menuEl || isMobile) {
```
JohannesDoberer, line 40:
```suggestion
    if (!isContextSwitcherDropdownShown || !menuEl || isMobile || !keyboardNavEnabled()) {
```

core/src/navigation/ContextSwitcherNav.svelte
```diff
@@ -22,11 +26,89 @@
...
+    if (menuItems.includes(document.activeElement)) {
+      return true;
+    }
+    const index = focusMenuOnOpen === 'last' ? menuItems.length - 1 : 0;
+    DropdownKeyboardHelpers.applyRovingTabindex(menuItems, index);
+    return menuItems.includes(document.activeElement);
+  }
+
+  function recaptureFocusFromTrigger() {
+    if (!isContextSwitcherDropdownShown || isMobile) {
```
JohannesDoberer, line 61:
```suggestion
    if (!isContextSwitcherDropdownShown || isMobile || !keyboardNavEnabled()) {
```

core/src/navigation/ContextSwitcherNav.svelte
```diff
@@ -22,11 +26,89 @@
...
+    if (!items.length) {
+      return;
+    }
+    if (!items.includes(document.activeElement)) {
+      focusOpenItem();
+    }
+  }
+
+  afterUpdate(() => {
+    if (!isContextSwitcherDropdownShown) {
```
JohannesDoberer, line 79:
```suggestion
    if (!isContextSwitcherDropdownShown || !keyboardNavEnabled()) {
```

.github/workflows/test.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       actions: write
       contents: read
     with:
-      buildcache_key: buildcache-${{ github.ref }}
+      buildcache_key: buildcache-${{ github.ref }}-${{ github.sha }}
```
JohannesDoberer, line 24: What is the reason regarding the pipeline change?
Please revert this changes. We had in the past some issues with caching. If there are issues with our pipeline we need a separate issue for it.

JohannesDoberer (changes requested):
Hey Kondrad,
I found some issues with customOptionsRenderer. If a config developer has it's own options he couldn't implement his own a11y functionality because the DropdownKeyboardHelpersClass expects that a menuItem has the class `fd-menu__link`. This is a breaking change.
A config like this demonstrates it:
```
navigation: {
...
  contextSwitcher: {
  ...
        customOptionsRenderer: (item, selected) => {
          const option = document.createElement('a');

          option.setAttribute('value', item.link);
          option.setAttribute('href', item.link);
          option.setAttribute('tabindex', '0');
          option.setAttribute('role', 'option');
          option.classList.add('lui-ctx-switch-option');
          option.textContent = item.label;

          if (selected) {
            option.setAttribute('aria-selected', 'true');
            option.classList.add('is-selected');
          }

          return option.outerHTML;
        },
        customSelectedOptionRenderer: (item) => {
          const option = document.createElement('a');

          option.setAttribute('value', item.link);
          option.setAttribute('href', item.link);
          option.setAttribute('tabindex', '0');
          option.setAttribute('role', 'option');
          option.setAttribute('aria-selected', 'true');
          option.classList.add('lui-ctx-switch-option', 'is-selected');
          option.textContent = item.label;

          return option.outerHTML;
        },
  }
}
```
This wouldn't work. I decided to turn the arrow up/down navigation if a customOptionsRenderer is implemented.

Another issue is regarding the new setContentInert() in ContextSwitcher.svelte, which marks background content as inert while the popover is open, but its selector document.querySelectorAll('.iframeContainer, iframe') matches every iframe in the document. Since a Luigi app can be embedded alongside other content in the same DOM (foreign iframes such as analytics, chat widgets, or a host application that Luigi itself runs inside), this inadvertently disables iframes that don't belong to Luigi — and in multi-iframe setups (SplitView, ViewGroups) it hits all of them, not just the content behind the popover. On top of that, the reset on close removes inert unconditionally from every match, so any inert that was already present beforehand (e.g. set by the embedding host) gets wiped out, because the code doesn't track which nodes it actually modified. This is a regression versus main, where the context switcher never touched iframes at all. The fix scopes the operation to Luigi's own visible microfrontend iframes via IframeHelpers.getVisibleIframes() and only restores the nodes it set itself.

Possible solution could look like this:
```
let inertedNodes = [];

  function setContentInert(isInert) {
    if (typeof document === 'undefined') {
      return;
    }
    if (isInert) {
      IframeHelpers.getVisibleIframes().forEach((node) => {
        if (!node.hasAttribute('inert')) {
          node.setAttribute('inert', '');
          inertedNodes.push(node);
        }
      });
    } else {
      inertedNodes.forEach((node) => node.removeAttribute('inert'));
      inertedNodes = [];
    }
  }
```

PR #5453, opened 2026-09-07
Config changed in core modular

### LINKED ISSUE #5414: core-modular: configChanged() does not propagate events to open modals
## Summary

When `Luigi.configChanged()` is called while a modal is open, core-modular does not send any lifecycle events to the modal MFE. In Luigi Core, the modal correctly receives events depending on the `routing.showModalPathInUrl` setting.

### With `routing.showModalPathInUrl: true`

Calling `Luigi.configChanged()` with a modal open:
- **Main area** MFE → `addContextUpdateListener` fires ("Luigi Client context updated")
- **Modal** MFE → `addInitListener` fires ("Luigi Client initialized")
  - The modal is closed and re-opened from the URL, creating a new iframe

How it works: `configChanged()` → `handleRouteChange(configChangedInitiated=true)` → `handleBookmarkableModalPath(true)` → reads modal path from URL → `openAsModal(_reuseIframeOpener=true)` → `resetMicrofrontendModalData()` + `openViewInModal()`

### With `routing.showModalPathInUrl: false`

Calling `Luigi.configChanged()` with a modal open:
- **Main area** MFE → `addContextUpdateListener` fires ("Luigi Client context updated")
- **Modal** MFE → `addContextUpdateListener` fires ("Luigi Client context updated")
  - The modal stays open and receives a context update (same as main area)

## Core-modular behavior (current)

Regardless of `showModalPathInUrl` setting:
- **Main area** MFE → `addContextUpdateListener` fires ✓
- **Modal** MFE → **nothing** — no events received ✗

The `_ui.update()` method re-renders navigation and calls `updateMainContent` for the content area iframe, but does not propagate context updates to modal iframes and does not handle the bookmarkable modal path.

## Expected behavior

1. **`showModalPathInUrl: true`**: On `configChanged()`, close and re-open modal from URL → modal MFE receives `addInitListener`
2. **`showModalPathInUrl: false`**: On `configChanged()`, send context update to modal iframe → modal MFE receives `addContextUpdateListener`

## Affected e2e tests

- `test-context-update-listener.cy.js` — "Context in modal" test is skipped until this is resolved
- Original luigirepo test: `cypress/e2e/tests/microfrontends.cy.js` → "Context Update Listener test" → "Context in modal"

## References

- Core: `core/src/App.svelte` line 366-372 (configChanged triggers handleRouteChange)
- Core: `core/src/services/routing.js` lines 524-537 (handleBookmarkableModalPath)
- Core: `core/src/App.svelte` line 1462-1466 (_reuseIframeOpener logic)

walmazacn:
**Description**

- improves events for config changed in core modular

**Related issue(s)**

Resolves #5414


core-modular/src/modules/ui-module.ts
```diff
@@ -171,6 +171,37 @@ const setWebcomponentCreationInterceptor = (
...
+  luigi: Luigi
+): Promise<void> => {
+  const showModalPathInUrl = luigi.getConfigValue('routing.showModalPathInUrl');
+
+  if (showModalPathInUrl) {
+    const modalService = serviceRegistry.get(ModalService);
+    const routingService = serviceRegistry.get(RoutingService);
+    const routeInfo = RoutingHelpers.getCurrentPath(luigi, true, true);
+
+    await modalService.closeModalsWithDirtyCheck();
```
JohannesDoberer, line 186: Dirty-check result is ignored
```
await modalService.closeModalsWithDirtyCheck();
await routingService.handleBookmarkableModalPath(routeInfo);
```
The return value of `closeModalsWithDirtyCheck()` is discarded. When the modal has unsaved changes, this method awaits the unsaved-changes dialog and returns false if the user cancels — in that case closeModals() is skipped and the modal deliberately stays open (modal.service.ts:101-114).

Since we don't check the result, execution continues into `handleBookmarkableModalPath` regardless. Because the container is still in the DOM (close was aborted), it takes the `if (modalContainer)` branch and calls `updateModalSettings(...)` on it — so the modal the user just chose to keep still receives a settings update on configChanged(). The user's "cancel" is effectively bypassed.

`handleRouteChange` already respects this contract `(if (!closed) return);` this path should do the same, e.g. bail out early when the dirty check reports the modal wasn't closed.

core-modular/src/modules/ui-module.ts
```diff
@@ -171,6 +171,37 @@ const setWebcomponentCreationInterceptor = (
...
+  withoutSync: boolean,
+  luigi: Luigi
+): Promise<void> => {
+  const showModalPathInUrl = luigi.getConfigValue('routing.showModalPathInUrl');
+
+  if (showModalPathInUrl) {
+    const modalService = serviceRegistry.get(ModalService);
+    const routingService = serviceRegistry.get(RoutingService);
+    const routeInfo = RoutingHelpers.getCurrentPath(luigi, true, true);
+
```
JohannesDoberer, line 185: Double dirty-check when the modal was navigated away. When there's no modalPath in the URL, we close once here, then handleBookmarkableModalPath hits its own if (!modalPath) branch and calls closeModalsWithDirtyCheck() again. Harmless on an already-closed modal, but it can trigger the dirty prompt twice.

JohannesDoberer (changes requested):
See comments

core-modular/src/modules/ui-module.ts
```diff
@@ -171,6 +171,39 @@ const setWebcomponentCreationInterceptor = (
...
+  context: Record<string, any>,
+  withoutSync: boolean,
+  luigi: Luigi
+): Promise<void> => {
+  const showModalPathInUrl = luigi.getConfigValue('routing.showModalPathInUrl');
+
+  if (showModalPathInUrl) {
+    const modalService = serviceRegistry.get(ModalService);
+    const routingService = serviceRegistry.get(RoutingService);
+    const routeInfo = RoutingHelpers.getCurrentPath(luigi, true, true);
```
JohannesDoberer: Given scenario:
```
useHashRouting: false,
showModalPathInUrl: true
```
- Open a modal
- call `Luigi.configChanged()`

Result: url still have modal data but modal does not reopen.
Reason is probably hard coded hashRouting:true as parameter

JohannesDoberer (changes requested):
When I open a drawer and call `Luigi.configChanged()` only the main area and not the mf in the drawer gets un updated context.
In addition for userSettings custom mf's the contextupdate does not work.

Don't know if this should fixed also here or in a follow-up isse

core-modular/src/modules/ui-module.ts
```diff
@@ -410,6 +447,7 @@ export const UIModule = {
             viewGroupContainer.updateViewUrl(resolvedViewUrl);
           } else {
             viewGroupContainer.updateContext(currentNode.context || {}, { withoutSync: !!withoutSync });
+            handleDialogContainers(currentNode.context || {}, !!withoutSync, luigi);
```
JohannesDoberer, line 450: handleDialogContainers is async but here is no wait or .catch()

core-modular/src/modules/ui-module.ts
```diff
@@ -423,6 +461,7 @@ export const UIModule = {
         } else {
           if (!preventContextUpdate && currentContainer) {
             currentContainer.updateContext(currentNode.context || {}, { withoutSync });
+            handleDialogContainers(currentNode.context || {}, !!withoutSync, luigi);
```
JohannesDoberer, line 464: handleDialogContainers is async but here is no wait or .catch()

core-modular/src/modules/ui-module.ts
```diff
@@ -171,6 +171,44 @@ const setWebcomponentCreationInterceptor = (
   }
 };
 
+const handleDialogContainers = async (
```
JohannesDoberer, line 174: unit test for this fn would be good

JohannesDoberer (changes requested):
With my last comment I was to general due to userSettingsDialog.
userSettings dialog gets also a context update but only userSettingsData.

Please check and compare always with core.

Another scenario:
showModalPathInUrl: false

- open a modal
- make this mf dirty (LuigiClient.uxManager().setDirtyStatus(true))
- call Luigi.configChanged

In core I'm getting a dirty check confirmation modal in core-modular not

JohannesDoberer (approved):
LGTM

PR #5462, opened 2026-09-17
harden: sanitize child_process call in bundle.js

anupamme:
## Summary
Harden input handling in `bundle.js` (flagged by semgrep).

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | javascript.lang.security.detect-child-process.detect-child-process |
| **Severity** | HIGH |
| **Scanner** | semgrep |
| **Rule** | `javascript.lang.security.detect-child-process.detect-child-process` |
| **File** | `bundle.js:31` |
| **Assessment** | Defensive hardening |

**Description**: Detected calls to child_process from a function argument `folder`. This could lead to a command injection if the input is user controllable. Try to avoid calls to child_process, and if it is needed ensure user input is correctly sanitized or sandboxed. 

## Threat Model Context

This is a private Node.js application (not published to npm). Vulnerabilities affect this application's own runtime only.

## Changes
- `bundle.js`

## Behavior Preservation
The change is scoped to 1 file on the vulnerable path.

---
*This patch removes an exploit primitive — a code pattern that, while not independently exploitable today, could be chained with other weaknesses by automated exploit-development tooling. Proactive removal of such primitives raises the bar against increasingly capable automated attack tools.*

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*


bundle.js
```diff
@@ -26,9 +26,7 @@ function runCommand(folder, index, totalFolders, operation) {
   const startTime = new Date();
 
   return new Promise((resolve, reject) => {
-    const command = `cd ${folder} && npm run ${operation}`;
-
-    exec(command, (error, stdout, stderr) => {
+    execFile('npm', ['run', operation], { cwd: folder }, (error, stdout, stderr) => {
```
hardl:
```suggestion
    execFile('npm run ' + operation, { cwd: folder, shell: true }, (error, stdout, stderr) => {
```
needed to also work on windows OS

hardl (changes requested):
Thanks for your contribution!

hardl (approved):

